PR #5, opened 2016-09-24
Logos, graphics, icons

nukeop:


nukeop:
Also better search bar

PR #37, opened 2017-04-17
Fix the website  text color same as the background 

sudoliyang:
fix the text color same as the background color

### DIFF
--- a/style.css
+++ b/style.css
@@ -88,7 +88,7 @@ body {
 }
 
 .mc-field-group>input {
-  background-color: #00000000;
+  background-color: transparent;
   border: 0px;
   border-bottom: 4px solid #8d53e7;
   width: 100%;


nukeop:
Could you please take a screenshot of the element in question?

This is what it's supposed to look like:

screenshot

Is your browser overriding some css settings and changing the input field background to white?

sudoliyang:
Hi 
On my laptop look like this 


macOS Version 10.12.4 (16E195)
Google chrome Version 57.0.2987.133 (64-bit)

sudoliyang:
You are correct！
My Google Chrome have default `user agent stylesheet`


Change the input field background color to `transparent` can resolve this problem.

nukeop:
Nice, it seems to work on other browsers too. If you'd like to make this change and commit, then I will merge everything. Thanks for the help.

sudoliyang:
Yes, I had already finish this commit. 😃

PR #46, opened 2017-09-27
Adds Arch Linux package to the list of avaliable packages.

mikelpint:
It is located in the .releases directory.


nukeop:
Thanks, I will modify the readme later and add a section about third party packages. If you're okay with that I could list you there as the maintainer of this one.

mikelpint:
Yes, I'm okay

nukeop:
Do you need to add windows binaries to .releases? I'm concerned they would just take space for no good reason. I don't really know how Arch packages work so let me know if it's necessary. Might also be a good idea to remove .deb and tar.gz (appimage should work fine).

Also, you've changed some fonts for some reason.

mikelpint:
I only need AUR and .pkg.tar.xz.

mikelpint:
I don't know why

mikelpint:
I will do a new pull request.

PR #50, opened 2017-10-05
Adds Arch Linux package.

mikelpint:
I don't know why the fonts are changed. Every time I do a git clone from the repository nukeop/nuclear they appear like that.


nukeop:
Thanks. Can you remove the fonts from your commit? I would like to avoid changing unrelated files.

mikelpint:
Every time I clone the repository that happens and I don't know why.

mikelpint:
I will fix that later.

nukeop:
Cool. In the new version fonts are loaded differently (as needed when building) so you won't run into this problem in 0.4.0 (which will be released later this year).

mikelpint:
Ok :grinning:

PR #59, opened 2018-02-20
Modified desktop file.

mikelpint:
I don't know if it is fine because my computer didn't let me build it.


nukeop:
It should be fine, but you also changed the versions of some packages. Could you remove that part? I update packages every now and then myself, but every time I have to test if everything keeps working as before, so I'd rather not do it right now.

mikelpint:
OK

mikelpint:
Fixed.

nukeop:
Hm, there are still some changes in package.json. If you leave only the changes after line 120, it'll be good.

PS. If you can't build it on your PC, check travis output (in this case it looks like it builds fine!)

mikelpint:
I think that it is fine now. Good night anyway.

nukeop:
Yep that's perfect. I can merge now, thanks for your help.

PR #109, opened 2019-01-11
Added an auto radio feature

trekiteasy:
When the auto radio player control option is enabled, and when we are currently playing the latest element of the queue, nuclear will add a random top track from a similar artist to be played next in the queue.


trekiteasy:
OK, will solve the PR quality issues and issue a new pull request.

PR #110, opened 2019-01-12
Added an auto radio feature

trekiteasy:
When the auto radio player control option is enabled, and when we are currently playing the latest element of the queue, nuclear will add a random top track from a similar artist to be played next in the queue.

### DIFF
--- a/app/actions/dashboard.js
+++ b/app/actions/dashboard.js
@@ -1,17 +1,14 @@
 import logger from 'electron-timber';
 import core from 'nuclear-core';
-import {
-  getBestNewAlbums,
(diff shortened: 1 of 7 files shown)


nukeop:
Thanks for this feature, this is pretty amazing and works really well. I built and ran your fork locally to make sure everything's fine. This is actually a feature I planned to add myself at some point (you can see it here: https://github.com/nukeop/nuclear/projects/3 ), so double thanks for implementing it. As a side note, be sure to setup git on your machine correctly so that your contributions are attributed to your account in the future.

PR #130, opened 2019-01-16
some fixes when values are undefined

sepiropht:
Some littles fixes i need to make it work for me. Awesome project!


nukeop:
Thanks for the contribution, I really appreciate every pull request I receive.

PR #139, opened 2019-01-18
Fixed non-master album info being fetched from the wrong source.

### LINKED ISSUE #134: Master Release not found
I get discogs: `master release not found` error with most of the artists

Steps:
 - Search for `Cardi B`
 - Try to play any song

CAnderson97:
http://api.discogs.com/masters/ was being used for all albums resulting in an error 404 if the album was just a release.
http://api.discogs.com/releases/ is used to get information on release albums.

fixes #134


CAnderson97:
This doesn't change the search results, so it shouldn't create duplicates. Release versions show up on the artist page if there is no master available.

nukeop:
Ok, I'll review what it does later today  and if it doesn't cause too much noise I'll merge it.

nukeop:
It does look like it solves some problems with searching for album details. Now if I could only figure out how to filter out all these re-releases, countless compilations, bootlegs and other spammy entries.

PR #155, opened 2019-01-19
Changed genre from using album.styles to album.genres. Fixed genre not showing up.

CAnderson97:
Unless there is specific reason not to, I think it is better to use genres instead of styles for showing the album genre.

The album genre was not showing for me on any album. "if (_.some(_.map([album.images, album.artists, album.genres], _.isEmpty)))" checks if album.genres is empty, so I think using "album.genres[0]" instead of "album.styles && Object.keys(album.styles) > 0 && album.styles[0]" should be fine.

### DIFF
--- a/app/components/AlbumView/index.js
+++ b/app/components/AlbumView/index.js
@@ -62,7 +62,7 @@ class AlbumView extends React.Component {
 
   render() {
     let { album } = this.props;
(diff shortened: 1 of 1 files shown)


nukeop:
Hm,I think it was "styles" in some instances too, but this seems to work, so I'll merge it.

PR #160, opened 2019-01-21
Fixed artists without artwork not loading.

CAnderson97:
When selecting an artist with no artwork I get the error "TypeError: Cannot read property '0' of undefined". This fix simply checks to make sure artist.images is defined before trying to access its elements.

I tested using artist "Omoi"


CAnderson97:
Good idea. Ill try that.

CAnderson97:
I changed it. The code is much simpler, thank you.

Issue #159, opened 2019-01-20
Dragging things around freezes them, might crash the app


vilanz:
As the title says, whenever I - for example - accidentally drag a song or a sidebar item somewhere, it gets frozen in place. In some situations - sidebar items for instance - it will inevitably make the app crash.

I'm running `electron:prod` on Ubuntu 18.10.

nukeop:
Dragging is handle by an external  package, I'll take a look if we can fix this.

qualitymanifest:
Looks like it might be related to https://github.com/electron/electron/issues/14430 and https://github.com/electron/electron/issues/14073

In those issues the user reported the problem just from dragging around text. If I type something in the search bar and then select and attempt to drag it, I get the same problem - the app freezes.

PR #168, opened 2019-01-21
Fix drag freezing issue when run from command line

qualitymanifest:
Closes https://github.com/nukeop/nuclear/issues/159

Used workaround from https://github.com/electron/electron/issues/12820#issuecomment-387798184

Dragging now works without issue when app is run from any `npm run electron:*` command. Tested using Linux Mint 19 (Ubuntu 18.04).

Per this comment https://github.com/electron/electron/issues/12820#issuecomment-452702533 this issue may be fixed in electron 3.0.14. If/when electron gets updated in this app, you can probably remove the `--disable-gpu` flags. Since that would likely require dealing with some breaking changes, this should work for now.


vilanz:
For some users, the --disable-gpu option fixes issues with Linux (glibc specifically) and the project's Electron version

nukeop:
Disabling GPU rendering altogether just to work around an Electron bug seems too drastic.

nukeop:
Maybe I'll add this as a debug option to be enabled by users should they so wish, I don't like the idea of disabling GPU rendering everywhere in all cases because of a bug.

qualitymanifest:
Fair enough, makes sense. Do you want me to close this? This is one of my first PRs so I'm not really sure what the proper etiquette is here.

nukeop:
I'll keep it open for some time so I don't forget to implement this. Thanks for the help either way, at least you found out how to fix this issue. I'll try updating electron to v4 first too.

nukeop:
I added this as an option in https://github.com/nukeop/nuclear/pull/190

Issue #193, opened 2019-01-24
Dashboard > Best new Music > Best new tracks - No possibility to play or add the track


trekiteasy:
I would expect to be able to easily play / add the track to the queue

trekiteasy:
This is now possible in dev version.

trekiteasy:
We lost this in the revamp of the dashboard...

sgraczyk:
Hi, I decided to dig into this issue as my first one. I managed to implement it although I'm a little bit worried about the UI/UX aspect of my solution. I've found that currently the possibility to play a track or add it to the queue is performed using a popup/tooltip. I used the same mechanism but I believe it looks a little bit odd. I've attached a screenshot as an example.

![image](https://user-images.githubusercontent.com/8076673/58103073-a1ddbe00-7be2-11e9-9006-4d47124bf3f7.png)

I'm open to different ideas regarding this issue, maybe I've missed a different way to add track (in UI/UX manner).

nukeop:
I think a better way to do this would be to add an empty star icon next to the track title, clicking it adds it to favorites. If it's already there, the button is replaced by a filled star and clicking it lets you remove it.
Maybe instead of using a popup you could add the same buttons under the review header? This way they're easy to find and obvious.

sgraczyk:
Ok, I just wanted to make sure if that popup was some kind of standard in the application. I'll add the buttons under the review header then. Will consider the star idea also.

sgraczyk:
Hi, I'm back with some spare time so I can finish this task. I've just managed to resolve conflicts so it shouldn't take me long to provide a version to review.

Hope that's fine and no other similar work is in progress ;)

nukeop:
Hey, very nice. Nobody else picked up this one so your work is still very much appreciated.

DRob1260:
I'd like to take a crack at this. I thought a good solution might be to update the `Dashboard/BestNewMusicTab/BestNewMusicContent` component to have a similar design to the top portion (see screenshot) of the `AlbumView` component. Thoughts?

<img width="1067" alt="Screen Shot 2020-08-02 at 8 39 26 AM" src="https://user-images.githubusercontent.com/19953794/89124559-a156dd80-d49d-11ea-95ea-226f083f2ed5.png">

nukeop:
Sounds nice, how would you switch between the different albums and tracks?

DRob1260:
[...]

PR #210, opened 2019-01-27
Added play button to Charts. Adds support for issue #193

ivyjsgit:


app/components/Dashboard/index.js
```diff
@@ -1,11 +1,13 @@
 import React from 'react';
 import { Tab } from 'semantic-ui-react';
 
+import '../../../app/actions/player'
```
nukeop: This won't do what you want it to do. Please look at the containers (in `app/containers` to see how to pass actions down to individual components).

trekiteasy:
I think this pull is not appropriate anymore as the trackrow component brings the play and add buttons to the charts.

PR #215, opened 2019-01-28
Added 'play now' button to top charts

WalkerFrederick:
I added a 'play now' button to the context-popup, just like how it might be available when you click on a song in search 

playbutton


app/components/Dashboard/ChartsTab/index.js
```diff
@@ -1,10 +1,13 @@
 import React from 'react';
+import { bindActionCreators } from 'redux';
```
nukeop: could you remove this import?

WalkerFrederick: Whoops, forgot to take that out. Yeah one sec

Issue #200, opened 2019-01-25
Under Home>Genres, you can't directly play a song


ivyjsgit:
Steps to reproduce:
1. Go to dashboard
2. Click Genres
3. Click a Genre
4. Click a song from the list.
Expected result:
Should display "Play song"
Actual result:
Only displays "add to queue"

WalkerFrederick:
Can we close this issue? @nukeop

nukeop:
Sure, i thought it'd close automatically.

PR #216, opened 2019-01-29
added the 'play now' option to the genre/tagview | fixes issue #200

WalkerFrederick:
Just like I did for the top charts I added the play now button to the genre/tagview


WalkerFrederick:
Why did that check fail?, it's literally the same thing I did to top charts

Issue #219, opened 2019-01-29
Code deduplication


nukeop:
This thread will be dedicated to finding and eliminating instances of needless code duplication. Anyone wanting to contribute something easy to do can either remove a duplication they find or submit one here for removal/refactoring.

trekiteasy:
In most cases, I think we can come up with something similar... and this would remove a lot of duplicated code and simplify further maintenance on track actions. This is work in progress. Will submit pull request when ready.

jonatascastro12:
I was thinking about using Wrapper and Factory pattern to handle TrackRow creation, as we have code duplication of TrackRow creation at:

app/components/PlaylistView/index.js
app/components/ArtistView/PopularTracks/index.js
app/components/AlbumView/index.js
app/components/SearchResults/TracksResults/index.js
app/components/Dashboard/ChartsTab/index.js
app/components/TagView/TagTopTracks/index.js

Does it make sense @nukeop, @trekiteasy?

nukeop:
It does to me and I'll be glad if you show me how to do something better. Nobody's even looked at the pile of code I churned out in the past 2 years for a long, long time, and I'm sure there are many spots where things can be improved a lot.

PR #221, opened 2019-01-29
Creation of a TrackRow component

trekiteasy:
As discussed in https://github.com/nukeop/nuclear/issues/219#issuecomment-458537841
The objective is to centralize the display of a Track in a component to simplify further maintenance or new interactions.


app/components/AlbumView/index.js
```diff
@@ -202,65 +204,36 @@ class AlbumView extends React.Component {
...
+              key={'album-track-row-' + index}
+              track={track}
+              index={'album-track-' + index}
+              clearQueue={this.props.clearQueue}
+              addToQueue={this.props.addToQueue}
+              startPlayback={this.props.startPlayback}
+              selectSong={this.props.selectSong}
+              musicSources={this.props.musicSources}
+              displayCover={false}
+              displayTrackNumber={true}
```
nukeop: for boolean props that are set to true, you only need to put their name there (as in `displayTrackNumber` instead of `displayTrackNumber={true}`. Instead of setting props to false, you can simply omit them and they'll be false by default.

app/components/ArtistView/PopularTracks/index.js
```diff
@@ -15,107 +14,77 @@ class PopularTracks extends React.Component {
...
+                    artist={artist}
+                    clearQueue={this.props.clearQueue}
+                    addToQueue={this.props.addToQueue}
+                    startPlayback={this.props.startPlayback}
+                    selectSong={this.props.selectSong}
+                    musicSources={this.props.musicSources}
+                    displayCover={true}
+                    displayArtist={false}
+                    displayDuration={false}
+                    displayPlayCount={true}
```
nukeop: Same as above.

app/components/PlaylistView/index.js
```diff
@@ -102,16 +94,51 @@ class PlaylistView extends React.Component {
...
           {this.renderPlaylistInfo()}
           <div className={styles.playlist_tracks}>
-            {playlist.tracks.map((track, i) =>
-              this.renderPlaylistTrack(track, i)
-            )}
+            <table>
+              {this.renderPlaylistTracksHeader()}
+              <tbody>
+                {playlist.tracks.map((track, index) => {
+                  _.set(track, 'artist.name', track.artist);
```
nukeop: Dangerous, this will mutate the props and interfere with Redux. If you want to change the props, best clone them before doing that. Also, why are you doing that here?

app/components/SearchResults/TracksResults/index.js
```diff
@@ -1,53 +1,45 @@
...
-            if (track) {
-              if (track.name && track.image && track.artist) {
-                {
-                  return this.renderTrack(track, i);
+        <table>
+          <tbody>
+            {(collection || []).slice(0, limit).map((track, index) => {
+              if (track && _.hasIn(track, 'name') & _.hasIn(track, 'image') && _.hasIn(track, 'artist')) {
+                if (!track.artist.name) {
+                  track.artist = { name: track.artist };
```
nukeop: As above, this will mutate the props. Better to use `Object.assign`, e.g. `let trackCopy = Object.assign({}, track, {artist: {name: track.artist}})`, this way you get a clone of the object.

app/components/TrackRow/index.js
```diff
@@ -0,0 +1,133 @@
...
+    ];
+
+    return (
+      <ContextPopup
+        key={'track-' + index}
+        trigger={
+          <tr className={styles.track}>
+            {this.props.displayCover ? <td
+              style={{
+                backgroundImage: `url(${_.get(track, 'image[0][#text]') || artPlaceholder})`,
```
nukeop: instead of using `||`, put that `artPlaceholder` as the last argument of `_.get`.

nukeop:
Great contribution. I added some comments to point out some minor things that can be done better. Once we get that out of the way, we can merge this.

app/components/TrackRow/index.js
trekiteasy: Done in commit #4957aac

trekiteasy:
Have done some changes. One question remain on how to deal with different data models depending on data sources...

app/components/ArtistView/PopularTracks/index.js
trekiteasy: Done in commit #711879d

app/components/PlaylistView/index.js
nukeop: The original idea about track metadata was this:
* All data originates with a Discogs search. Initial track data is copied verbatim from what Discogs returns
* The above is augmented with lastfm data, by adding a "lastfm" key that contains data copied from lastfm verbatim
* As a track is added to the play queue, it is further augmented with stream data, returned by appropriate plugins

trekiteasy: I understand. But for instance, for Track Search, we do not rely on discogs but on lastfm, so the data doesn't have exactly the same format...

nukeop: Ok, for now let's just get rid of instances where global state is mutated, and we can worry about a uniform track class later.

nukeop: Simply clone the object from the redux state before changing its properties.
So e.g. instead of:
`_.set(track, 'artist.name', track.artist);`

do something like this:

```javascript
const newTrack = _.cloneDeep(track);
_.set(newTrack, 'artist.name', newTrack.artist);
```

Or use `Object.assign` to clone the object, similar to how it's done in all the reducers.

nukeop:
Excellent.

Issue #226, opened 2019-01-30
Mute control


trekiteasy:
A click on the volume symbol should
- put the volume to 0
- change the volume symbol to a "muted" symbol

Another click on the "muted" symbol should
- put back the volume to **where it was initially** 
- change the volume symbol back to the initial symbol

Chandankkrr:
I can implement it later tonight

Chandankkrr:
@nukeop @trekiteasy  Pushed a PR for implementing volume mute control. Had to introduce state in  `VolumeControl` component with `isVolumeMuted` and `volume` properties. 

Thought of implemeting mute control by passing in a flag to `react-sound` package but it does not have props that supports this. 

The way to implement mute control is by passing 0 to the volume props of the `Sound` component within `react-sound`. The value of the volume slider is stored in `volume` property of the state  to prevent the volume slider from jumping around when mute is on.

trekiteasy:
This is implemented in current dev build. I close the feature request.

SharlSherif:
Hi, but it doesn't visually drag the volume slider back to 0 when muted :
![image](https://user-images.githubusercontent.com/29375572/56507220-e5c89f00-6520-11e9-9252-64f675f71158.png)
is that intended or a bug ?

SharlSherif:
incase its not an intended behavior, in `\app\components\VolumeControls\VolumeSlider\index.js`
i edited the `render` of `VolumeControls`  class from 

```
render () {
    return (
      <div className={styles.volume_controls_container}>
        <PlayOptionsControls
          toggleOption={this.props.toggleOption}
          settings={this.props.settings}
        />
        <div
          className={styles.volume_speaker_control}
          onClick={this.toggleMute.bind(this)}
        >
          <FontAwesome name={this.props.muted ? 'volume-off' : 'volume-up'} />
        </div>

        <VolumeSlider
          fill={this.props.fill}
          handleClick={this.handleClick.bind(this)}
        />
      </div>
    )
  }
```

to 
```
render () {
    return (
      <div className={styles.volume_controls_container}>
        <PlayOptionsControls
          toggleOption={this.props.toggleOption}
          settings={this.props.settings}
        />
        <div
          className={styles.volume_speaker_control}
          onClick={this.toggleMute.bind(this)}
        >
          <FontAwesome name={this.props.muted ? 'volume-off' : 'volume-up'} />
        </div>

        <VolumeSlider
[...]

PR #229, opened 2019-01-31
Added ability to mute volume when clicking on volume icon

Chandankkrr:
Implemented Feature request (https://github.com/nukeop/nuclear/issues/226)

Volume icon changes to fontawesome's `volume-off` icon when clicking on it

> When volume is not muted

volume_speaker_control_on

> When volume is muted
volume_speaker_control

### DIFF
--- a/app/actions/player.js
+++ b/app/actions/player.js
@@ -6,6 +6,8 @@ export const PAUSE_PLAYBACK = 'PAUSE_PLAYBACK';
 export const UPDATE_PLAYBACK_PROGRESS = 'UPDATE_PLAYBACK_PROGRESS';
 export const UPDATE_SEEK = 'UPDATE_SEEK';
 export const UPDATE_VOLUME = 'UPDATE_VOLUME';
+export const MUTE_ON = 'MUTE_ON';
+export const MUTE_OFF = 'MUTE_OFF';
 export const UPDATE_PLAYBACK_STREAM_LOADING = 'UPDATE_PLAYBACK_STREAM_LOADING';
 
 export function startPlayback() {
@@ -55,6 +57,31 @@ export function updateVolume(volume) {
   return {
     type: UPDATE_VOLUME,
     payload: volume
+  };
+}
+
+export function muteOn(){
+  return {
+    type: MUTE_ON,
+    payload: null
+  };
+}
+
+export function muteOff(){
+  return {
+    type: MUTE_OFF,
+    payload: null
+  };
+}
+
+export function toggleMute(mute) {
+  return dispatch => {
+    if (mute){
+      dispatch(muteOn());
+      dispatch(updateVolume(0));
+    } else {
+      dispatch(muteOff());
+    }
   };
 }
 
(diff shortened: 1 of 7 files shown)


nukeop:
State in presentational components is to be avoided. We already keep volume here:
https://github.com/nukeop/nuclear/blob/master/app/reducers/player.js

The volume controls are rendered by the top level `App` container:
https://github.com/nukeop/nuclear/blob/master/app/App.js#L228

And it also has access to the `player` which can be passed down to `VolumeControls` and then to `VolumeSlider`. This way everything is kept in redux, avoiding stateful components. That's where you should add the 'muted' flag too.

Chandankkrr:
Alright, that makes sense. Will make those changes and push them out.

Chandankkrr:
@nukeop Please review updated changes

app/actions/player.js
```diff
@@ -6,6 +6,9 @@ export const PAUSE_PLAYBACK = 'PAUSE_PLAYBACK';
 export const UPDATE_PLAYBACK_PROGRESS = 'UPDATE_PLAYBACK_PROGRESS';
 export const UPDATE_SEEK = 'UPDATE_SEEK';
 export const UPDATE_VOLUME = 'UPDATE_VOLUME';
+export const TOGGLE_MUTE = 'TOGGLE_MUTE';
+export const MUTE_ON = 'MUTE_ON';
+export const MUTE_OFF = 'MUTE_OFF';
```
fabianschwarzfritz: Wouldn't "MUTE" and "UNMUTE" be the better name for these constants? Just a suggestion...

app/actions/player.js
```diff
@@ -58,6 +61,31 @@ export function updateVolume(volume) {
...
+    type: MUTE_OFF,
+    payload: null
+  };
+}
+
+export function toggleMute(muted) {
+  return dispatch => {
+    if (muted) {
+      dispatch(muteOn());
+      dispatch(updateVolume(0));
```
nukeop: I think this line can (and should be removed), we won't be able to restore previous volume on unmuting otherwise.

app/actions/player.js
Chandankkrr: Agree, thanks for the feedback. I'll update the values.

nukeop: Yeah, in general we want action types to be verbs ("do thing" as opposed to "thing on")

app/actions/player.js
Chandankkrr: @nukeop Removing the `dispatch(updateVolume(0))`  wont work. Initially when I started implementing the ability to mute volume when clicking on the volume icon, I looked at how the existing sound output works. I came across `SoundContainer` which renders a `Sound` component from `react-sound` npm package which ultimately uses the `SoundManager2` library. The `react-sound` package does not accept a flag to `mute` or `unmute` the volume. The way to implemet currently is by passing in 0 for volume. This is the reason I introduced state for the `VolumeControl` component to save the volume slider's value.

nukeop: Check my comment in your pull request, we'll need to edit another line to make it work.

Chandankkrr: Yep missed the bit there. Thanks for that. Will push the changes soon.

Chandankkrr: Also, do we want to change from `mute` to `unmute` if the user changes the volume slider?

nukeop:
Btw, don't worry about merging, merge conflicts, etc, I'll handle that when you finish working on your branch.

nukeop:
Thanks for your help.

Chandankkrr:
No problem. Thanks for your feedback too.

Issue #223, opened 2019-01-30
Multiple Popover in view 


Chandankkrr:
Currently clicking through queued songs opens up an associated popup with stream information. Ideally would like to close existing popup before rendering a new one.

![image](https://user-images.githubusercontent.com/5574015/51975117-9b152900-24d5-11e9-89d1-1cbdaf4a6269.png)

WalkerFrederick:
Same kind of thing happens when you click on a song in the top-charts, opens a context window but doesn't close existing ones.

PR #232, opened 2019-02-01
Multiple queue popovers fix

WalkerFrederick:
just using toggleOpen instead of handleOpen will get rid of existing popups. I have no clue why but I think it has something to do with line 22 

> this.container.click();	

fixes issue #223


nukeop:
I'm pretty sure this won't be reliable, and trying to open the same popup the second time will fail, but I'll check this out later.

WalkerFrederick:
that's what I thought but it seems to be fine on my machine.

nukeop:
I just checked this and it really does seem to be working, I'm not sure why either. I don't remember why that extra click was added there, probably a workaround for some Semantic UI quirk.

PR #247, opened 2019-02-08
Feature/lyrics

trekiteasy:
A Lyrics search feature : when a track is playing, we can go to the Lyrics page (from the left menu) to get the lyrics.


nukeop:
Would be good if we were able to check if the lyrics actually exist, it seems that a http result 200 is not enough. Maybe there's a package that searches for lyrics with fallbacks? If not, this would be a good candidate to go into nuclear-core, or its own separate package (like I did with the Pitchfork best new music scraper).

I'm getting some nonsense as lyrics sometimes:

2019-02-09-020212_722x230_scrot

app/components/LyricsView/index.js
```diff
@@ -0,0 +1,52 @@
...
+    return (<div className={styles.lyrics_text}>{lyricsStr}</div>);
+  }
+
+  renderLyricsHeader () {
+    let track = this.props.track;
+    return (
+      <div className={styles.lyrics_header_overlay}>
+        <div className={styles.lyrics_header_container}>
+          <div className={styles.lyrics_artist_name_container} >
+            <h1>{track.name} by {track.artist}</h1>
```
nukeop: Better to use the Header component, for uniform style across all views, instead of a plain <h1>

trekiteasy:
You are right : 
- I spinned-off the lyric part to an external package (simple-get-lyrics).
- there were already some fallbacks  to look on different sources, but I added some filtering to avoid the "Unfortunately..." message
- I used the Header component for the title (and changed the line-height to better display multiline titles)

nukeop:
Fantastic, great work.

PR #248, opened 2019-02-08
Autoradio Averager + enhancement

trekiteasy:
Added autoradio averager
- autoradio code was separated to a different file
- autoradio is now based on similar tracks (and not only similar artists)
- autoradio takes into account a defined number of tracks from the queue to compute next track
Next steps :
- the different parameters are currently const variables. This should be simplified and made changeable through apps settings (something like : how random should be the autoradio : 0 = very conservative 1=big variations)


nukeop:
I'm reluctant about coupling all that autoradio code with SoundContainer, since that container was supposed to be very simple: give it a stream URL, and playback parameters, and you get a playing song. I know there was already some last.fm-related scrobbling code in there, but this is a whole new ball game.

To solve it in a cleaner way, it would be better to move everything that's not related to playing songs outta there - I mean, last.fm stuff, autoradio stuff, and even switching songs on stream end, etc. Instead, make a list of subscribers to SoundContainer events, and on each event SoundContainer notifies the subscribers with relevant info instead of directly calling their methods. This way we can decouple all that unrelated code and make it easier to extend in the future.

If you want, you can implement this yourself, but if you don't feel like doing it (since this would require a significant effort), I can just merge this PR and do it later myself. Up to you which way you'd like to go with this.

trekiteasy:
I understand and was initially reluctant as well to put everything here (that's also why I separated some of the code to a new file).
Being fairly new to react (and not being a professionnal developper), I'm not sure I am confident enough to reorganize the code properly. I could help but would need more guidance on what to put where and how to put in place subscribers etc.

nukeop:
Ok, it's fine, we can rework this later.

Issue #192, opened 2019-01-24
Dashboard > Best new Music > Best new albums - Not to possible click on the album


trekiteasy:
I would expect to be able to open an album view from the dashboard when I click on a new album

nukeop:
For many new albums Discogs won't have their data yet. Same goes for #193

trekiteasy:
Ok... I understand. 
Maybe it is something we can at least try to fetch, give the possibility to open the album / track if found ?
Otherwise, I strongly believe this "Best new album" / "Best new tracks" sections are worse than having nothing ; it generates frustration for me.
If no solution is found, I'd advise to put some lastfm charts of some other charts as a first tab.

nukeop:
Best we can do is open the search page with the album/track filled in. Anyway the best new music page needs to be reworked slightly in terms of alignment, i never liked how it scales when there's a lot of text and could be more useful.

trekiteasy:
We lost this in the dashboard revamp...

nukeop:
Yeah, I'm aware. I'll be restoring this soon.

PR #250, opened 2019-02-08
Fix to #192 : add ability to click on album name in dashboard to open

trekiteasy:
… the album view

If the album is not found, we open the artist view.


nukeop:
I originally made only the artist clickable because the albums in this view are all pretty new, and usually there's no Discogs data for them, but this is a good fallback.

nukeop:
Looks like I was mistaken or Discogs improved their data sources, this works flawlessly. Great work.

PR #253, opened 2019-02-10
Toast notifications, closes #169

### LINKED ISSUE #169: Toast notifications
I'm thinking of adding toast notifications (think something like desktop notifications, but inside the player, likely in the bottom right or left corner) to display encountered errors and some async actions (such as download started/completed etc.).

Would you find these notifications helpful or annoying? Do you have any ideas regarding these you'd like me to consider?

Also, I welcome all suggestions on how to implement this best, since I've only done something like this once in another project. Should we use a third party react package?

nukeop:


nukeop:
I'll resolve conflicts tomorrow, the feature is ready to be used though.

There are actions that expose creating toasts and can be used by any container. This is as simple as it gets.

Issue #254, opened 2019-02-10
Add title attribute to icon buttons


websterchris:
**Nuclear Version** : 0.4.4
**OS**: Windows 10

Just stumbled across this repo and looks really great. 

I struggled initially trying to figure out what some of the icon buttons (e.g shuffle or queue repeat) did, so I wonder if it is worth adding a title attribute to the buttons to help the user understand what they are especially if the icon meaning is ambiguous?

Chandankkrr:
@nukeop I can implement this. Nice feature suggestion @websterchris. I was struggling to figure out what the buttons did as well.

Chandankkrr:
Could use `react-tooltip` https://react-tooltip.netlify.com/ if we want to control the position of the tooltip.

nukeop:
We already have tooltips from Semantic UI, no need for an extra package, but this will do for now.

nukeop:
Already implemented as a submenu with proper labels.

PR #259, opened 2019-02-12
Added title attribute to player controls

Chandankkrr:
Added title attribute to player controls #254


nukeop:
For consistency, we could use the `prettyName` attribute from here for those buttons: https://github.com/nukeop/nuclear/blob/master/app/constants/settings.js

Chandankkrr:
Yep, I had a look at the file, they are used in the setting section and the text for `loopAfterQueueEnd` is really long.

nukeop:
Ok, let's leave that for later then.

PR #262, opened 2019-02-12
merge read settings with default settings

charjac:
fix#238


nukeop:
Really smart way to do this, thanks.

PR #266, opened 2019-02-13
fix PlayerControls dont toggle icon when nothing on playlist

charjac:
fix#265
i am not sure about the css cursor.


charjac:
I coulb maybe replace undefined by null to make codacy happyer .....

PR #270, opened 2019-02-13
Make README more clear

JakeConnors376W:


nukeop:
You're removing the colorful language, which makes it for fun. Hell, that single paragraph about Electron has already been quoted on at least 3 other websites. No way I'm gonna neuter that.

ghostslayer989:
What are you doing? People can be religiously opposed from electron don't remove that m8.

Issue #173, opened 2019-01-22
docker version?


stkac:
# someone know how make nuclear run using docker?
## I mean using PM2 Entrypoint

thanks in advance,

nukeop:
I'll add this to the backlog, build&run process is well documented and pretty simple so this shouldn't be too much of a problem.

gabriel376:
Hi,

I would like to take this one.
This is a simple testing I created to allow build and run Nuclear inside a Docker container:

Dockerfile:
```
FROM node

RUN apt-get update && apt-get install -y libnss3 libgtk-3-0 libx11-xcb1 libxss1 libasound2

WORKDIR /opt
RUN git clone https://github.com/nukeop/nuclear

WORKDIR /opt/nuclear
RUN npm install && npm run build:dist

CMD ["npm", "run", "electron:prod"]
```

To build the image:
`docker build -t nuclear .`

To run the image:
`docker run --rm --net=host --env="DISPLAY" --volume="$HOME/.Xauthority:/root/.Xauthority:rw" --device /dev/snd nuclear`

(I am running Ubuntu 18.04 bionic)

Would be great if we can have dockerfiles for building and running Nuclear, for DEV and PROD. We can publish it to https://hub.docker.com too. The dockerfile is cloning the repository but another option is to share the source directory from host to container using volumes.

Any suggestions or tips ?

nukeop:
hm, since we'll likely host this dockerfile inside this repo (preferably at the root level), the git clone step can be omitted.

If you'd like, feel free to make a pull request with production/development images, and if you want you can host the images yourself on Docker hub, then add yourself to th readme in the third party packages table.

nukeop:
Btw, after `npm install`, `npm run build:dist` and `npm run build:electron` we can delete everything on the dev dependencies list to reduce the size of the image. Actually I'd try to approach this from another angle, build the scripts locally first and then the dockerfile can just copy the built binaries. Then, there's no need to even install the dev dependencies in the first place.

gabriel376:
I will do that, thanks!

PR #271, opened 2019-02-13
add npm script to build and run nuclear:prod using docker (#173)

gabriel376:
This PR implements #173

**Changes**
1. Created a Dockerfile 
2. Added 2 NPM scripts to package.json

**Details**
1. Dockerfile is based on Node image. It install some Video and Audio dependencies and run NPM scripts to build Nuclear electron pack
2. NPM has now one script to build Nuclear docker image `npm run build:docker` and another script to run Nuclear using docker `npm run electron:docker`

**Improvements**
1. Dockerfile can be based on Alpine image to generate smaller Docker image
2. We can have more NPM scripts using docker to build DEV Nuclear or Nuclear targeting other platforms
3. I tested using Ubuntu 18.04 Bionic host. I believe it will work with most Debian-based distros. We need to test in other distros and OS (mac, windows, fedora etc), especially the arguments passed to `docker run` like `--volume=\"$HOME/.Xauthority:/root/.Xauthority:rw\"`
4. We can host this docker image to https://hub.docker.com/


nukeop:
Looks and works great, thank you for the contribution.

gabriel376:
Thank you!

Issue #98, opened 2018-08-15
Replace dbus/mpris with pure js module


nukeop:

nukeop:
Moved to 0.6.0.

acrisci:
I'm working on [mpris-service](https://github.com/emersion/mpris-service) and going around doing mpris implementations on electron players. Is this still needed?

nukeop:
Yes, and in fact we were using your module originally, but dropped support due to problems with packaging the native dbus module in Appimage and other formats. Seeing as you've apparently transitioned to another dbus interface, we could be able to restore the code we've removed temporarily.

acrisci:
Ok cool. I have commit access to the [underlying dbus library](https://github.com/acrisci/node-dbus-next) so if you have any problems with it we can fix it.

acrisci:
I'm looking at this now. I tested some of the old mpris code and it doesn't work when I put it into the latest.

Trying to figure out how to programatically get the player to play, set the volume, set the position, etc. The actions seem to be tied up in the react components and I don't know much about react.

nukeop:
I used to handle this on the server side, with no contact with React.

Some commented out code is here:
https://github.com/nukeop/nuclear/blob/master/server/main.dev.linux.js

nukeop:
The idea is that the React part sends the relevant events via ipc, and the Electron part parses them and uses mpris to set everything as needed.

acrisci:
Ah ok the only things that are not implemented to be fully compatible with [playerctl](https://github.com/acrisci/playerctl) are:

* [ ] a getter to return the position in the current track when requested from dbus
* [ ] a player event when the track seeks to a new position
* [ ] a way to set the volume on the player
* [ ] a way to know when the volume changes
* [ ] a way to set the shuffle status on the player
* [ ] a way to know when the shuffle status on the player changes
* [ ] a way to get the loop status of the player
* [ ] a way to know when the loop status of the player changes
* [ ] a way to open a file from a uri

nukeop:
Done by @charjac

PR #278, opened 2019-03-09
feat: add a simple express rest api to remotly control nuclear

charjac:
All endpoints are POST except when specified.

playlist:
 - /nuclear/playlist GET (list of playlist)
 - /nuclear/playlist { name: 'myplaylist' }
 - /nuclear/playlist/:name DELETE

queue
 - /nuclear/queue/empty
 - /nuclear/queue GET

docs:
 - /nuclear/docs  GET swagger UI (html page)
 - /nuclear/docs/swagger.json GET

player:
  - /nuclear/player/play
  - /nuclear/player/pause
  - /nuclear/player/stop
  - /nuclear/player/play-pause
  - /nuclear/player/volume { value: 50 }
  - /nuclear/player/seek { value: 50000 }
  - /nuclear/player/now-playing GET
  - /nuclear/player/mute

window:
  - /nuclear/window/maximize
  - /nuclear/window/minimize
  - /nuclear/window/quit

settings:
 - /nuclear/settings GET
 - /nuclear/settings/:option GET
 - /nuclear/settings/:option { value: true }

If this suit your need, i can add a /player/now-playing endpoint and playlist management.

#120

PR #279, opened 2019-03-09
Update mpris implementation

acrisci:
Add a basic mpris implementation by updating to mpris-service 2.0 and
uncommenting old mpris code.

Ref #98


acrisci:
TODO:

* [ ] a getter to return the position in the current track when requested from dbus
* [ ] a player event when the track seeks to a new position
* [ ] a way to set the volume on the player
* [ ] a way to know when the volume changes
* [ ] a way to set the shuffle status on the player
* [ ] a way to know when the shuffle status on the player changes
* [ ] a way to get the loop status of the player
* [ ] a way to know when the loop status of the player changes
* [ ] a way to open a file from a uri

nukeop:
I think there will be conflicts with https://github.com/nukeop/nuclear/pull/278. We'll need to reconcile the changes in your and @charjac pull request to minimize the duplicated work.

nukeop:
Api is ready to go.

acrisci:
Ok ready for testing.

nukeop:
Fantastic, thank you for your contribution. I will test it as soon as possible.

webpack.config.electron.js
```diff
@@ -1,12 +1,18 @@
...
+const path = require('path');
 
 module.exports = env => {
   const entry = env && env.LINUX ? './server/main.dev.linux.js' : './server/main.dev.js';
 
   return {
     entry: entry,
+    resolve: {
+      alias: {
+        jsbi: path.resolve(__dirname, 'node_modules', 'jsbi', 'dist', 'jsbi-cjs.js')
```
nukeop: what's this?

acrisci: I'm not sure I need to look into why this works. @martpie came up with this. See [this issue](https://github.com/emersion/mpris-service/issues/23) on mpris-service.

martpie: JSBI ships with a `.mjs` version of the library. The thing is Webpack's `electron-main` will try to import this `.mjs` instead of the classic common JS file. Then things goes terribly wrong.

To me, this solution is a hack, but I am not sure if the problem comes from `jsbi` builds, or `node-dbus-next` (which seems to be written in cjs, but without a transpiler/bundler)

nukeop:
I was just looking for a program to use mpris from the terminal, and found playerctl. Only after reading the readme and clicking the author's profile did I realize that it's you. Small world, or maybe you're just the most prominent in that niche.

acrisci:
My thing is that I primarily do community-driven Linux Desktop projects :sunglasses:

This is part of a big project I'm doing to support mpris on electron players. It was inspired by people complaining on the Playerctl issue tracker that electron players don't work. I wrote about it here:

https://dubstepdish.com/index.php/2019/03/17/the-great-node-mpris-project/

nukeop:
I also found out that dbus doesn't play nicely with tmux (apparently this combo causes an invalid socket location being saved in the environment, breaking playerctl).

nukeop:
2019-03-18-223726_677x431_scrot


Have you seen something like this before?
This occurs on launch and is apparently caused by dbus-next, although it might be something on my end.

nukeop:
Yep, it can't connect to dbus at all.

acrisci:
Ah ok you're running a code path I haven't tested in dbus-next.

What is the value of `$DBUS_SESSION_BUS_ADDRESS`?

nukeop:
`unix:abstract=/tmp/dbus-pSgvVqOdpL,guid=1b9bd7435b409b88d4058c195c900e3d`

acrisci:
I'm pretty sure this is a problem with electron disabling some node api for the filesystem.

[GPMDP](https://github.com/MarshallOfSound/Google-Play-Music-Desktop-Player-UNOFFICIAL-) works ok with the abstract socket code path, so that's promising. I'm trying to see if there are differences in the webpack or electron configuration that get this to work.

The origin of the error is [here](https://github.com/TooTallNate/node-bindings/blob/master/bindings.js#L141-L184) within the node-bindings library, a dependency of the node `abstract-socket` library.

They are trying to get the name of the file to find the `abstract-socket` bindings using the node `Error` class and getting a stack trace.

However, the stack trace object (`st[i].getFileName()` below) is returning `undefined` so it can't build the search path properly. I think if this function worked, then this bug would be fixed.

```js
  Error.prepareStackTrace = function(e, st) {
    for (var i = 0, l = st.length; i < l; i++) {
      fileName = st[i].getFileName();
      if (fileName !== __filename) {
        if (calling_file) {
          if (fileName !== calling_file) {
            return;
          }
        } else {
          return;
        }
      }
    }
  };

  // run the 'prepareStackTrace' function above
  Error.captureStackTrace(dummy);
```

acrisci:
Here is [an issue](https://github.com/TooTallNate/node-bindings/issues/50) on the node-bindings repo with a similar issue. This seems to be related to webpack doing some sort of `eval` on the bindings library. This issue seems to have been resolved in another project [here](https://github.com/Walid-Shouman/Langformy/commit/13eb1c55b73ea073f7c3cda29fbeec7199dd0443). I tried just setting the webpack `node` config to

```js
node: {
  __filename: true,
  __dirname: true
}
```

That allows `fs` to be used, but still doesn't resolve the issue so there might be more to it.

acrisci:
Ok, so the real issue is that abstract-socket and bindings cannot be put in the bundle because the logic to find the bindings depends on the location of `abstract-socket` within the `node_modules` directory. And you can't include the compiled file in the bindings anyway.

This fixes the issue, but I'm not sure what the implications of using the `externals` is in webpack.

nukeop:
Does this work when we build the whole program in an Appimage, snap, etc? The whole problem with native dbus implementation was that it was impossible to bundle it correctly for these build targets, and it caused more trouble than it was worth. I'm fine if we have to use some black magic to get it to work, I'll test it later today.

nukeop:
#555 implements mpris in our updated codebase. Thanks for helping us along the way.

PR #282, opened 2019-03-12
Disable back and forward buttons when no history exists

hibroseph:
I saw that the back and forward history buttons do not disable when you cannot go back and forwards.

This also prevents when nuclear is loaded to go to the root and display nothing in the main container.

I didn't report this issue because it was an easy fix and I really wanted to contribute something and I knew you would jump on it really quickly since it was so simple.

open to code review

### DIFF
--- a/app/components/NavButtons/index.js
+++ b/app/components/NavButtons/index.js
@@ -1,33 +1,50 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import FontAwesome from 'react-fontawesome';
-
+import cx from 'classnames';
 import styles from './styles.scss';
 
-const NavButtons = props => {
+class NavButtons extends React.Component {
 
-  let enableBackButton = true;
-  let enableForwardButton = true;
-
-  if (props.historyCurrentIndex == 1) {
-    enableBackButton = false;
+  constructor(props) {
+    super(props);
+    this.enableBackButton = this.enableBackButton.bind(this);
+    this.enableForwardButton = this.enableForwardButton.bind(this);
   }
 
-  if (props.historyCurrentIndex == (props.historyLength - 1)) {
-    enableForwardButton = false;
+  enableBackButton(currentHistoryIndex) {
+    if (currentHistoryIndex == 1) {
+      return false;
+    } else {
+      return true;
+    }
   }
 
-  return (
-    <div className={styles.nav_buttons}>
-      <a href='#' onClick={(enableBackButton ? props.back : null)} className={(enableBackButton ? null : 'disable' )}>
-        {enableBackButton ? <FontAwesome name='chevron-left'/> : <FontAwesome name='chevron-left'/>}
-      </a>
-      <a href='#' onClick={(enableForwardButton ? props.forward : null)} className={(enableForwardButton ? null : 'disable')}>
-        {enableForwardButton ? <FontAwesome name="chevron-right"/> : <FontAwesome name='chevron-right'/>}
-      </a>
-    </div>
-  );
-};
+  enableForwardButton(currentHistoryIndex, historyLength) {
+    if(currentHistoryIndex == (historyLength - 1)) {
+      return false;
+    } else {
+      return true;
+    }
+  }
+
+  render() {
+    let { back, forward, historyLength, historyCurrentIndex} = this.props;
+
+    return (
+      <div className={styles.nav_buttons}>
+        <a href='#' onClick={(this.enableBackButton(historyLength) && back)} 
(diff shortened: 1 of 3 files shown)


app/components/NavButtons/index.js
```diff
@@ -5,13 +5,25 @@ import FontAwesome from 'react-fontawesome';
...
 const NavButtons = props => {
+
+  let enableBackButton = true;
+  let enableForwardButton = true;
+
+  if (props.historyCurrentIndex == 1) {
+    enableBackButton = false;
+  }
+
+  if (props.historyCurrentIndex == (props.historyLength - 1)) {
```
nukeop: This one and the condition above could be encapsulated in their own functions for simplicity.

app/components/NavButtons/index.js
```diff
@@ -5,13 +5,25 @@ import FontAwesome from 'react-fontawesome';
...
+    enableForwardButton = false;
+  }
+
   return (
     <div className={styles.nav_buttons}>
-      <a href='#' onClick={props.back} >
-        <FontAwesome name='chevron-left'/>
+      <a href='#' onClick={(enableBackButton ? props.back : null)} className={(enableBackButton ? null : 'disable' )}>
```
nukeop: instead of using trinaries, try using && and classnames module, e.g.:

```javascript
//at the top
import cx from 'classnames';
// ...
<a href='#' onClick={enableBackButton && props.back} className={cx({'disable': !enableBackButton})}>
```

app/components/NavButtons/index.js
```diff
@@ -5,13 +5,25 @@ import FontAwesome from 'react-fontawesome';
...
+        {enableBackButton ? <FontAwesome name='chevron-left'/> : <FontAwesome name='chevron-left'/>}
       </a>
-      <a href='#' onClick={props.forward}>
-        <FontAwesome name='chevron-right'/>
+      <a href='#' onClick={(enableForwardButton ? props.forward : null)} className={(enableForwardButton ? null : 'disable')}>
+        {enableForwardButton ? <FontAwesome name="chevron-right"/> : <FontAwesome name='chevron-right'/>}
```
nukeop: I think this line is unnecessary - it should look like before

nukeop:
Thanks for the PR. I left some remarks, so let's iterate on your code a little and then we can merge it.

hibroseph:
Okay, I made the changes you requested. 

If you could take a look and tell me what you think.

I am getting a warning about using the logical operation, I attached a screen shot of the warning later when I can

app/components/NavButtons/index.js
```diff
@@ -1,21 +1,50 @@
...
+class NavButtons extends React.Component {
+
+  constructor(props) {
+    super(props);
+    this.enableBackButton = this.enableBackButton.bind(this);
+    this.enableForwardButton = this.enableForwardButton.bind(this);
+  }
+
+  enableBackButton(currentHistoryIndex) {
+    if (currentHistoryIndex == 1) {
```
nukeop: Better to return this directly (and also use === instead of ==):

```javascript
return currentHistoryIndex === 1;
```

Similarly in the other function.

nukeop:
Nice. One small thing more and we'll be ready to go.

hibroseph:
image

Here is the image of the warning and I updated the code.

Also do we use a formatter like prettier or something?

nukeop:
Ok, I guess you should handle this like the warning tells you to. And please change the functions that check if buttons are enabled/disabled to return the boolean value directly.

There is eslint if you want to format the code automatically, `npx eslint [filename] --fix`.

hibroseph:
What do you mean to return the boolean value directly? 
and i'll switch it today after classes.

nukeop:
When you have a function that looks like this:
```javascript
if(condition) {
    return false;
} else {
    return true;
}
```

You can replace its body with `return !condition;` and it will have the same effect.

hibroseph:
oh okay, so essentially it would be
`if(condition) {
     return !condition;
} else {
     return condition;
}`

What is the benefit of doing that? Just not creating a new boolean variable?

nukeop:
Shorter, simpler, and more explicit. Whatever condition is after `if`, it's evaluated to either true or false anyway, so there's no point in duplicating that.

hibroseph:
Okay! I added those new changes!

nukeop:
Looks like I wasn't clear enough in what I meant, but don't worry, I'll correct this manually in a sec and merge it.

hibroseph:
Gotcha! Now I see exactly what you mean!

nukeop:
Great. As for the media buttons, I think somebody else already did that, or was supposed to do that.

Issue #286, opened 2019-03-13
Adding disabling features to the play button/skip buttons


hibroseph:
Adding the feature of what was added to the navigation buttons where there is no hover or active state and it is darkened for the buttons to stop or skip music. When you are able to perform an action, there is an active state and hover state.

nukeop:
@charjac I was pretty sure you added something like this a while ago?

charjac:
I just add a css style when button is not clickable:

```css
.player_button_disabled a {
  cursor: not-allowed;
}
```
I am sure it would be much better with darkened or shadow style.

charjac:
https://github.com/nukeop/nuclear/commit/8028caa17a1fd9cec9b109067b0466b7278e8c8a

nukeop:
Ah, ok, now I remember. In this case yes, this would be a great improvement.

hibroseph:
Great! I'll take it on. I am going to take a lot from what we did with the history button. Is there anything else you want me to be aware of?

nukeop:
Not really, we'll talk more when you submit your code. Thanks for the help.

PR #289, opened 2019-03-14
[WIP] Adding hovering and active states to the player buttons

hibroseph:
Going to keep updating here so you can check on what I am doing and give input as I am coding. I still need to add those styling elements to the forward and backward. 

What do you think of the styling currently on the pause button?
#286


nukeop:
To keep things consistent, it's better to use colors from the palette (vars.scss), instead of directly setting colors.

hibroseph:
Affirmative.

nukeop:
To be clear, feel free to use scss functions like darken or rgba (to set opacity) with palette colors.

hibroseph:
I just want to say, it took me **WAY** too long to realize that the sass operators (like the ampersand & or the color) need to have spaces. A lack of spaces between will alter their operation.

hibroseph:
How do they look right now? Should I do an active state?

nukeop:
You don't need to use & if you use a space.

```css
.a {
  .b { }
}
```

Is identical to
```css
.a .b { }
```

Which in both cases selects all bs inside as. However,
```css
.a {
  &.b {

  }
}
```
selects everything that has both class a and b at the same time.

hibroseph:
What about with puesdos? Don't I need the ampersand in front of puesdos, such as

```(javascript)
&:not(.player_button_disabled) :hover {
    transition: $short-duration; 
    color: darken($white, 50%);
  }
```

nukeop:
Yes, the ampersand is a shorthand for referring to the selector from the level above, so e.g.
```css
.a {
 &:hover { }
}
```

Refers to `a:hover`. 

The selector you created in your comment refers to any hovered-over element that is a child of an element that does not have the class `.player_button_disabled` (depending on how this is nested).

hibroseph:
That is what I intended.

Should I be more specific in my code about the children that I can hover
over?

nukeop:
As long as it works as you intended, it's all good.

hibroseph:
@nukeop sorry I didn't let you know what I had done. Want to take a look at it now?

nukeop:
Sure, I'll review soon after you commit your changes, no rush.

nukeop:
Hi, are you still working on it, or can it be improved/finished by somebody else? Or would you rather have us not use your code?

hibroseph:
Sorry about that! I thought I had committed my changes. Let me see where I left it at and commit those changes (if I have any)

nukeop:
Ok, cool.

Issue #287, opened 2019-03-13
Graceful exception handler


nukeop:
Many times we had an issue where some innocuous error crashes the whole program with it. Now that we have toasts, we could just display an error notification, log the error, and carry on. All uncaught exceptions can go through this system, although it might cause some buggy behavior.  Better than restarting the program though, to be sure.

nukeop:
Fixed by @charjac

PR #290, opened 2019-03-14
feat(error): add a global ErrorBoundary component

charjac:
According to this: https://reactjs.org/blog/2017/07/26/error-handling-in-react-16.html
when a component throw an error it should be remove from the tree.

#287

### DIFF
--- a/app/containers/ErrorBoundary/index.js
+++ b/app/containers/ErrorBoundary/index.js
@@ -1,0 +1,54 @@
+import logger from 'electron-timber';
+import React from 'react';
+import { withRouter } from 'react-router';
+import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
+import { bindActionCreators } from 'redux';
+
+import * as ToastActions from '../../actions/toasts';
+
+const initialState = { hasError: false };
+
+class ErrorBoundary extends React.Component {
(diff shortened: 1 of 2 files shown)


app/containers/ErrorBoundary/index.js
```diff
@@ -0,0 +1,54 @@
...
+class ErrorBoundary extends React.Component {
+  state = initialState;
+
+  static getDerivedStateFromError() {
+    return { hasError: true };
+  }
+
+  componentDidCatch(error) {
+    logger.error(error);
+    this.props.actions.error('error', 'Something wrong happen');
```
nukeop: happen**ed**

nukeop:
Nice, I think this will work way better for unexpected errors than what happens now.
We could also use electron-timber (already in use in some components) to log these, if possible. Can you try adding that?

app/containers/ErrorBoundary/index.js
charjac: done ! sorry for this i am french and we are terrible in foreign language :)

charjac:
I already use electron-timber in the componentDidCatch hook l20 of ErrorBoundary/index.js

nukeop:
Right, I missed that completely.

PR #301, opened 2019-04-02
[WIP] Add local file support

charjac:


charjac:
I add a simple proove of concept, i had to use the http api to serve the file.

app/actions/local.js
```diff
@@ -0,0 +1,46 @@
...
+    payload
+  };
+}
+
+export function scanLocalFoldersSuccess(payload) {
+  return {
+    type: SCAN_LOCAL_FOLDER_SUCCESS,
+    payload: payload.reduce((acc, track) => ({
+      ...acc,
+      [track.id]: track
```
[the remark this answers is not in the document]
charjac: It start here server/http/api/localfile.js, then it goes there server/file/index.js, then an event is send to the IpcContainer and the action is dispatched into redux.
The id is a simple uuid/v4.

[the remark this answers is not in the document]
nukeop: Nice, this is a good way to do this

charjac:
I think it only need some ui. How do you imagine it ?

server/http/api/localfile.js
```diff
@@ -0,0 +1,118 @@
...
+          .filter(artist => artist.includes(req.body.artist))
+          .map(artist => byArtist[artist])
+          .flat()
+          .filter(track => track.name.includes(req.body.track))
+          .map(track => ({
+            id: track.id,
+            title: track.name,
+            duration: track.duration,
+            source: 'Local',
+            stream: `http://127.0.0.1:${port}/nuclear/file/${track.id}`,
```
nukeop: Maybe it's possible to use the file:// scheme?

app/containers/SoundContainer/index.js
```diff
@@ -161,6 +161,8 @@ class SoundContainer extends React.Component {
       ).stream;
     }
 
+    console.log('render sound', streamUrl, player.seek);
```
nukeop: debug statement?

nukeop:
Very good. The way I see it is a list of directories that you've already added to your library, with simple controls like add more directories, remove those already added, etc. Maybe scan them right away, or put a button for scanning in there too.

For now, it'll be fine if we just have a long list of TrackRow components under that, representing all songs in your library. Later we can think about browsing by album, artist, etc. Maybe make the list sortable, using something like this: https://react.semantic-ui.com/collections/table/

server/http/api/localfile.js
charjac: if we want to use file:// schema we will have to disable some electron security (dont remember everything but the guy on stackoverflow said it is not a good idea).

app/containers/SoundContainer/index.js
charjac: oops :)

nukeop:
I'm going to checkout the code tomorrow and do some more thorough review of the functionality as it's a huge feature and I want to do it justice.

nukeop:
I noticed that if there's no album art the queue item popup is broken. It'll be worthwhile to fix that after this is merged.

nukeop:
I'd put the menu where you add folders in the Library section, since it's empty initially and there's no obvious way of adding music. Probably at the top, or in a tab.

nukeop:
This time I think it'll be crucial to have good options for browsing the local library, so we'll need sorting, filtering (with an input box at the top), and maybe album view like some players.

charjac:
i will work on this today

charjac:
it works, but its not beautiful :D.

nukeop:
Awesome, this is really a great feature. We can think about making it beautiful over time, what matters is that it's implemented and it works.

nukeop:
What this means is that all features for 0.5.0 are here and we can release soon!

PR #314, opened 2019-04-22
when the audio is muted, you can't click on the slider to unmute & change the volume level at the same time

SharlSherif:
I added this feature and thought you might be interested into implementing it.
there might be a better way to achieve this.
image


nukeop:
Thanks for the contribution, once you correct that cosmetic issue I'll merge this.

Issue #234, opened 2019-02-04
React/Redux Devtools


kjm425:
Anyone been able to get redux/react devtools working? I saw that they were commented out in server/main.dev.js, but I was wondering if anyone knew of a fix. Thanks

nukeop:
Last time I checked it was a known bug in Electron, but maybe it's been fixed by now. Need to verify.

charjac:
Hi, i check it is still not fixed, however it is possible to use https://github.com/zalmoxisus/remote-redux-devtools wich work just fine. I can make a pr if you want.

nukeop:
Okay, I'd appreciate that..

trekiteasy:
I'm curious : could you explain a bit more in detail how you use it with the current electron project ?

charjac:
There is many way to use it, personnaly i use the vscode extension:
 - you can use the standard extension (enable remote mode)
 - https://github.com/zalmoxisus/atom-redux-devtools
 - https://github.com/jkzing/vscode-redux-devtools
 - http://remotedev.io/local/
 - i once used a standalone app for debugging a react native app but i dont remember where to find it

nukeop:
I got the remote redux tools working on firefox with no config at all, it just connects.

PR #326, opened 2019-05-06
chore: add react and redux devtool the proper way

charjac:
looks like it works again !!!! #234


nukeop:
Oh, fantastic. Thanks for handling this.

PR #333, opened 2019-05-21
#193 Add possibility to add track to favorites in Best New Music

sgraczyk:
Related issue: #193 

The current implementation assumes usage of an existing mechanism of showing a popup with actions listed in the title. Yet to be decided if that is fine in UI/UX manner.

image

PS. I still need to adapt my JS code formatter to rules in this project as my VS Code keeps fixing some parts of the code automatically.

**UPDATE 2019.08.08**
Implemented `Add to favorites` from Best Track Item perspective using simple icon.
image


.gitignore
```diff
@@ -65,4 +65,7 @@ bundle.electron.js
 # Prettier configuration file
 .prettierrc
 
-package-lock.json
\ No newline at end of file
+package-lock.json
+
+#editors configuration 
+.vscode/
```
sgraczyk: I might also ignore it locally if that's not expected here

app/components/Dashboard/BestNewMusicTab/BestNewMusicContent/index.js
```diff
@@ -1,8 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import _ from 'lodash';
```
sgraczyk: Unused

app/components/Dashboard/BestNewMusicTab/BestNewMusicContent/index.js
```diff
@@ -17,34 +16,34 @@ const BestNewMusicContent = props => {
   if (item === null) {
     return null;
   }
-  
+
   return (
     <div className={styles.best_new_music_content}>
       <div className={styles.review_header}>
         <div className={styles.thumbnail}>
-          <img alt={ item.title } src={ item.thumbnail }/>
```
sgraczyk: My automatic code formatter integrated with ESLint keep removing those spaces. I can fix that, although I haven't found the appropriate setting for that.

app/components/Dashboard/BestNewMusicTab/BestNewMusicContent/index.js
```diff
@@ -64,23 +63,13 @@ BestNewMusicContent.propTypes = {
   item: bestNewItemShape,
   artistInfoSearchByName: PropTypes.func,
   albumInfoSearchByName: PropTypes.func,
-  addToQueue: PropTypes.func,
```
sgraczyk: These are unused in this component

app/components/Dashboard/BestNewMusicTab/BestNewMusicMenu/BestNewMusicCard/index.js
```diff
@@ -1,39 +1,75 @@
...
         </div>
         <div className={styles.card_artist}>
-          { item.artist }
+          {item.artist}
         </div>
       </div>
     </div>
   );
+
+  const track = {
```
sgraczyk: I had to provide `TrackPopupContainer` with the correct format of track

app/components/Dashboard/BestNewMusicTab/BestNewMusicMenu/BestNewMusicCard/index.js
```diff
@@ -1,39 +1,75 @@
...
+      trigger={card()}
+      track={track}
+      artist={item.artist}
+      title={item.title}
+      thumb={item.thumbnail}
+    />
+  ) : card();
 };
 
+export const bestNewItemShape = PropTypes.shape({
```
sgraczyk: Moved the base shape due to errors occurring when the shape was defined in a parent component. The shape was used before it was initialized, now everything seems to be fine.

app/components/Dashboard/BestNewMusicTab/BestNewMusicMenu/BestNewMusicCard/index.js
```diff
@@ -1,39 +1,75 @@
...
+});
+
 BestNewMusicCard.propTypes = {
   item: bestNewItemShape,
-  onClick: PropTypes.func
+  onClick: PropTypes.func,
+  withPopupMenu: PropTypes.bool
+};
+
+BestNewMusicCard.defaultProps = {
```
sgraczyk: Added missing `defaultProps`

app/components/Dashboard/BestNewMusicTab/BestNewMusicMenu/index.js
```diff
@@ -33,15 +33,16 @@ const BestNewMusicMenu = props => {
...
           props.tracks.map(track => {
             return (
-              <Menu.Item link key={ track.title }>
+              <Menu.Item link key={track.title}>
                 <BestNewMusicCard
-                  item={ track }
-                  onClick={ () => props.setActiveItem(track)}
+                  item={track}
+                  onClick={() => props.setActiveItem(track)}
+                  withPopupMenu
```
sgraczyk: Show popup for tracks

nukeop:
Thanks for the contribution and the overall cleanup, much appreciated. I shared some ideas for improving this even more in the original issue thread. You could implement those, if you feel like it, or I could merge your PR right away and then somebody else (maybe me) could follow up with more changes. Up to you.

nukeop:
Btw, are you with Livechat? I applied there for a job once.

sgraczyk:
I like your idea without the popup, It was my initial one also but I followed a pattern found in the source code. I'll implement the play, add to queue/favorites, download functionalities as discussed in the issue.

And yup, I'm working at LiveChat on a daily basis right now.

nukeop:
Hi again, are you still working on this or can this be improved/finished by somebody else?

sgraczyk:
Hi @nukeop . Sorry for the late response. I did continue the work as discussed nearly a month ago, but had to stop due to other responsibilities. I'm really sorry for not mentioning that earlier. 

The topic can be finished by someone else as I'm going on holidays tomorrow and won't be available for the next month. 

After the break, I'm still willing to contribute as this application is a really interesting topic for me.

A short comment on the issues I've met during my work.
* It's hard to determine if track from `Best` section is a favorite one as we're missing common identifier. AFAIR most of the tracks stored in a redux store can be identified by and `id` from last.fm or discogs (don't remember which one). `Best` tracks are comming from some other source which makes it hard to determine if it's a favorite.
* I decided to simply compare `artist - track name` which is a little bit clunky solution but it works. I created a selector for that, but as it operates on an array it is a little bit non-optimized.
* `Best` track endpoint tend to return track names wrapped with additional `"` so a trimming is expected in the implementation.

nukeop:
Thanks for the comments! I think I will finish what you started before you're back, but I'm sure there will be lots of stuff to do then too, if you're still interested in contributing.

app/actions/favorites.js
```diff
@@ -16,9 +17,13 @@ export function readFavorites() {
 
 export function addFavoriteTrack(track) {
   const favorites = store.get('favorites');
-  favorites.tracks = _.concat(favorites.tracks, track);
+  const newTrack = {
+    ...track,
+    uuid: uuidv4()
```
sgraczyk: I've noticed the track is removed from favorites by `uuid` but usually when it's added it's missing that identifier. I've decided to add it with each `addFavoriteTrack` action as it helped me in converting best track item to a favorite one.

app/components/Dashboard/BestNewMusicTab/BestNewMusicMenu/BestNewMusicCard/index.js
```diff
@@ -1,40 +1,110 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import { Icon } from 'semantic-ui-react';
 
+import { removeQuotes, createLastFMLink } from '../../../../../utils';
+import { favoriteTrackShape } from '../../../../../constants/propTypes';
 import styles from './styles.scss';
 
+function toFavoriteTrack({ artist, title, thumbnail }) {
```
sgraczyk: best item has totally different structure. In order to add it to favorites we need to convert it.

app/components/Dashboard/BestNewMusicTab/BestNewMusicMenu/BestNewMusicCard/index.js
```diff
@@ -1,40 +1,110 @@
...
+    image: ['small', 'medium'].map(size =>
+      ({
+        '#text': thumbnail,
+        size
+      }))
+  };
+}
+
+const FavoriteIcon = ({ isFavorite, onClick }) =>
+  <Icon className={styles.card_favorite} name={isFavorite ? 'heart' : 'heart outline'} size='large' onClick={onClick} />;
```
sgraczyk: I'm having issues with `star outline` `Icon`. For me, no version of `semantic-ui-react` provides the font with an outline version of `star`. I'd be grateful if someone could check if it's only my problem. Used `heart` icon temporarily.

app/selectors/favorites.js
```diff
@@ -0,0 +1,4 @@
+export function getFavoriteTrack(state, artist, title) {
```
sgraczyk: Getting favorite track by `artist` and `title`.

app/selectors/favorites.js
```diff
@@ -0,0 +1,4 @@
...
+  return state.favorites.tracks
+    .find(track => track.artist.name === artist && track.name === title);
```
sgraczyk: I'm considering simply using `toLowerCase()` here or more advanced comparison with 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare

In the second option, the main question is how we should treat the accents. `'a' = 'á'`?

app/containers/DashboardContainer/index.js
```diff
@@ -10,21 +10,35 @@ import * as ToastActions from '../../actions/toasts';
...
-    musicSources={musicSources}
-  />
-);
+class DashboardContainer extends React.Component {
+  constructor(props) {
+    super(props);
+  }
+
+  componentDidMount() {
+    this.props.actions.readFavorites();
```
sgraczyk: If we want to know if some tracks are among favorites, we need to fetch them in DashboardContainer.

app/containers/BestNewMusicCardContainer/index.js
```diff
@@ -0,0 +1,49 @@
...
+
+BestNewMusicCardContainer.defaultProps = {
+  item: null,
+  onClick: () => { }
+};
+
+function mapStateToProps(state, ownProps) {
+  return {
+    favoriteTrack: ownProps.item
+      ? getFavoriteTrack(state, ownProps.item.artist, removeQuotes(ownProps.item.title))
```
sgraczyk: Here I'm removing the quotes that are coming from `pitchfork`. We can consider removing the directly after fetching data to avoid any future problems.

app/constants/propTypes.js
```diff
@@ -0,0 +1,18 @@
...
+    name: PropTypes.string
+  }).isRequired,
+  url: PropTypes.string,
+  image: PropTypes.arrayOf(PropTypes.shape({
+    '#text': PropTypes.string,
+    size: PropTypes.oneOf('small', 'medium', 'large', 'extralarge')
+  }))
+});
+
+export const favoriteTrackShape = Object.assign(trackShape, PropTypes.shape({
```
sgraczyk: Some shapes I assumed would be useful

sgraczyk:
@nukeop I've decided to do #193 in small steps to make it easier to review. It would be nice if you could take a look. I've left some comments to describe some of the issues I met.

app/actions/favorites.js
nukeop: Uuids for favorite tracks were not generated, because they used to be added only from the queue, and the queue adds uuids to its items when they're added to it:

https://github.com/nukeop/nuclear/blob/4566bc2ae6f57a1200e07398e045a22c857a0350/app/actions/queue.js#L19

Now that you're implementing adding favorite tracks from other sources, I suppose this will be necessary. What is important to keep in mind is to clone the track on adding so that we do not mutate the state and mess with the queue's ids, so what you did is a good way to handle this.

app/components/Dashboard/BestNewMusicTab/BestNewMusicMenu/BestNewMusicCard/index.js
nukeop: Nice, I couldn't figure out how to do it.

app/constants/propTypes.js
nukeop: Yep, great

app/selectors/favorites.js
nukeop: Yep, use `_.deburr`

app/utils.js
```diff
@@ -34,10 +34,26 @@ export function stringDurationToSeconds (duration) {
...
   return typeof selectedStream === 'undefined'
     ? streams ? streams[0] : null
     : selectedStream;
 }
+
+export function removeQuotes(text) {
+  return String(text).replace(/[“”]/g, '');
+}
+
+export function createLastFMLink(artist, track) {
```
nukeop: Should be tested with at least few examples of tricky artist names so we're sure it'll cover most of the typical cases

nukeop:
I finally got around to reviewing. Nice work overall, I only have a couple of suggestions.

app/utils.js
sgraczyk: Add 2 more advanced cases to tests

app/selectors/favorites.js
sgraczyk: Done

app/components/Dashboard/BestNewMusicTab/BestNewMusicMenu/BestNewMusicCard/index.js
sgraczyk: Resolved by using `disabled` prop of `Icon` component which renders favorite/non-favorite as follows:

image

sgraczyk:
Suggestions applied.

nukeop:
Ready to merge?

sgraczyk:
> Ready to merge?

Ready

sgraczyk:
Conflicts resolved

PR #342, opened 2019-06-04
fix(local): fetch missing metadata from acousticId api

charjac:
for now its just a POC. it look like it works on linux but i don't have enough mp3 to be sure neither a windows and mac computer for test (too lazy for a vm :)).


nukeop:
I might try to test it on a vm today.

nukeop:
I got around to this finally. It doesn't build for me on Windows, unfortunately.

charjac:
:'( :'( :'(
Could you copy paste the error log here please ?

I will fight my laziness and try to fix it in a vm \0/.

nukeop:
i will try to reproduce it, but i didn't paste it because it wasn't very descriptive. maybe i'll try running it natively.

charjac:
I try to build nuclear on a windows10 64 family edition, the build was ok, but it fails at running :'(.
Here is the error:

Capture d’écran de 2019-07-01 13-46-02

charjac:
In development i got this:

Capture d’écran de 2019-07-01 16-20-53

looks like the error is triggered by lodash in electron-store package .... 😰😰😰

charjac:
i get the same error when trying to build from master :/

Capture d’écran de 2019-07-01 17-56-45

charjac:
but when i download the portable .exe from release tab it works .......
windows is so fucking drunk :'( :'( :'( 

How and where do you build it ?

nukeop:
I manually patched some stuff in the compiled javascript to get rid of this error, it's not viable long term.
This great contributor patched the third party package that was the root of this issue (I think): https://github.com/nukeop/nuclear/issues/362#issuecomment-506302785
I'll update package.json.
At least I think that was the related issue.

charjac:
if the package swagger-spec-express is a problem, i can easily replace it with something else.

nukeop:
I rebuilt with the patched version of the package and it indeed looks like the issue is gone.

nukeop:
Another person is contributing an upgrade for the travis config file so that we can have continuously built Windows binaries too: https://github.com/nukeop/nuclear/pull/380

When that's merged we can move to releasing on every push.

charjac:
sounds great !!!
Could you push the new package.json to master so i can rebase and continue my investigation ?

nukeop:
Ok, done.

charjac:
thx

charjac:
finally i found the problem:
it come from app/persistence/store.js

it seems like electron-store doesn't like to set a key with an undefined value on windows.....

it still dont run with wine (not really surprised) but has a new error \0/, i try on a vm later today.

charjac:
I think we are good now.

charjac:
for now the acousticId licence is hard coded and associated with my email address.
I think you should create your own and then we should think about where we put it (environment ? settings ?).

nukeop:
Normally I put all the keys in `globals.js`. I can register one on nuclear@gumblert.tech, which serves as the official email address for Nuclear.

charjac:
This key is intended to be used in the node.js process so for now we could leave it here.

I was thinking about a great refactor of the server part.
With all the recent new feature, it become a little messy.
What do you think ?

nukeop:
Yeah, definitely. I'll need to unify the server file so that it detects the environment (dev/prod) and platform and we don't have to keep the files in sync. Other than that i think the modules are separated pretty well but if you have ideas how to improve that further, feel free to implement them.

charjac:
ok cool as soon as it works normally on windows i go on this.

still buggy on windows ...

charjac:
looks like its ok now

nukeop:
Great, I'm gonna check it out soon, and merge if it works fine.

charjac:
Looks like the rebase has broke something .... I check this later today.

nukeop:
I refactored that bit of the code that was broken earlier. If this broke your changes I'll try to fix it myself.

charjac:
ok its nothing.
Actually the library view can only works if api is enabled.

We can:
  - add a message in the view if api is not enabled
  - separate the part of the api wich is needed for the library to work

nukeop:
I think 1) is more realistic.

charjac:
ready for review :)

package.json
```diff
@@ -69,9 +71,12 @@
     "moment": "^2.20.1",
     "mousetrap": "^1.6.2",
     "music-metadata": "^3.5.3",
+    "node-fetch": "^2.6.0",
```
nukeop: You don't need that, we have isomorphic-fetch already.

package.json
```diff
@@ -69,9 +71,12 @@
...
     "nuclear-core": "0.0.5",
     "nuclear-ui": "0.0.6",
     "numeral": "^2.0.6",
+    "once": "^1.4.0",
```
nukeop: Is this used anywhere? In any case it's better to use lodash's `_.once`

server/lib/platform.js
```diff
@@ -0,0 +1,17 @@
+import { platform } from 'os';
```
nukeop: You don't need that, we already have `electron-platform` (https://www.npmjs.com/package/electron-platform). And it has all the platform detection functions you need. I think this also gets rid of the `os` dependency.

nukeop:
Looks good, I wonder how this work in practice. How fast is it? Maybe it'd be worthwhile to provide an option to skip all unidentified tracks instead of looking up their metadata?

package.json
charjac: mmmh i think it was stream related, and i finally remove it during a refacto

charjac:
I dont know, Actually the meta are cached in the store, so if the user doesn't move his files the meta are only fetched once (if no meta present in the file).

charjac:
thx ! I start working on the server refactor

Issue #336, opened 2019-06-01
Add current track to favorites/playlist - bug if the queue is empty


nukeop:
If the queue is empty this function will add a garbage empty track. Needs to be solved.

mallowcat:
I'll work on this if you want to assign to me  (=^ω^=)

nukeop:
Sure, feel free to take this issue. I can't assign you since we can only assign maintainers with full access to the repo, but I'm sure nobody else will work on this.

mallowcat:
Fixed by not allowing tracks without least the name defined. Wasn't sure if there are scenarios where the artist is left undefined? I could add a check for that too if the artist name is always expected.

https://github.com/nukeop/nuclear/pull/343

nukeop:
Closed by #343

PR #343, opened 2019-06-04
Resolving 'Add current track to favorites/playlist' #336

mallowcat:
Fixed by not allowing tracks without names defined.


nukeop:
Awesome, thanks for the bugfix.

PR #347, opened 2019-06-07
fix: fix typo due to last refactor

charjac:
fix#346 
I think i catch 'em all :)

### DIFF
--- a/app/components/AlbumCover/AlbumInfo/index.js
+++ b/app/components/AlbumCover/AlbumInfo/index.js
@@ -2,16 +2,16 @@ import React from 'react';
 
 import styles from './styles.css';
 
-const AlbumInfo = () => (
+const AlbumInfo = ({ artist, title }) => (
   <div className={styles.album_info_container}>
     {this.props.nameOnly ? (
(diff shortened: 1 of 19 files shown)


nukeop:
awesome, thanks for a quick fix. I like this style of destructuring props right in the function arguments, I think I will use it from now on.

charjac:
:) dont merge it right now i forgot to add new french translation

nukeop:
Yeah I noticed. I will wait until you're finished.

nukeop:
Btw, do you know if there's a way to check if a translation is complete when compared to the list of english strings? We could add a check like that to the tests too.

charjac:
good idea !! jest + enzyme seems to be the good test stack or do you prefer mocha ?

charjac:
i think we are good now

nukeop:
We already have enzyme and mocha along with some really basic tests. Feel free to add Jest for new tests if you feel more comfortable with it though, I don't think it's going to mess anything up.

charjac:
no mocha is cool (and fast) and tha api are almost the same.

charjac:
its ready to merge (i think)

Issue #345, opened 2019-06-05
Translations (looking for contributors)


nukeop:
We're looking for contributors who are willing to translate Nuclear to their native languages (or any languages that they know well). Currently @charjac has done all the work required to add new translations, and also translated all the strings to French.

New locales should be placed here: https://github.com/nukeop/nuclear/tree/master/app/locales

**Languages currently available:**
- English
- French (by @charjac)
- Dutch (by @Vistaus)
- Danish (by @Hansen1992)
- Spanish (by @mlucas94 )
- Polish (by @kazimierczak-robert and @gradzka)

**Proposed:**
- Chinese (simplified) (by @lizzyd710)
- Russian (by @ramstore07)

charjac:
When adding translations, a few things need to be done:
 - add the json file with the correct name in app/locales and keep it empty
 - update the i18n script in package.json by adding your locale 
```
{
  ...
  "scripts": {
    ...
     "i18n": "sync-i18n --files 'app/locales/*.json' --primary en --languages fr [your locale] --space 2"
  } 
  ...
}
```
 - run the command
```
npm run i18n
```

you can now translate your locale file from english to your language.

Hansen1992:
Hey @nukeop, I would like to help you translate Nuclear from English to Danish if this has any interest?

nukeop:
Absolutely, in general if you don't see a language on the list above, we want it added. Thanks for the help.

lizzyd710:
What kind of things or phrases would you like to be translated? I'm a native English speaker but I took a few years of Chinese (Mandarin, so I write simplified). I may not be able to do much, but I would be happy to start and do as much as I can.

nukeop:
You can see a list of all strings that have to be translated here: https://github.com/nukeop/nuclear/blob/master/app/locales/en.json

Even if you only translate a part of it, any missing strings fall back on default English. And some is better than nothing.

lizzyd710:
> 
> 
> When adding translations, a few things need to be done:
> 
>     * add the json file with the correct name in app/locales and keep it empty
> 
>     * update the i18n script in package.json by adding your locale
> 
> 
> ```
> {
>   ...
>   "scripts": {
>     ...
>      "i18n": "sync-i18n --files 'app/locales/*.json' --primary en --languages fr [your locale] --space 2"
>   } 
>   ...
> }
> ```
> 
>     * run the command
> 
> 
> ```
> npm run i18n
> ```
> 
[...]

PR #350, opened 2019-06-09
fix(i18n): add missing code for nl locale (my bad)

charjac:
I forgot one file #345 (now its ok)


charjac:
there where also a problem with history in some components including SearchResult, to fix it i simply wrap all component wich use history with the withRouter decorator

PR #351, opened 2019-06-11
Added sort by name in Downloads

saimfd:
Added sort functionality in downloads section. List is sorted by clicking on Name header. Works in both ascending and descending order of Song Name.


nukeop:
This is great, thank you for the contribution.

PR #354, opened 2019-06-14
Polish translation

kazimierczak-robert:
[#345](https://github.com/nukeop/nuclear/issues/345)
Polish translation by @kazimierczak-robert and @gradzka


nukeop:
Thank you, this is great.

Issue #356, opened 2019-06-16
Program doesn't start if the http API port is in use


nukeop:
Described in more detail here: https://github.com/nukeop/nuclear/issues/337#issuecomment-502473060

To prevent this bug, we could use this package:
https://www.npmjs.com/package/get-port

And inform the user about the new port in the settings section.

schippas:
Hello, I am interested in working on this project, Is anyone else working on this issue already, or can I work on it?

nukeop:
You can take it, I haven't started working on it yet.

leonlafa:
Are you going to look into this @schippas?  If not I can have a PR in the next 30 mins or so...

PR #357, opened 2019-06-18
the api now only uses ports that are available

leonlafa:


leonlafa:
PR made for issue:
https://github.com/nukeop/nuclear/issues/356

nukeop:
Thanks, great contribution.

leonlafa:
> Thanks, great contribution.

Currently has a choice of 10 ports to choose from, we can extend this functionality if needed.

nukeop:
I think that will fix the problem we had. Looks like it builds, time to merge.

leonlafa:
> I think that will fix the problem we had. Looks like it builds, time to merge.

Brilliant, I'll keep my eye out for other issues

Issue #358, opened 2019-06-18
Improve language selection UX


nukeop:
The dropdown for selecting languages needs 2 changes:
- Change the styles (possibly using the dropdown from Semantic UI) so that the look & feel is similar to the rest of the program
- Display user friendly names for available languages

schippas:
I am looking into it right now.

nukeop:
Brilliant! I made this more as a reminder to myself when I get home, but if you want to try fixing this, even better.

nukeop:
Finished.

PR #359, opened 2019-06-18
Display user friendly names for available languages

schippas:
This is for issue #358


nukeop:
You sure are fast. Does this also replace the dropdown with a nicer one?

schippas:
Well, I replaced the React Select dropdown with the Semantic UI Dropdown, but it looks about the same. Other than that, I didn't change its appearance.

PR #372, opened 2019-06-22
Play song when it's doubleclicked (TrackRow)

olafur-palsson:
By injecting a prop into TrackPopupContainer via TrackRow.

Sorry for verbose commit, I accidentally did auto-indent with vim. :(

### DIFF
--- a/app/components/TrackRow/index.js
+++ b/app/components/TrackRow/index.js
@@ -18,7 +18,7 @@ class TrackRow extends React.Component {
     this.playTrack = this.playTrack.bind(this);
   }
 
-  getThumbnail(track) {
+  getThumbnail (track) {
     return _.get(
       track,
       'image[1][\'#text\']',
@@ -54,7 +54,7 @@ class TrackRow extends React.Component {
     this.setState({ doubleClick: true });
   }
 
-  renderTrigger(track) {
+  renderTrigger (track) {
     const {
       displayCover,
       displayTrackNumber,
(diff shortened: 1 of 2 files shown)


nukeop:
Can you run eslint on this? Might take care of the formatting.

olafur-palsson:
Shit I'm horrible at this git business. I'm going to study the docs more. Thanks for putting up with the mess. :)

app/components/TrackRow/index.js
```diff
@@ -43,6 +49,11 @@ class TrackRow extends React.Component {
     );
   }
 
+  playTrack () {
+    console.log(this)
```
nukeop: Could you remove this?

app/containers/TrackPopupContainer/index.js
```diff
@@ -23,13 +23,33 @@ const TrackPopupContainer = props => {
...
+          name: props.title,
+          thumbnail: props.thumb
+        });
+      actions.selectSong(0);
+      actions.startPlayback();
+    }
+  };
+
+  if (doubleClick) {
+    playSong();
```
nukeop: Anything that dispatches actions should stay out of the render method, and only do so in response to some event.

nukeop:
Sorry it took me a few days to get back to you. Your solution works, but it should be structured differently. The TrackRow should dispatch the actions necessary to start playing the song, since it responds to the double click event. Like I commented in the code, action logic should stay out of render methods, due to the risk of being launched at unexpected times when React decides to re-render the component.

I'm thinking that maybe we should get rid of the popup appearing when you click anywhere on the track row, maybe there should be a dedicated button (with a vertical ellipsis) at its edge, so we could get clean double clicks without opening the popup.

olafur-palsson:
Sorry, my master branch was polluted. I will migrate this to another pull request.

PR #375, opened 2019-06-24
Minor change to .eslintrc.js

olafur-palsson:
I noticed a couple of inconsistencies in style. Majority of code passes lint with this option added.


nukeop:
I think we've been using spaces inside curly braces in many places, especially in props. I don't remember many cases of the object shorthand syntax though. Does this cause many changes if you run eslint with --fix?

nukeop:
Ok, I can agree with those changes. Are you still working on something more or can I merge this?

olafur-palsson:
It's done now. Can merge. :)

app/actions/index.js
```diff
@@ -41,9 +41,7 @@ export function sourcesSearch (terms, plugins) {
     Object.assign(searchResults, plugins.musicSources[i].search(terms));
   }
 
-  return {
-    type: SOURCES_SEARCH
-  };
+  return { type: undefined };
```
nukeop: I think this will introduce a bug.

app/actions/toasts.js
```diff
@@ -48,6 +49,7 @@ function generateNotification(title, details, icon, type, settings) {
     },
     type)));
 
+    // BUG: _ is undefined
```
nukeop: Lodash inserts its _ into global scope.

app/actions/toasts.js
```diff
@@ -37,6 +37,7 @@ export function info(title, details, icon, settings) {
   return generateNotification(title, details, icon, {info: true}, settings);
 }
 
+/** @deprecated */
```
nukeop: This shouldn't be deprecated...

app/actions/index.js
nukeop: Nvm, actually this function can probably be deleted altogether.

olafur-palsson: Yeah, I wasn't sure what it was going to be user for so I just exposed the bug.

olafur-palsson:
This one is ready.

PR #376, opened 2019-06-25
Refactor App.js a bit without major changes.

olafur-palsson:
Basically I was refactoring App.js but sometimes I wanted to remove parameters so I went to other files.


app/App.js
```diff
@@ -283,15 +281,15 @@ class App extends React.Component {
...
+  renderVolumeControl () {
     return (
       <VolumeControls
         fill={this.props.player.volume}
         updateVolume={this.props.actions.updateVolume}
         muted={this.props.player.muted}
         toggleMute={this.props.actions.toggleMute}
         toggleOption={this.props.actions.toggleOption}
-        settings={settings}
+        settings={this.props}
```
nukeop: Is this correct?

app/components/NavButtons/index.js
```diff
@@ -21,16 +21,16 @@ class NavButtons extends React.Component {
   }
 
   render() {
-    let { back, forward, historyLength, historyCurrentIndex} = this.props;
-
+    const history = this.props.history;
```
nukeop: Instead of this, we could use `withRouter` to enhance this component with history directly.

app/components/Spacer/index.js
```diff
@@ -1,7 +1,4 @@
 import React from 'react';
+import styles from './styles.scss';
 
-const Spacer = ({ style }) => (
-  <div style={Object.assign({}, { flex: '1 1 auto' }, style)} />
-);
-
-export default Spacer;
+export default () => (<div className={styles.spacer}/>);
```
nukeop: In some cases we want to pass down the style prop.

nukeop:
Thanks for the quality improvements. I made some comments after reviewing the code, we can merge after we resolve them.

app/App.js
olafur-palsson: Nope, breaks the equalizer. Should be `this.props.settings`.

app/components/Spacer/index.js
olafur-palsson: You're right, even though we have no current cases of it being used that way.

(App.js had one case which I refactored out to the corresponding scss).

app/components/NavButtons/index.js
olafur-palsson: Fixed.

app/components/Spacer/index.js
nukeop: Ok, in this case it can stay like this, thanks.

nukeop:
Thanks for the changes, I will resolve merge conflicts today and then I will merge everything.

PR #378, opened 2019-06-28
[WIP] Button to delete tracks from playlist

Hansen1992:
Hey nukeop, I made some changes in my effort to create a delete function that deletes tracks. So far it's not working but I'm pretty sure I've done something right some places. I tried to create a function just like the function that delete the entire playlists by mimicking some of the code from the "deletePlaylist" function. I just don't know, if that is even the right thing to do.

I appreciate the help and hope that the adding I made is not all bad.


nukeop:
Actually, all you need is the updatePlaylist function, since it modifies an existing playlist by replacing it with a new version of itself. I will show you an example of what I mean a bit later.

nukeop:
https://github.com/nukeop/nuclear/blob/b178551b56b290e048cc247561c37c8fc69a5a89/app/components/PlaylistView/index.js#L45

If you look at this function, renaming is accomplished by cloning the playlist, changing the name of the new copy, and calling updatePlaylist. The old playlist will be replaced with this copy, because they have the same id.

To delete a track from a playlist, you should create a similar function, that instead of changing the name of the cloned playlist will filter its tracks (just don't use `_.remove`, on the original playlist, because it mutates the state, and that's a big no-no in Redux). An example of what the filtering could look like:
```javascript
updatedPlaylist.tracks = _.filter(updatedPlaylist.tracks, track => track.id !== trackToRemove.id )
```

BTW. This workflow makes me think it'd be a good idea to prepare a couple of simple functionalities to implement for people who are learning and would like to contribute to some open source project but who don't feel too confident yet, so that they can solve a real life problem while being guided in a friendly way, both learning and contributing.

Hansen1992:
Thanks for the guidance, I'll try my best to implement the above example-code into a delete-track function! 

To answer your above-mentioned idea about making some issues that contain a couple of simple functionalities to implement for people who are learning and would like to contribute to some open source project:
I'm currently working as a curator in Denmark but while taking some upgrading IT courses, I was absolutely overwhelmed with awe when I found out what you could do with code. 
I'm currently preparing myself and my skills for searching for jobs in the software-developing industry and none of my friends know anything about code, so for you to help me solve this issue in your app means the world to me. You not only learn how to decode another persons work but you also learn how things are connected and mapped in a framework like react. 
I'm sure others, who are in a similar situation as I'm in, and are trying to detach themselves from youtube tutorials and actually throw themselves at a real issue that needs to be solved for a real app will benefit greatly from it just as it has benefitted me.

nukeop:
It was right to have `removeTrack` have `id` as its argument. That's the information by which you can distinguish the tracks on the playlist. You don't need to touch `actions/playlists.js`. The place for this function is in `PlaylistView/index.js`, in the `PlaylistView` class.

Take a look how `renamePlaylist` updates the playlists (starts here: https://github.com/nukeop/nuclear/blob/620700a733e376d3e5b51b5f8e84e1c79f0b3d47/app/components/PlaylistView/index.js#L45).
You have to do something similar, except instead of changing the name you are going to change the list of the tracks. As a hint, you're going to need the playlist and the track id as arguments.

The steps that you'll need to take are:
* clone the original playlist
* filter out the track clicked by the user
* replace the cloned playlist's track list with the filtered one
* update the playlist using the clone

And then after you have this function, you can connect it to the button like you already did (by supplying it in the `onDelete` prop of `TrackRow`).

Hansen1992:
Thanks, I'll give it my best shot!

nukeop:
You almost got it. Just remember that you don't have to modify anything in actions/playlists.js to make this work. You can revert your changes there. Maybe I'm explaining this too chaotically. I'll try to explain it better in a bit.

Hansen1992:
Hi nuke, I'm sorry it's been a while since I last made a pull request, I've been on vacation. 
I deleted the code I had made in actions/playlists.js, so this should now be in order.

nukeop:
No problems, we don't have deadlines here. You can contribute at your own pace. I'll test your code now.

Hansen1992:
The code is still not working, but I'm pretty sure I'm getting close.

Hansen1992:
Hey @nukeop  this should be the final and working code if I'm not wrong :-)

nukeop:
Yes, I think you've nailed it. I'll verify so that we're 100% sure.

Btw, this could be a good opportunity to add a test for this function too. I'll check if I can add code to your branch and if not I'll add it in a comment and ask you to commit it, if you agree.

nukeop:
I noticed that you need to add a new field to the track table header at the beginning.

2019-08-06-011506_795x134_scrot

nukeop:
I'll add the tests after you correct this small issue and merge everything, so you don't have to worry about it.

nukeop:
https://github.com/nukeop/nuclear/pull/432/commits/4ebaf745ca3c7650790cf5b56578049b81fb1fd5

This is how I intend to test the updatePlaylist function after your changes, for reference. I implemented dependency injection for deletePlaylist here, updatePlaylist will be refactored similarly.

Hansen1992:
Hi Nuke, I'm experiencing some issues with opening the app on my computer common to the error I've had before with only the logo of the app appearing when running the app in the browser. Even though I am pretty sure I can code the last bit without looking at the result of the code in the app, I still want to make sure it works before I push it to Github. I just want to make you know, that I haven't forgot about this issue and that i'll soon be pushing the final bit of code up :-)

nukeop:
No problem, take as much time as you need.

Hansen1992:
Hi nuke, do I need to install docker now in order to run the dev part of the app? 
I previosly runned the app from gitbash, can this still be done?

nukeop:
I've changed the architecture in a major way recently. You can either try using docker (it's just one command to run dev mode now), or you could clone the repo, cd to `packages/app`, and run the program the old way (`npm run watch`, and `npm run electron:dev` in a separate shell).

PR #379, opened 2019-06-29
Doubleclick

olafur-palsson:
Here it is, the doubleClick feature implemented in the way that you requested. :)


app/components/TrackRow/index.js
```diff
@@ -43,6 +55,15 @@ class TrackRow extends React.Component {
     );
   }
 
+  playTrack () {
+    console.log(this.getThumbnail(this.props.track));
```
nukeop: could you remove this debug statement?

nukeop:
Thanks, looking good. Just one small change and it'll be ready to merge.

app/components/TrackRow/index.js
olafur-palsson: It is

app/components/TrackRow/index.js
```diff
@@ -1,18 +1,30 @@
...
 import TrackPopupContainer from '../../containers/TrackPopupContainer';
 import { formatDuration } from '../../utils';
 
 import styles from './styles.scss';
 
 class TrackRow extends React.Component {
-  getThumbnail(track) {
+  constructor () {
+    super();
+    this.state = { doubleClick: false };
```
nukeop: I think we won't need this in this component anymore, and it can be converted to a stateless one.

nukeop:
One more thing I just noticed. I also tested the code and it works well.

olafur-palsson:
Should be good now.

nukeop:
Great, thank you.

nukeop:
Your change broke the functionality, since `this` is unbound in `playTrack`. You'll need to bind it.

olafur-palsson:
Sorry for taking long, I was on a vacation in Spain with no laptop.

nukeop:
No problem, we don't have deadlines here. Thanks for the contribution.

PR #380, opened 2019-06-30
Windows build

tusharchitti:
I made change to travis.yml. I also added a command in package.json : build:dist:windows.
Please review.

### DIFF
--- a/package.json
+++ b/package.json
@@ -149,6 +149,7 @@
     "svg-inline-loader": "^0.8.0",
     "url-loader": "^1.0.1",
     "utf-8-validate": "^5.0.2",
+    "webpack-sources": "1.0.1",
     "webpack": "^4.12.1",
     "webpack-dev-server": "^3.1.4"
   },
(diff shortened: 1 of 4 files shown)


webpack.config.prod.js
```diff
@@ -17,7 +17,7 @@ const config = {
     fs: "empty"
   },
   mode: 'production',
-  devtool: 'source-map',
+  devtool: process.env.NODE_ENV === "development" ? "source-map" : "none",
```
nukeop: Since this is  only used in production, we can leave this as "source-map". According to dev tools it's safe to use source maps in production: https://webpack.js.org/configuration/devtool/

And this way we have more detailed information when encountering bugs.

webpack.config.dev.js
```diff
@@ -25,7 +26,7 @@ const config = {
     publicPath: '/'
   },
   mode: 'development',
-  devtool: 'source-map',
+  devtool: process.env.NODE_ENV === "development" ? "source-map" : "none",
```
nukeop: This can be left as 'source-map' in all builds.

package.json
```diff
@@ -148,6 +150,7 @@
     "svg-inline-loader": "^0.8.0",
     "url-loader": "^1.0.1",
     "utf-8-validate": "^5.0.2",
+    "webpack-sources": "1.0.1",
```
nukeop: I think this package isn't used anywhere, could you remove this dependency?

nukeop:
Thanks for your contribution. I only have a couple of small comments and we'll be ready to merge.

package.json
tusharchitti: yes i will remove this.

Issue #362, opened 2019-06-21
Frequent HTTP 403 errors with YouTube plugin


Mongogamer:
- **Nuclear version**: 0.4.5 commit 02776aa6989a10ff73b72b6d61a707af26ab14e3, fdbdc3a8f27ebba6d58fe4f7aed7aba4b0ba36ae
- **OS**: Windows 10 Pro x64 1903
- **NodeJS version**: 10.16.0
---
**Description of the issue**: The YouTube plugin is currently encountering very frequent HTTP 403 errors when trying to search for results and streaming audio. Sometimes the search does not encounter a 403 and loads the metadata of the track but audio will still 403 and not play. The only way to get audio to play is to switch to either local or Soundcloud source. The problem seems to be caused by the YouTube API quota being exceeded.

**Expected behavior**: HTTP 403 errors obviously aren't intended to happen. Expected bevavior would be a HTTP 200.

---

**Steps to reproduce**:

1. Set YouTube plugin as default music source.
2. Search for any song (I used Blossom by Au5)
3. Attempt to play the song.

---

**Possible solution**: Let users provide their own YouTube API key as an advanced option or automatically switch to Soundcloud whenever a 403 is encountered.

---

[Devtools output when trying to play song while quota is exceeded](https://i.imgur.com/pFHhA25.png)
[Devtools network information for failed request](https://i.imgur.com/nCRk3N4.png)
[Devtools response from YouTube API](https://i.imgur.com/ejvt855.png)

nukeop:
Thanks for a very detailed report. I've seen a similar issue with mpsyt and it was possible to clear the cache to get it to work again. I think the key itself is fine, but we might have to find a way to clear the cache automatically to avoid this.

Mongogamer:
Deleting cache and data inside of electron devtools sadly did not fix the problem for me. Deleting the cache files from the filesystem also did not help. At the moment metadata is loading for me and not throwing 403 errors anymore but playback is still receiving them. Very rarely playback starts working even if the cache is not cleared.

The problem seems to be caused by ytdl-core as I found this issue https://github.com/fent/node-ytdl-core/issues/434 with lots of users complaining about constant 403 errors.

I assume this will be fixed by the developer of ytdl-core soon.

nukeop:
Time to update the package then.

Mongogamer:
The problem is fixed with the new package version. However when starting a production version it throws this error ![](https://i.imgur.com/zV18xOT.png)

[...]

PR #382, opened 2019-06-30
Allow for custom Youtube search API keys via setting

df-andrews:
Issue:
A number of 403 errors do to API key access limits on the youtube search API. These are reproducible using the released version of the app across all platforms. This is however an issue that can be inconsistent due to the nature of an API request limit and how that limit is reset.

Related Issue:
#362

Fix:
I added a setting to the settings page for the youtube API key. It defaults to the key that exists in the globals file..

### DIFF
--- a/app/constants/settings.js
+++ b/app/constants/settings.js
@@ -1,3 +1,4 @@
+import globals from '../globals';
 import settingType from './settingsEnum';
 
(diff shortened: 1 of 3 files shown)


nukeop:
Awesome, thank you for your help.

Issue #396, opened 2019-07-10
Containers share common code.


olafur-palsson:
Majority of the lines that are unique are found in render functions, which is a relatively small part of each file.

nukeop:
This is a consequence of the organic growth of the codebase. Containers also need to have their mapStateToProps and mapDispatchToProps, which are going to end up pretty similar most of the time. Can you provide any examples of code that could be refactored?

olafur-palsson:
I'll start working on a prototype container when I get back to my laptop. (I'm on vacation but I wanted to add this to my todo-list)

nukeop:
Cool, keep me updated.

nukeop:
We've refactored the code in a huge way since then, I think this can be closed now.

PR #399, opened 2019-07-12
Containers

olafur-palsson:
Here is more or less what I was thinking. Apply to rest of containers?


olafur-palsson:
Ref  #396

nukeop:
Your change will give all containers all the actions. I think we need to go in the opposite direction, to only give the containers the action they need, and to separate them instead of keeping them all in one huge object (bad design decision I made at some point). 

If you want to reduce the amount of boilerplate, we should look at what functionalities are similar across many containers, and isolate them in higher order components. E.g. we could create a `withQueue` HOC that would provide access to the queue and add some commonly used methods for manipulating it (go to next item, go to previous item, add item, remove item, etc.). I don't know if that in particular is needed, just an example.

olafur-palsson:
I agree, ideal code is both never repeated and only exposed to methods it needs. My interpretation was that containers are used to limit the scope of their components.

I'm going to think about this and see if I can come up with something. :)

PR #400, opened 2019-07-12
I removed a couple of unused and/or <div/> like components

olafur-palsson:


app/App.js
```diff
@@ -224,7 +224,7 @@ class App extends React.Component {
 
   renderFooter () {
     return (
-      <Footer className={styles.footer}>
+      <div className={styles.footer}>
```
nukeop: Better to use the footer tag here for semantics

app/App.js
```diff
@@ -218,13 +218,13 @@ class App extends React.Component {
...
         </a>
-      </div>
+      </footer>
     );
   }
 
   renderFooter () {
     return (
-      <div className={styles.footer}>
+      <footer className={styles.footer}>
```
olafur-palsson: I also changed the 'sidebar_footer' for consistency

app/App.js
olafur-palsson: Done.

nukeop:
There is one change that I don't understand:
 app/components/Footer/styles.css → app/containers/wrapComponent.js 

Shouldn't it just be deleted?

PR #401, opened 2019-07-12
linting finished, also added 'test' folder to lint process

olafur-palsson:
This version passes linting.


nukeop:
We don't need to import lodash everywhere, because once loaded it's inserted into the global scope.

nukeop:
same for mocha

olafur-palsson:
My plans with this PR is to make the code show no linting errors. With this PR and #402 we can enforce linting through Travis.

I think the linting should always show no errors and this causes a linting error. Since the fix doesn't affect performance I think the benefit outweighs the cost. What do you think?

PR #402, opened 2019-07-12
Made changes to enforce linting

olafur-palsson:
Linting currently behaves like an extra feature rather than a rule of conduct. I think it should be both in `travis` and `webpack`

#401 allows to start with a clean slate.


nukeop:
With #401 merged, could you try merging from upstream to see if the build will succeed now?

PR #408, opened 2019-07-15
Remove explicit Lato font family

karevn:
Semantic UI library used defines its own CSS font stack, which has sensible fallbacks for Russian and other languages, so without these declarations at least the search bar looks way better.

### DIFF
--- a/app/app.global.scss
+++ b/app/app.global.scss
@@ -6,7 +6,6 @@ html {
   height: 100%;
   cursor: default;
   -webkit-user-select: none;
-  font-family: 'Lato';
 }
 
 body {
--- a/app/app.global.scss
+++ b/app/app.global.scss
@@ -6,6 +6,7 @@ html {
   height: 100%;
   cursor: default;
   -webkit-user-select: none;
+  font-family: Lato, 'Helvetica Neue', Arial, Helvetica, sans-serif;
 }
 
 body {
--- a/app/components/SearchBox/styles.scss
+++ b/app/components/SearchBox/styles.scss
@@ -44,9 +44,7 @@
   outline: none;
   background-color: $background3;
 
-  font-family: Lato;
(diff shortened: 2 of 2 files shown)


nukeop:
I think Semantic UI uses the same font.

karevn:
Yes, it does. But it contains a font stack instead of this single font, like this:
```css
font-family: "Lato", "Helvetica Neue", Arial
```

It makes Semantic UI's variant more reliable for systems having no Lato font installed.

nukeop:
It's always going to be available, because it's always linked by Semantic UI. And this way the font is uniform even for components that do not use Semantic UI.

karevn:
Here is what happened to the search bar before my change:
Screenshot 2019-07-16 01-16-08

Please note the font difference to the rest of the screen. That happened because the line:
```css
font-family: 'Lato';
```
told Electron: "here you must use Lato; if it's unavailable, use the default font (serif)"
And having this line removed, it uses a wider rule mentioned above, which tells Electron:
"Use Lato; if it's unavailable - use Helvetica Neue, if it's unavailable - use Arial", so it ends up with more suitable fonts.

nukeop:
Yes, I'm aware. But this shouldn't have changed anything visually. Since the font is available in the sidebar, and the best new music component, it should be available in the search bar. So there's a different bug here.

I think it's the lack of quotes in search bar css. So while it's fine to remove that, I'll ask you to restore the line you deleted from `app.global.css`, since it applies the font to the non-semantic UI components. This way the font will be uniform across the whole app.

karevn:
Totally agree on the styling of the rest of the app. So, I think it's best to replace the "project-wide" and make it the match Semantic UI's declaration. It should guarantee the consistency for the whole project.

nukeop:
Cool, thanks for the improvement.

PR #411, opened 2019-07-18
Babel 7, babel cache, eslint for hooks

karevn:
1. Babel 7
2. Faster builds due to Babel cache enabled
3. Eslint config for React hooks


nukeop:
Something's not right with your config, travis fails to build it.

karevn:
Yep, I'll check it, thanks!

karevn:
@nukeop seems to run fine by now

nukeop:
Faster builds are always nice.

What's wrong with babel polyfill? Why do we need core-js now? I'd like to avoid that package if possible because of advertisements during installation.

karevn:
It's because Babel polyfill was [deprecated](https://babeljs.io/docs/en/babel-polyfill)

nukeop:
I'll check if I really need core-js if I'm only targeting electron. That package should be avoided like the plague.

karevn:
Lol. Can't agree it's a kind of a plague, while totally agree that unused code should not be included into the project :) I'll do some checks on my side too.

karevn:
It seems to work without core-js, but IMO some more testing is required.

nukeop:
I tested most of the functionality and I think this can be merged. It really does build noticeably faster too, thanks for that. Looks like electron doesn't need polyfills and regenerator runtime is all that's needed. Thank you for your work, it's very helpful.

Issue #387, opened 2019-07-02
Create Enzyme tests for all major views


nukeop:
We need to create some tests that will let us at least test if all the major views render correctly with some mocked data. This will let us avoid regressions that keep happening recently. This will likely be implemented in parts as there are many different views that need to be safeguarded like that.

"Good first issue" applies to any number of tests, even if it's just one small test for one component.

ItzAmir:
What about snapshot tests? you can test if component renders correctly and also be aware of changes. If you think they can help tell me to start writting them!

ItzAmir:
Are there any components which need to have a test more than others?

nukeop:
Any of the following, in random order:
- Play queue and its items
- Play controls (play button, previous/next)
- Left hand menu
- Album view
- Artist view

All of these are the most important elements that are responsible for the most commonly used functionality.

ItzAmir:
I will work on AlbumView and ArtistView

jenyeeiam:
I just made a PR for the Play controls 
https://github.com/nukeop/nuclear/pull/447

mahimshridhar:
Can i try anything here ?

nukeop:
Sure, pick whatever you like.

PR #416, opened 2019-07-25
Tests for VerticalPanel, WindowControls, Navbar and WindowButtons

jayjamesjay:
#387 

Small portion of tests for components:
- VerticalPanel
- WindowControls
- Navbar
- WindowButtons


nukeop:
Thanks, this is a great contribution, we need as many rendering tests as possible - recently we've had some broken functionality on master that could have been avoided precisely by having more tests.

PR #419, opened 2019-07-27
Update zh.json

hedyhyry:


hedyhyry:
I added translated another whole lot of things, go to diffs and you’ll find out.

nukeop:
Thanks, very helpful. Bit by bit we're getting to a complete translation.

Issue #412, opened 2019-07-20
Graphic designs contributions are welcome


nukeop:
This isn't really an issue, but an attempt to encourage potential non-technical contributors.

We welcome all kinds of contributions, not only code. In particular, I'd like to make clear that graphic designs of all kinds are also very much appreciated. Example of what could be contributed in this regard:

- A graphic illustrating a proposed redesign of some functionality to improve its UX
- A new logo proposal
- A proposal of a completely new skin for the entire program
- Graphic assets for the official website, such as backgrounds, screenshots, or images
- Advice regarding the UX design of some functionality
- CSS changes to improve styling
- Merchandise designs, such as stickers or t-shirts
- Ideas for new functionalities, possibly with mockups illustrating a proposed GUI

Those are just some of the things you could contribute if you're not coding, but would still like to help. You can even do it without using git at all, just open an issue and we'll help you find the best way to add whatever you create.

awalus:
Hello, I have some suggestion for a logo. (It's not ready yet, it's just an idea.)

![Nuclear Logo Proposition](https://user-images.githubusercontent.com/55683811/68019035-e776da80-fc9a-11e9-8741-49578be77aa3.png)

CodeWithBlaze:
hello i want to contribute to this project .i have never cotributed before please guide me how can i get started in this project

nukeop:
What would you like to contribute?

CodeWithBlaze:
i am a coder . i know javascript and java almost 60% and i like your project and just want to contribute . so anything for me??

nukeop:
Almost 60% of what?

CodeWithBlaze:
java and javascript

nukeop:
You can look at the open issues and pick anything in the bug or feature request category.

CodeWithBlaze:
Okkk thnx

PR #428, opened 2019-08-03
Visual upgrades: Player hover states, plugins view #412

nikolajjepsen:
A few visual upgrades specifically in the footer player and the plugins view. 
Individually described in commits.


nukeop:
Thanks, this makes styles more consistent.
There are some durations defined here: https://github.com/nukeop/nuclear/blob/master/app/vars.scss

Could you use those in your changes? Or do you think it'll look better if we leave your durations?

nikolajjepsen:
Looks just fine with the defined lowest duration.

nukeop:
Great, I'll merge as soon as Travis finishes building.

PR #430, opened 2019-08-05
Unit tests for components

ItzAmir:
#387
A couple of tests for the following components:

- Card
- AlbumList
- HelpModal
- NavButtons
- FavoriteTracksView
- LyricsView

I had to stub some dependencies using proxyquire so if you are not comfortable with that, tell me a better way to change it.


nukeop:
Thank you for the contribution, this is really helpful and I see that you put a lot of effort into this, which is great. We had some problems in the past involving some components not rendering correctly after unrelated changes and your tests will help us catch that in the future.

Would you be able to use chai-spies instead of sinon for the spies? We already have chai as a dependency and maybe it would be easier? Unless you think sinon is better.

ItzAmir:
> Would you be able to use chai-spies instead of sinon for the spies? We already have chai as a dependency and maybe it would be easier? Unless you think sinon is better.
 Done!
I used sinon because it was more popular but personally prefer chai-spies because it is simpler and enough to spy functions.

nukeop:
You broke one test accidentally, but I'm going to merge your changes anyway then fix it myself. Great contribution, thank you.

ItzAmir:
Couldn't understand why that test failed it works on my machine.
Also, I didn't change anything about that test.

nukeop:
No problem, it's already fixed. 

The problem was that lodash inserts its `_` into the global scope when imported, but when you only import a single component in the test, lodash isn't loaded, so I had to add the import to the component itself.

PR #433, opened 2019-08-06
Got rid of high CPU usage when idling

karevn:
Nuclear used to consume a considerable CPU time due to the animated "loader" being rendered in the background. I changed the HTML to make it completely disappear upon first render.

### DIFF
--- a/app/app.global.scss
+++ b/app/app.global.scss
@@ -17,7 +17,7 @@ body {
 }
 
 body.dimmed {
-  #react_root {
+  #react-root {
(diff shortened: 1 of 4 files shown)


karevn:
@nukeop can you please check what's happened to the CI?

nukeop:
A moving, completely obscured gradient hammers CPU? WTF is wrong with Chromium?

That CI fail is just a network error, nothing to worry about. Thanks for the contribution.

PR #438, opened 2019-08-07
Scrollable left sidebar + non-shrinking gategory headers

karevn:


nukeop:
Makes sense, nice change 👍

PR #439, opened 2019-08-09
A fix for seekbar height / font size in Linux

karevn:
After one of the latest fixes, a track duration started to drop off the seekbar under Manjaro. I made its height depended on the real font size in this commit.


nukeop:
Maybe 1em will be more foolproof? This way we won't need to keep both values in sync.

karevn:
Yes, thanks for mentioning!

PR #441, opened 2019-08-11
Tidier tags representation

karevn:
image
image


nukeop:
Thanks, looking good. The alignment is better too.

PR #442, opened 2019-08-11
Disallow playing queue items which are being loading + no spinner for…

karevn:
… them.


app/components/PlayQueue/QueueItem/styles.scss
```diff
@@ -1,8 +1,17 @@
 @import '../../../vars';
 
+@keyframes appear {
```
nukeop: Thanks for this change, I implemented this originally when I had no idea how to use css animations.

nukeop:
Very useful, thank you. You're really contributing a lot!

PR #443, opened 2019-08-11
Don't show empty search result categories

karevn:
image

### DIFF
--- a/app/components/SearchResults/AllResults/index.js
+++ b/app/components/SearchResults/AllResults/index.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import _ from 'lodash';
 import artPlaceholder from '../../../../resources/media/art_placeholder.png';
 
 import Card from '../../Card';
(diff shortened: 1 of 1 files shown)


app/components/SearchResults/AllResults/index.js
```diff
@@ -92,20 +93,20 @@ class AllResults extends React.Component {
...
       return <div>{this.props.t('empty')}</div>;
     }
 
     return (
       <div className={styles.all_results_container}>
-        {this.renderArtistsSection()}
-        {this.renderAlbumsSection()}
-        {this.renderTracksSection()}
-        {this.renderPlaylistSection()}
+        {artistsLength > 0 && this.renderArtistsSection()}
```
nukeop: Maybe all these could be replaced with `_.isEmpty`? This way there would be no need to declare the consts above

karevn: I think it won't be better, as `info` can be `null`, and checking `_.isEmpty(null)` is quite ambiguous - you have to remember the exact specification of Lodash call to tell if it's true or false. Moreover, `_.isEmpty` does some typechecking, which adds some performance overhead. So, I think the current solution is a little bit wordy, but readable, fast, and bulletproof.

nukeop: Ok, good point. However, you could simplify the syntax a bit like this:
```javascript
    const tracksLength = _.get(this.props.trackSearchResults, 'info.length', 0);
    const artistsLength = _.get(this.props.artistSearchResults, 'length', 0);
```

and similarly for the rest

karevn: I tend to use arrays in Lodash get calls for three reasons:
1. It's faster as it does not involve any string parsing, and literal arrays can be extracted to constants to further improve the performance in critical paths.
2. It's magnitudes more reliable when you need dynamic expressions, especially when strings and numeric indices are mixed.
3. And taking in account  1 and 2, I use arrays in other cases for sake of consistency :)

nukeop: Nice, I didn't know that. I guess it can stay like that if there's a good reason for it.

PR #444, opened 2019-08-11
Repeat search on Enter keypress in the search box

karevn:


app/components/SearchBox/index.js
```diff
@@ -8,6 +8,17 @@ import { useTranslation } from 'react-i18next';
 
 const SearchBox = ({ handleSearch, loading }) => {
   const { t } = useTranslation('search');
+  const [latestSearch, setSearch] = useState('');
+  const handleKeyDown = useCallback((e) => {
+    if (e.which === 13) {
```
nukeop: Could this 13 be bound to a const somewhere, preferably in this file?
I'm guessing it corresponds to the enter key, so you could name it KEY_RETURN or something.

karevn: 👍

PR #446, opened 2019-08-11
Add redux-logger in a development environment

karevn:


app/store/configureStore.js
```diff
@@ -9,11 +10,17 @@ export default function configureStore(initialState) {
     ? compose
     : window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__; 
 
+  const logger = createLogger({});
```
nukeop: We already have redux dev tools, which imo provide superior functionality to this logger.

This, however, could be useful if it logged certain important actions to stdout using `electron-timber`.

karevn: It seems that I've overlooked this miraculous tool. Let me give it a shot.

karevn: Oh. My. Closing this PR.

nukeop: Heheh, glad you found out about something useful!

PR #447, opened 2019-08-11
Play pause test

jenyeeiam:
My first ever PR! Suggestions welcome :)

Enzyme tests for the PlayerControl components

### DIFF
--- a/test/enzyme.playercontrols.test.js
+++ b/test/enzyme.playercontrols.test.js
@@ -1,0 +1,43 @@
+import 'jsdom-global/register';
+import React from 'react';
+import { expect } from 'chai';
+import { shallow, mount, render, configure } from 'enzyme';
+import { describe, it, before } from 'mocha';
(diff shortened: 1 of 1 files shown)


nukeop:
Thanks for the contribution, looks really good.
I think you forgot to add jsdom-global to dev dependencies. Do you think it's necessary for the tests to run?

Take a look here: https://github.com/nukeop/nuclear/blob/c1c671ca9ae127a6f1c774a96ab8a2a62ee67a88/test/setup.js
JSDOM is already included in test setup, maybe this will already setup the environment the way you want to?

jenyeeiam:
Oh ya ok. I was importing jsdom because I was getting errors when I was trying to do a full DOM rendering with mount(). What's the best way to do it with this setup file?

jenyeeiam:
Sorry! This was a result of me needing this in order to run my test individually. Making my changes now

nukeop:
Now it's passing, thanks.

PR #449, opened 2019-08-12
Got rid of Happypack

karevn:
The build process without Happypack is consistently faster by 10-15%, so this tool does no good for this project.

### DIFF
--- a/package.json
+++ b/package.json
@@ -126,7 +126,6 @@
     "eslint-plugin-react-hooks": "^1.6.1",
     "extract-text-webpack-plugin": "^3.0.2",
     "file-loader": "^1.1.6",
-    "happypack": "^5.0.0",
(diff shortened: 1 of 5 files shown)


nukeop:
I think the project grew beyond the point where it could improve build performance. I remember it speeding up the build back when I added it. Let's compare the build time when Travis finishes. The previous build was 16841ms.

karevn:
For me, it's about 19-21 seconds with Happypack (warmed up) vs 16-18 without it on macOS. On Linux, it's like 13 secs vs 10-11. All - using 2015 Macbook Air. I can test macOS performance on a more powerful machine tomorrow.

nukeop:
The build is down to 11294ms on Linux, fantastic improvement. Windows is still ridiculously slow at 43793ms.

PR #455, opened 2019-08-18
Replaced hardcoded strings for proper i18n resources

emlautarom1:
I removed some of the hardcoded texts, specially in the "track popup", for i18 resources and also I added a Spanish version for those strings.

Also, concerning code style, some components use `useTranslation` and others use `withTranslation` decorator. I'm not a react dev, but couldn't we standarize a way to handle internationalization?


nukeop:
The json is malformed, could you fix that?

emlautarom1:
It should be ok now.

PR #459, opened 2019-08-21
Jest + React 16.9 + snapshot tests for components

karevn:
Using snapshot tests makes component tests much easier to write / test - no need to invent matchers to check if elements are in their places. Also, fixed some places here and there to make tests more reliable.


nukeop:
I decided that we're not going to go ahead with jest, perhaps for a new package in the future.

PR #462, opened 2019-08-22
Feature/redux query

karevn:


karevn:
Please note: this one is not supposed to be merged as it is, it's just a demo. To be viable in a real world, it should be further refactored not to keep all the code in the same file.

karevn:
Over a month has passed. Any news?

nukeop:
I opted not to use this, sorry.

PR #473, opened 2019-09-16
#193 Add best new album to favorites

sgraczyk:
Part of #193 issue.

Hi,

So I've decided to continue working around adding items to favorites from `Best` page perspective. As I've noticed it is now possible to add albums to `loved ones` I went for it. It proved far more complicated as I was unable to simply `convert` the album from `pitchfork` to shape of favorites. I was missing Discogs `master_id`. Initially I decided to drop my work as I was thinking it would be too complex, but finally, I've decided to at least show it so you could think about it. There are a couple of things worth to mention before any review:

* Checking if an album is a among favorite ones is still done via `artist` & `title` comparison.
* Additional gotcha here is that `Discogs` may provide an artist with a number in paratheses, meaning there are more artists with such name. (eg. `Florist (2)`).
* Before adding the album to favorite, we need to find it in a `Discogs` database.
* We may not find any album matching search criteria.

Considerations for future. 
* Fetch `Discogs` info right after fetching `Pitchfork` data.


app/actions/favorites.js
```diff
@@ -56,7 +58,12 @@ export function addFavoriteAlbum(album) {
 
 export function removeFavoriteAlbum(album) {
   const favorites = store.get('favorites');
-  _.remove(favorites.albums, { id: album.id });
+  if (album.uuid) {
```
sgraczyk: I've used the `uuid` but I'm not sure about it. I've added handling of both `id` and `uuid` for backward compatibility.

app/components/Dashboard/BestNewMusicTab/BestNewMusicMenu/BestNewMusicCard/index.js
```diff
@@ -24,24 +30,94 @@ function toFavoriteTrack({ artist, title, thumbnail }) {
   };
 }
 
+function getReleaseInfo({ artist, title }) {
```
sgraczyk: I didn't want it to be a part of `redux` action as I didn't see any data to be stored.

app/components/Dashboard/BestNewMusicTab/BestNewMusicMenu/BestNewMusicCard/index.js
```diff
@@ -24,24 +30,94 @@ function toFavoriteTrack({ artist, title, thumbnail }) {
...
+    })
+    .then(releaseResponse => releaseResponse.json())
+    .catch(error => {
+      logger.error(error);
+
+      throw error;
+    });
+}
+
+function showSuccessInfo(action, item, settings, type) {
```
sgraczyk: Just for `DRY` purpose

app/components/Dashboard/BestNewMusicTab/BestNewMusicMenu/BestNewMusicCard/index.js
```diff
@@ -24,24 +30,94 @@ function toFavoriteTrack({ artist, title, thumbnail }) {
...
+      }
+
+      if (type === ItemType.ALBUM) {
+        getReleaseInfo(item)
+          .then(releaseItem => {
+            actions.addFavoriteAlbum(releaseItem);
+            showSuccessInfo(actions.info, item, settings, type);
+          })
+          .catch(() => {
+            actions.info(
```
sgraczyk: I believe handling failed fetch would be nice.

app/rest/Discogs.js
```diff
@@ -17,23 +17,27 @@ function searchQuery (terms, count = 15) {
...
 }
 
-function search (terms, type, count = 15) {
+function search(terms, type, count = 15) {
   return fetch(searchQuery(terms, count) + '&type=' + type);
 }
 
-function releaseInfo (releaseId, releaseType) {
+function findAlbum(artist, title) {
+  return fetch(addToken(`${apiUrl}database/search?type=master&per_page=1&release_title=${encodeURIComponent(title)}&artist=${encodeURIComponent(artist)}`));
```
sgraczyk: More direct way to find an album

app/selectors/favorites.js
```diff
@@ -15,3 +15,26 @@ export function getFavoriteTrack(state, artist, title) {
...
+  const normalizedTitle = _.deburr(title.toLowerCase());
+
+  return _.find(state.favorites.albums, album => {
+    const albumArtist = album.artists[0];
+
+    if (!albumArtist) {
+      return false;
+    }
+
+    const artistQuantifierRegex = /\s\([0-9]*\)$/;
```
sgraczyk: Regex that removes the last occurrence of `({number})` in artist name like `Florist (2)`. It should ignore any parentheses inside. May fail in case of any artist really ending with `({number)`

test/utils.test.js
```diff
@@ -18,7 +18,7 @@ describe('utils', () => {
     });
   });
 
-  describe.only('createLastFMLink', () => {
```
sgraczyk: oops :)

nukeop:
Wow, thanks a lot, this is great. I will review and test this soon.

app/components/Dashboard/BestNewMusicTab/BestNewMusicMenu/BestNewMusicCard/index.js
```diff
@@ -24,24 +30,94 @@ function toFavoriteTrack({ artist, title, thumbnail }) {
   };
 }
 
+function getReleaseInfo({ artist, title }) {
+  let masterId;
+  return findAlbum(artist, title)
```
nukeop: We don't need this here. Best new music card should only be concerned with presentation, and the minimal amount of logic. To add a favorite album, you only need what's already available from Pitchfork, i.e. artist, name, and thumbnail. Everything else will be looked up when we click the album.

If memory fails me though, and it doesn't work that way, then it'll be worthwhile to move this piece of logic to the favorite albums container, so that the discogs lookup is performed upon clicking the album. We really only need the id, so that we know where to go, but soon I might be changing that so we end up only looking up stuff after we go to the album page.

app/components/Dashboard/BestNewMusicTab/BestNewMusicMenu/BestNewMusicCard/index.js
```diff
@@ -24,24 +30,94 @@ function toFavoriteTrack({ artist, title, thumbnail }) {
...
-  favoriteTrack,
+  favoriteItem,
   onClick,
   settings,
-  withFavoriteButton
+  type
 }) => {
   const { artist, title, thumbnail } = item;
 
+  const handleFavoriteIconClick = useCallback((e) => {
```
nukeop: Use `recompose` and its `withHandlers`

app/components/Dashboard/BestNewMusicTab/BestNewMusicMenu/BestNewMusicCard/index.js
nukeop: This can still be an action, even if it's not storing anything in the redux store. Same for the piece of logic earlier. Anything that performs http requests should be kept out of presentational components.

app/constants/itemType.js
```diff
@@ -0,0 +1,5 @@
+export default {
+  ALBUM: 'album',
```
nukeop: also worth to consider artists here

app/selectors/favorites.js
nukeop: Yep, I didn't realize that. It looks like this is an internal Discogs suffix added to distinguish between artists with the same name. Doesn't help us since we look at ids anyway.

nukeop:
Ok, I finished the code review. There are some things that I'd like to be changed, if you feel like it.

sgraczyk:
Thanks for the review, I'll try to adapt the solution to your ideas during this week. In case of any questions from my side, I'll let you now.

Issue #467, opened 2019-09-05
Configure downloads directory


Mongogamer:
- **Nuclear version**: 0.5.0 commit 5afd2c678e8b05771fbd3da055cd2f728d9251f2
- **OS**: Windows 10 Pro x64 1903
- **NodeJS version**: 10.16.0
---
According to this https://github.com/nukeop/nuclear/issues/150#issuecomment-456225563 the ability to configure the download directory was a planned part of the downloads feature. It seems this feature never made it into the final version. After looking through some of the download code it seems like implementing this would not be too hard.

PR #477, opened 2019-09-21
Finalize download directory setting

Mongogamer:
Part of #467 issue.
This change is required for Nuclear to use the newly created setting. Without this Nuclear always downloads to the default download directory even if the setting is changed.


nukeop:
Yep, thanks. I wanted to add this today, but you were faster. I will also add a box to change this setting to the downloads screen.

PR #482, opened 2019-09-26
Magordon97

magordon97:
On the Downloads page, split Artist-Song category into two separate categories: Artist and Song. Allow sort by artist name. Allow sort by track name.


app/components/Downloads/DownloadsList/index.js
```diff
@@ -23,6 +24,23 @@ const DownloadsList = ({
...
+                });
+                setArtistSort(false);
+              } else {
+                items.sort((a, b) => {
+                  return a.track.artist.name.toLowerCase() < b.track.artist.name.toLowerCase();
+                });
+                setArtistSort(true);
+              }
+            }
+            }>{t('Artist')} {
```
nukeop: This, along with `t('Song')` won't do what you want it to do. This is a translation function and the way it works is by getting the strings from appropriate locale by their key. E.g. for english, it looks here: https://github.com/nukeop/nuclear/blob/master/app/locales/en.json

For downloads, it uses the string located under the "downloads" key. To get the artist/title pair for the song in the download queue, you can just access its object directly.

nukeop:
Nice contribution! I found one small mistake, could you please take a second look? I will be happy to merge after it's fixed.

PR #496, opened 2019-10-03
#489 Disabling more queue options if queue is empty

### LINKED ISSUE #489: A file is being downloaded even though download queue is empty
Even when the download queue is empty, it allows me to click 'download', and an unknown file begins downloading. This file becomes `undefined - undefined.m4a`. 

Screen Shot 2019-09-30 at 11 00 55 PM
Screen Shot 2019-09-30 at 10 57 33 PM
Screen Shot 2019-09-30 at 10 57 41 PM

Could I tackle this issue? Possibly remove the option to download when the queue is empty? Open to other suggestions. Would love to contribute.

badsketch:
Resolves #489.

Previously, a file would be downloaded even if the queue was empty. This has already been resolved with https://github.com/nukeop/nuclear/commit/f43012b3aa6e26e549d23d1173f29c329a6b8db4, so this is a UX change to disable the entire options menu if the queue is empty.

First PR. Looking for any feedback. 👍

### DIFF
--- a/packages/app/app/components/PlayQueue/QueueMenu/index.js
+++ b/packages/app/app/components/PlayQueue/QueueMenu/index.js
@@ -59,7 +59,7 @@ class QueueMenu extends React.Component {
           {
             !compact &&
               <QueueMenuMore
-                disabled={!items.length}
+                disabled={_.isEmpty(items)}
                 clearQueue={clearQueue}
                 updatePlaylist={updatePlaylist}
                 addFavoriteTrack={addFavoriteTrack}
--- a/packages/app/app/components/PlayQueue/QueueMenu/index.js
+++ b/packages/app/app/components/PlayQueue/QueueMenu/index.js
@@ -59,6 +59,7 @@ class QueueMenu extends React.Component {
           {
             !compact &&
               <QueueMenuMore
+                disabled={!items.length}
                 clearQueue={clearQueue}
                 updatePlaylist={updatePlaylist}
                 addFavoriteTrack={addFavoriteTrack}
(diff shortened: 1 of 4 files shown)


packages/app/app/components/PlayQueue/QueueMenu/index.js
```diff
@@ -59,6 +59,7 @@ class QueueMenu extends React.Component {
           {
             !compact &&
               <QueueMenuMore
+                disabled={!items.length}
```
nukeop: Use `_.isEmpty(items)`, which will be more obvious

nukeop:
Thanks for this, looks very good. I only have 1 minor comment and we can merge.

badsketch:
Ah okay, that makes sense. I'll keep that in mind for the future. 👍 

Also, I wasn't sure what would be considered best practice. 

- check if `currentItem == null` inside `<QueueMenuMore />` 
- create `disabled` prop and check if `items == empty` inside `<QueueMenu />` which is passed to `<QueueMenuMore />` **(what i ended up doing)**
- create `disabled` prop and check if `items == empty` inside `<PlayQueue />`  which is passed to `<QueueMenu />` which is passed to `<QueueMenuMore />` 

I know the end result is trivial, but do you know what is considered the proper react pattern for keeping things extensible and readable? Hope the question makes sense.

nukeop:
I think what you chose is the best approach. We want the queue context menu to be able to be disabled for  any reason, so it's smart to have a prop for that. Queue menu already needs a list of all items, so we can also quickly check if it's empty without adding more code to other components. So in other words, what you did results in the best reusability and the smallest amount of extra code.

Issue #491, opened 2019-10-01
Automatically choose translation based on locale


nukeop:
We could automatically set language based on current locale (at least on Linux). This would be relatively easy to implement, so contributions are very welcome. If anyone decides to take it, I'll be able to provide some guidance so don't hesitate even if you're a newbie.

eprotagoras:
@nukeop I think I would be interested in picking this up. This would be my first time contributing!

nukeop:
Cool, thanks for the contribution. How advanced are you? Do you feel confident to tackle this on your own?

eprotagoras:
I've worked with React before, although this will be my first time using Electron. After some investigation, it looks like we do use `i18next`'s language detector plugin to automatically set language using `app.getLocale()` in `i18n.js`, although we prioritize a selected language option persisting from a previous session.
```
const languageDetector = {
  init: Function.prototype,
  type: 'languageDetector',
  detect: () => getOption('language') || remote.app.getLocale(),
  cacheUserLanguage: Function.prototype
};
```
It looks like this is expected behavior, unless we wanted to prioritize the current locale. What do you think @nukeop ?

nukeop:
Nice, although there's still one thing missing. It would be good for the automatically detected language to be displayed in settings, as currently the input box is empty if you don't select anything manually. Do you think you could do that?

eprotagoras:
Sounds good, although after playing around with a couple of different locales, it seems like the only locale that isn't automatically displayed in the input box if you don't select anything is `en-US`, since `i18n.language` is already being passed as the `value` prop into the Dropdown. However, the language options don't have an entry for `en-US`, only `en` in general. All the other locales I tested (ie, `de`, `fr`, `it`, etc) displays initially.

I found that adding an `en-US` option worked, since now the callback attached to `i18n` sets the language option to a valid key.
```
{
    name: 'language',
    category: 'program-settings',
    type: settingType.LIST,
    prettyName: 'language',
    placeholder: 'language-placeholder',
    options: [
      { key: 'de', text: 'Deutsch', value: 'de' },
      { key: 'dk', text: 'Dansk', value: 'dk' },
      { key: 'en-US', text: 'English (US)', value: 'en-US' },
      { key: 'en', text: 'English', value: 'en' },
      { key: 'es', text: 'Español', value: 'es' },
      { key: 'fr', text: 'Français', value: 'fr' },
      { key: 'it', text: 'Italiano', value: 'it' },
[...]

PR #499, opened 2019-10-03
Issue #491: Added en-US option for settings language dropdown

eprotagoras:
Fix for Issue #491: Automatically choose translation based on locale. 

Added en-US option for settings language dropdown so that i18n `languageChanged` callback sets the language option to a valid key while using `en-US` locale.


nukeop:
Thanks! Even though the fix is only a single line, you had to do some debugging to figure that out, and I really appreciate that.

PR #505, opened 2019-10-07
added 3 icons to the about modal

jenyeeiam:


packages/app/app/components/HelpModal/index.js
```diff
@@ -48,6 +62,19 @@ const HelpModal = () => {
...
         <Modal.Description>{agplDisclaimer}</Modal.Description>
       </Modal.Content>
+      <Modal.Content>
+        <div className={styles.social_icons}>
+          <div className={styles.icon}>
+            <Icon link size={'big'} name='twitter' onClick={handleTwitterClick}/>
+          </div>
+          <div className={styles.mastadon}>
+            <Image src={mastadonImg} onClick={handleMastadonClick} />
+          </div>
```
jenyeeiam: I really had trouble doing this with the svg. So I converted it to a png and went with that. If you have a better solution then I'd be glad to do it like that

jenyeeiam:
Huh, I'm not sure why this failed?

nukeop:
It's just windows, don't worry.

We're loading svg as text, so to make it work you have to use __html, like here: https://github.com/nukeop/nuclear/blob/b9feeb853b5c5644a36442e33d21fca3075a3d01/packages/app/app/components/Dashboard/GenresTab/index.js#L46

Do you want to try with svg or should I merge?

jenyeeiam:
Yeah I did notice the `dangerouslySetInnerHTML` that and I still couldn't get the svg to show up. I don't know if it had something to do with the modal? Not sure 😞. In any case, here's the screenshot 
Screen Shot 2019-10-07 at 8 39 39 AM

If in the future you want to upgrade fontawesome, version 5 includes the mastadon icon.

nukeop:
Looks awesome! Thanks for the help.

PR #509, opened 2019-10-08
#500 Updating react-beautiful-dnd and fixing drag logic

### LINKED ISSUE #500: Unexpected dragging behavior from play queue
Dragging outside the play queue causes errors.
To Reproduce:
1) Add a song to queue
2) Drag the song outside of the queue. It should pop back to the queue
3) Drag the song again outside of the queue. It will freeze in place

Screen Shot 2019-10-04 at 12 30 56 AM

Could I investigate this issue and improve dragging functionality?

badsketch:
Resolves #500 dragging issues.

Updated react-beautiful-dnd v9.0.0 to v11.0.0. 

Dragging properly indexes the songs between the source and destination

**before**: Dragging used to keep the current song at its index.
sameIndex

**now**: Dragging the current song moves it to where it is dragged.
flicker

But the issue is now if the current song is dragged, then the background flickers between its old and new position 😕. I've tried using `React.memo` hoping it's a rerendering issue, but it doesn't work. I've also tried a constant time approach to calculating where the current song should be instead of the `findQueueItemIndex` linear search. Wondering how to pinpoint where the delay is coming from. Any suggestions?

If I can't fix it, should I consider just removing my last commit and only PRing the package update and index fix?

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,7 +124,3 @@ In case of errors with dbus/mpris, try removing optional dependencies from `pack
 ## License
 
 This program is free software: you can redistribute it and/or modify it under the terms of the GNU Affero General Public License as published by the Free Software Foundation, either version 3 of the License, or (at your option) any later version.
-
-#### Support on Beerpay
-
-[![Beerpay](https://beerpay.io/nukeop/nuclear/badge.svg?style=beer-square)](https://beerpay.io/nukeop/nuclear)  [![Beerpay](https://beerpay.io/nukeop/nuclear/make-wish.svg?style=flat-square)](https://beerpay.io/nukeop/nuclear?focus=wish)


packages/app/app/components/PlayQueue/index.js
```diff
@@ -47,7 +54,7 @@ class PlayQueue extends React.Component {
 
     return this.props.items.map((el, i) => {
       return (
-        <Draggable key={i} index={i} draggableId={i}>
+        <Draggable key={i} index={i} draggableId={`${i}`}>
```
badsketch: breaking change was that `draggableId` has to be a string now. 

I initially tried using `uuid` for the `key` and `draggableId`, but it was causing a different flickering issue. Plus, adding a playlist to the playqueue twice will cause repeats in `uuid`. Looking for any alternatives/suggestions.

nukeop:
Could you try to rebase or cherry pick so that there are only your commits in the PR? It looks like a bunch of other work got caught too and it makes it harder to evaluate your work.

Were you able to figure out why the id causes flickering? Is it because of some inefficient comparison somewhere?

badsketch:
No, I wasn't able to, but I'm gonna take another look tonight. 

As for the commit cluttering, do you know what might have been the cause? I occasionally `git fetch/merge upstream master` into my remote branch. Is that bad practice?

nukeop:
Yes, there's no need to do that, unless there are drastic changes across the whole codebase. Merge conflicts should be resolved at merge time, and it should ideally be automatic or take only small manual changes. No need to merge master back since it'll only make things more difficult to review, and will be done anyway when the PR is accepted.

PR #512, opened 2019-10-09
Display stream id only if exist

Zalaxx:
The stream Id was always displayed on the stream info, regardless the track has one or not.

| Before | </img>|                                                                                                                  
|:------:|:----------------------------------------------------------------------------------------------------:|
| **After** | </img> |

### DIFF
--- a/packages/app/app/components/QueuePopup/index.js
+++ b/packages/app/app/components/QueuePopup/index.js
@@ -102,13 +102,11 @@ class QueuePopup extends React.Component {
             <label>{t('title')}</label>
             <span>{selectedStream.title}</span>
           </div>
-          {selectedStream.id ? (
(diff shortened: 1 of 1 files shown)


packages/app/app/components/QueuePopup/index.js
```diff
@@ -102,10 +102,14 @@ class QueuePopup extends React.Component {
             <label>{t('title')}</label>
             <span>{selectedStream.title}</span>
           </div>
-          <div className={styles.stream_id}>
-            <label>{t('id')}</label>
-            <span>{selectedStream.id}</span>
-          </div>
+          {selectedStream.id ? (
```
nukeop: You can use `selectedStream.id &&` instead of the ternary.

PR #513, opened 2019-10-09
Use https instead of http

Zalaxx:
Some links were still in http, while sites support and redirect to https. There is no reason to stay in http.


nukeop:
You're correct, thanks for the change.

PR #514, opened 2019-10-09
fix min size of the steam info

Zalaxx:
As the stream_info takes the size on the background image, when the thumbnails are smaller than 124px, not all the information can be displayed properly. 124px comes from the size of a thumbnail of a Youtube music.

| Before | </img>
|---------| --------------------------------------------------------------------------------------------------------------------------------------------------------|
| **After** | </img> |


nukeop:
Great job, I remember seeing this issue some time ago but I forgot about it. You're on a roll today.

PR #515, opened 2019-10-10
fix artPlaceholder for local library

Zalaxx:
The artPlaceholder was missing for the local library.

| Before | </img> |
|------------|------------------------|
| **After** | </img> |


Zalaxx:
I just noticed that we could remove the second get since'image[0][#text]' is only for last.fm url.

Zalaxx:
Ok my bad, the second commit was wrong, all the tracks have the artPlaceholder without the second .get.

PR #518, opened 2019-10-12
Comparing the song name instead of the whole track object 

jenyeeiam:
This should help with adding duplicated tracks to favorites. The isMatch function was comparing the entire track objects which would not necessarily be deeply equal due to the play count or whatever. It should be sufficient to compare just the song name and that would be the check to whether you can add this track to your favorites or not

### DIFF
--- a/packages/app/app/components/TrackRow/index.js
+++ b/packages/app/app/components/TrackRow/index.js
@@ -53,7 +53,7 @@ class TrackRow extends React.Component {
 
   canAddToFavorites() {
     return _.findIndex(this.props.favoriteTracks, (currentTrack) => {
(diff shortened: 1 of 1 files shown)


nukeop:
Good idea, although soon I will try to solve this problem of track identity more comprehensively like I partially did for artists already:
https://github.com/nukeop/nuclear/blob/master/packages/app/app/structs/artist.js

Sometimes we need to perform comparisons that check the ids, and sometimes we're interested in tracks having matching artist/title. So a struct for tracks will need to expose several comparison functions for different purposes anyway. The structs will also be moved to a separate package.

jenyeeiam:
Sounds good. Let me know if there's anything I can help with

PR #521, opened 2019-10-13
improve source fallback

Zalaxx:
If there is no selectedStream, we just try to use Youtube as a source. This means that music only on Soundcloud cannot be played without changing the favorite music source in the plugin settings. And as "favorite music source" says in the name, the software should use the favorite source, but automatically use the other sources if the favorite one is not available. I've used a bit the software with the modification, and it works perfectly.


nukeop:
The idea is sound, but I don't like the code. It would be better to just use `_.filter` and `!_.isNil(stream)` then get the first item from that list.

PR #523, opened 2019-10-13
Fix drag behavior when songs with same ID are in the play queue

badsketch:
-fixed dragging causing issues when the id's are the same
dup

-improved repositioning logic so finding where the current song is in constant time instead of linear search

### DIFF
--- a/packages/app/app/components/PlayQueue/index.js
+++ b/packages/app/app/components/PlayQueue/index.js
@@ -54,7 +54,7 @@ class PlayQueue extends React.Component {
 
     return this.props.items.map((el, i) => {
       return (
-        <Draggable key={el.uuid} index={i} draggableId={el.uuid}>
+        <Draggable key={`${el.uuid}+${i}`} index={i} draggableId={`${el.uuid}+${i}`}>
           {(provided) => (
             <div
               ref={provided.innerRef}
--- a/packages/app/app/reducers/queue.js
+++ b/packages/app/app/reducers/queue.js
@@ -81,8 +81,25 @@ function reduceRepositionSong(state, action) {
   const [removed] = newQueue.splice(action.payload.itemFrom, 1);
   newQueue.splice(action.payload.itemTo, 0, removed)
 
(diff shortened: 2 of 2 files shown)


packages/app/app/components/PlayQueue/index.js
```diff
@@ -54,7 +54,7 @@ class PlayQueue extends React.Component {
 
     return this.props.items.map((el, i) => {
       return (
-        <Draggable key={el.uuid} index={i} draggableId={el.uuid}>
+        <Draggable key={`${el.uuid}+${i}`} index={i} draggableId={`${el.uuid}+${i}`}>
```
badsketch: this isn't any better than just using the index as the key, which i know is bad, but I want to leave it in to remind myself to find better ways to create unique keys. will open up a new issue soon.

nukeop:
I don't have a better idea. Maybe we could generate new, separate uuids for queue items. Using ids as keys isn't that bad especially if most of the time we're only going to have a few dozen elements. The only downside is that React has to render the entire list from scratch on every redraw, instead of just moving some elements. I doubt it's going to have a big performance impact. If it does we can worry about it then.

badsketch:
The second time the playlist is loaded into the playqueue, the set of songs stay in the loading state (In the gif, I demo the dragging issue before it becomes noticeable). I think this is because they share the `uuid`, so the `addStreamsToQueueItem` adds it to their first appearance only. That's another issue related to duplicate playlists being played.

I haven't looked too deep at the logic, but is there a reason we have to save the `uuid` with the playlist songs? What if we generate `uuid` ONLY when songs enter the queue. So if I click play once, that generates uuids on the fly, if i click play on the same playlist, each song gets a whole new uuid, instead of now, where they are bound to the song when a playlist is created. When playlists are created, we strip away the uuid before saving them.

If the `uuid`'s have to remain, then you're right, how about a separate `queueUUID` property?

nukeop:
The problem of track identity remains a huge pain. I will remember to add a separate uuid for queues when I work on the track struct...

PR #528, opened 2019-10-14
Translated files

s0meon3:
Translated README and Electron files to portuguese.
Added link to translated README in the original README file.


nukeop:
Great job, you even translated my rant.

PR #531, opened 2019-10-16
Improvements to the local library feature

Venryx:
Changes:
* Greatly improved the speed of library scanning by having the scanning of tracks run in parallel. (currently set to 10 concurrent track-scans, as this is small enough it doesn't cause jitter when receiving child-process data)
* Added progress-bar for the library-scan operation. (progress bar to right of refresh button)
* Fixed that an error when fingerprinting a track (within `fpcalc`, when calling `fetchAcousticId`), would cause the entire library scan operation to fail.
* Fixed that exceeding the acoustic ID quota would cause a confusing "destructuring" error, instead of just logging the fetch-response error message.

Here's a screen-capture of the progress-bar being updated:

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,8 +30,7 @@
     "build:all": "cd packages/app && npm run build:dist && npm run build:electron && npm run build:all"
   },
   "dependencies": {
-    "eslint": "^6.5.1",
-    "tiny-async-pool": "^1.0.4"
+    "eslint": "^6.5.1"
   },
   "devDependencies": {
     "lerna": "^3.16.4"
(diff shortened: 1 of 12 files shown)


package.json
```diff
@@ -30,10 +30,14 @@
     "build:all": "cd packages/app && npm run build:dist && npm run build:electron && npm run build:all"
   },
   "dependencies": {
-    "eslint": "^6.5.1"
+    "eslint": "^6.5.1",
+    "tiny-async-pool": "^1.0.4"
```
nukeop: Please add dependencies to specific packages. The root package.json should only have dependencies that are needed to run lerna.

nukeop:
Nice job, the only thing I'd like changed is to move the dependency to its appropriate place and then we can merge this. Thanks for contributing.

Venryx:
Okay, moved the dependency into `/packages/app/package.json`.

PR #535, opened 2019-10-17
Fix that a filter (in Library view) with no matches breaks the UI

Venryx:
Changes:
* Fixed that entering a filter text that doesn't match any tracks would hide the filter textbox, and show a "library empty" message (instead of a "no search results" message)
* Added some missing entries to LibraryView.PropTypes.

### DIFF
--- a/packages/app/app/containers/LibraryViewContainer/index.js
+++ b/packages/app/app/containers/LibraryViewContainer/index.js
@@ -9,7 +9,8 @@ import * as SettingsActions from '../../actions/settings';
 function mapStateToProps(state) {
   const lowercaseFilter = _.lowerCase(state.local.filter);
   const checkFilter = string => _.includes(_.lowerCase(string), lowercaseFilter);
-  const filteredTracks = _.filter(_.values(state.local.tracks), track => {
+  const unfilteredTracks = _.values(state.local.tracks);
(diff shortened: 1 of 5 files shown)


packages/app/app/containers/LibraryViewContainer/index.js
```diff
@@ -31,10 +32,12 @@ function mapStateToProps(state) {
   }
 
   return {
+    tracks: state.local.direction === 'ascending' ? tracks : tracks.reverse(),
+    filterApplied: tracks.length < unfilteredTracks.length,
```
nukeop: why not just use `!_.isEmpty(state.local.filter`?

nukeop:
Nice way of handling it, great contribution overall.

packages/app/app/containers/LibraryViewContainer/index.js
Venryx: That works too. Depends on if you want an empty library with someone typing a filter into it to receive the same message "Library is empty" or rather "No search results" (despite there being no library contents to begin with). I figured the first behavior is a bit more useful (emphasizing the empty library), but either works (since both messages are accurate).

Want me to change it?

nukeop: No, after consideration your way of handling it is better.

PR #537, opened 2019-10-17
Replace 'link above' text with mps-youtube URL

tanmayrajani:
Since this note moved to its own file, I think this 'link above' was missed out on. So fixed that.


nukeop:
You're right, thanks for noticing.

PR #539, opened 2019-10-19
Change playlist actions to mimic album view 

### LINKED ISSUE #530: Playqueue resets to first song when adding playlist or album even with queue items
When a playlist or album is added to the playqueue, it immediately begins playing at the first song. However, if there are already songs in my playqueue and I am in the middle of it, then it results in resetting at the beginning so I lose my progress. Maybe we could add a check to only reset the `currentSong` to 0 if the queue is empty? If so, could I make the change?

I also want to ask, if I were to implement this, I see that one way is to add a conditional to check if queueItems is empty and then `selectSong(0)` in the following snippet
```
  addPlaylistToQueue (
    streamProviders,
    playlist,
    addTracks,
    selectSong,
    startPlayback
  ) {
    addTracks(streamProviders, playlist.tracks);
    selectSong(0);
    startPlayback();
  }
```

However, the conditional would have to involve access to redux `state.queueItems`, and I would have to add that to `mapStateToProps` in the `<PlayListViewContainer/>`. Would that cause overhead? Would it be better to handle this logic inside an action to isolate it from the component? Thanks!

badsketch:
resolves #530 
-added option for a playlist to be added to the end of a queue under the '...' options menu

the translations are now missing an "add playlist to end of queue" line for playlists. 
"add album to end of queue" and "playlist" translations do exist though.
Should I possibly butcher some languages and assume I can just replace "album" with "playlist" in each corresponding language translation as a temporary solution?

### DIFF
--- a/packages/app/app/components/AlbumView/index.js
+++ b/packages/app/app/components/AlbumView/index.js
@@ -6,6 +6,7 @@ import { withTranslation } from 'react-i18next';
 import _ from 'lodash';
 
 import ContextPopup from '../ContextPopup';
+import PopupButton from '../ContextPopup/PopupButton';
 import TrackRow from '../TrackRow';
 import * as Utils from '../../utils';
 import {safeAddUuid} from '../../actions/helpers';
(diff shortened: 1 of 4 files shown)


nukeop:
Nope, the approach to translations is that when developing features we only provide English language strings, and the missing strings can be provided by contributors in separate pull requests. Incomplete translations will be filled in with English strings.

PR #544, opened 2019-10-29
translated in albanian language

sidorelauku:


nukeop:
Thanks, great job translating this. Could you edit the other files to enable your translation in the program? It's described here: https://github.com/nukeop/nuclear/wiki/Translating

Or would you rather I do that?

Issue #331, opened 2019-05-11
Local file problems with incomplete mp3 tags


Mongogamer:
- **Nuclear version**: 0.4.5 commit 4fb263009ca64ed7a731f40e28772a2b799ecb19
- **OS**: Windows 10 Pro x64 1809
- **NodeJS version**: 10.15.1
---
**Description of the issue**: When using the new local file feature and scanning folder for files whenever Nuclear encounters mp3 files that don't have a valid title tag set it will label them as a single undefined title. If such an undefined title is  in the library the loading of local streams becomes very inconsistent and they will often hang at "Stream still loading."

**Expected behavior**: Either ignores untagged files or uses the filename as an alternative to the title tag.

---

**Steps to reproduce**:

1. Add any folder with untagged mp3 files to the library.
2. Scan the folder for files.
3. Attempt to play undefined song.

---

**Possible solution**: Add a check if the title tag exists and if it does not, use the filename as the title.

---

[Undefined song without mp3 tags](https://i.imgur.com/IoXEVf2.png)
[No undefined song with mp3 tags](https://i.imgur.com/N6diwt7.png)
[Devtools output when trying to play undefined song](https://i.imgur.com/jiRR4QL.png)

nukeop:
Thanks for reporting this. One way of solving this could be using the filename (or the entire path) to distinguish untagged files. We could also think about using Musicbrainz to automatically tag music for convenience.

@charjac for reference

Mongogamer:
2daa65d1fe32c059f1dd14d8e6ef3e01907c0cd4 seems to have broken playback of local files for me. It seems no valid streams are passed to Nuclear regardless of files being properly tagged or not.
This line seems to be the cause as it passes `undefined` as its response.
https://github.com/nukeop/nuclear/blob/2daa65d1fe32c059f1dd14d8e6ef3e01907c0cd4/server/http/api/localfile.js#L87

[Devtools output](https://i.imgur.com/B9ppDb7.png)
[Network capture](https://i.imgur.com/o2US3xv.png)

Files without any tags do show their name properly in the library now however.

charjac:
I will investigate this later today, thx for reporting.

charjac:
ok i see where the problem is i will push a fix today

charjac:
#395 should fix this (the good old NullPointerException)

Mongogamer:
[...]

PR #546, opened 2019-10-30
Adding virtualization to the local-library track-rows

### LINKED ISSUE #545: When developing, window refuses to load if error occurs during startup
The code for launching the electron window, in development, is here: https://github.com/nukeop/nuclear/blob/b7ff2bb95e9bd33005ac49db38ae161b305458ce/packages/app/server/main.dev.js#L35-L140

Note this section:
```
win.once('ready-to-show', () => {
  win.show();
});

win.webContents.openDevTools();
```

That looks innocent, and usually it works fine.

However, guess what happens when an error occurs during program startup?

First, the dev-tools are opened. Then, the error gets hit. This causes the just-opened dev-tools to pause execution, which causes the `ready-to-show` event to never trigger, meaning the window is never shown!

The fix is simple -- open the dev-tools *after* the window is shown:
```
win.once('ready-to-show', () => {
    win.show();
    // this must run after win.show(), otherwise startup errors cause
    // dev-tools to pause execution, causing "ready-to-show" to never trigger
    win.webContents.openDevTools();
  });
```

I can confirm this fixes the issue -- at least for the startup error I was getting (the normal errors on library startup, in their browser-functionality-testing try-catch sections; the problem is that I had "Pause on caught exceptions" enabled from the previous run).

I'll submit the fix soon (as part of my virtual-list pull-request).

Venryx:
* Fixed that, in development, program would fail to launch if an error occurred during startup. (resolves #545)
* Added virtualization of the track-rows in LibraryView. This dramatically improves performance for large local libraries, and solves the extreme jittering/freezing that occurs when attempting to play a song while having the library page open.

Here's a video of the virtualized, much-smoother library UI in action (previously it would jitter/freeze so much during song playback that pressing a button would have multi-second delays):

This resolves the performance issue mentioned in #331. (though I still recommend being cautious about placing frequently-updating state in the Redux store, the overall UI responsiveness is acceptable now)

### DIFF
--- a/packages/app/app/components/LibraryView/index.js
+++ b/packages/app/app/components/LibraryView/index.js
@@ -9,10 +9,12 @@ import {
   List,
   Progress,
   Segment,
-  Table
+  Table,
+  Ref
 } from 'semantic-ui-react';
 import { useTranslation } from 'react-i18next';
 import _ from 'lodash';
+import ReactList from 'react-list';
 
 import Header from '../Header';
 import TrackRow from '../TrackRow';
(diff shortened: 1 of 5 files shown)


nukeop:
Great job, thank you for contributing.

PR #551, opened 2019-11-10
Fix album height on artist page

Zalaxx:
The height of the album on a artist page was too high if there were juste some albums. 

| Before |  |
| --- | --- |
| After |  |


Venryx:
But what if some users like the appearance of the before image better? (jk)

nukeop:
Nice, thanks. I'm overhauling the album cards anyway and putting them in storybook, but this can be used in the meantime.

PR #552, opened 2019-11-12
Fix testing with other nuclear packages

badsketch:
There's an issue when trying to create integration or just general tests in `nuclear/app` and I want to use components from `@nuclear/ui`. 

ex. `example-test.js`
```
import React from 'react';
import chai from 'chai';
import spies from 'chai-spies';
import { shallow } from 'enzyme';
import { describe, it } from 'mocha';

import QueueMenuMore from '../app/components/PlayQueue/QueueMenu/QueueMenuMore';
import {QueueItem} from '@nuclear/ui' //<-- OFFENDING LINE
```

Currently, it won't work because babel isn't properly compiling those package files. 

Screen Shot 2019-11-12 at 12 31 08 AM

After looking it up, it seems to be an issue with `node_modules` not being ignored, but I see we're doing that in our `babel.config.js` file. Anyways, after looking at https://github.com/babel/babel/issues/9511, I added babel/register to our testHelper instead of our `--require @babel/register` through npm scripts. Now it works as expected, which will allow us to import components from existing nuclear packages in the monorepo.

### DIFF
--- a/packages/app/package.json
+++ b/packages/app/package.json
@@ -14,7 +14,7 @@
     "build:electron:linux": "webpack --progress --colors --env.LINUX=true --config=webpack.config.electron.prod.js",
     "build:electron": "webpack --progress --colors --config=webpack.config.electron.prod.js",
     "build:docker": "docker build -t nuclear .",
-    "test": "mocha --require ./test/testHelper.js --require @babel/register --require regenerator-runtime --timeout 10000 --prof --recursive",
+    "test": "mocha --require ./test/testHelper.js --require regenerator-runtime --timeout 10000 --prof --recursive",
     "pack": "electron-builder --dir -c.extraMetadata.main=dist/bundle.electron.js",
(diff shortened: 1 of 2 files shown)


nukeop:
Sorry, but I already did almost the exact same thing on a branch I'm working on now: https://github.com/nukeop/nuclear/commit/fe93b483e55b43e41f221b0d1ad95cef65f1f675

badsketch:
Oh cool, good sanity check for me then 👍

badsketch:
Hey, what's the best practice if I need this change in my branch? If I think I want to submit a PR before your branch is merged into master, do I copy over your edits fe93b483e55b43e41f221b0d1ad95cef65f1f675 manually? Or should I just wait until your feature is complete?

nukeop:
You can pull from `feature/album-grid`, and open a PR to that branch.

PR #554, opened 2019-11-23
(#536) player progress bar not clearing

### LINKED ISSUE #536: Clearing queue doesn't clear song progress bar
If a song is playing and the queue is cleared, the pink progress bar is still left over where instead it should disappear.

Gonna go ahead and fix it. I wonder if this is a regression? How come I feel like this wasn't happening before. Tried looking at commit history and don't see anything that might have affected it, but I didn't dig too deeply. Feel like I'm going crazy haha. I'll try to add some tests for it.

badsketch:
resolves #536 
- clearing queue resets the progress bar
- removing last song from queue resets the progress bar
- fix back and forward buttons not being disabled properly


packages/app/app/actions/player.js
```diff
@@ -36,6 +36,14 @@ export function togglePlayback(currentState) {
   };
 }
 
+export function resetPlayer() {
+  return dispatch => {
+    dispatch(updatePlaybackProgress(0,0));
+    dispatch(pausePlayback());
```
badsketch: commenting this out allows the test the pass successfully because ipcrenderer is undefined

nukeop:
Ok, so the best way would be to inject `sendPaused` into `pausePlayback`. Just change it to be `export function pausePlayback(sendPaused)`, and remove the import. You'll also need to pass `sendPaused` in `resetPlayer` and accept it as an argument there as well. While you're at it, you can also give `sendPlay` the same treatment.

Then, in the actual component, import the required function from the file it's defined in, and provide it as a prop using recompose's `withProps`. Export the augmented component as a default, but also export the unaugmented version as a named import. You can import that in the test, and once again use `withProps` to inject a mocked version of `sendPaused`. This way you can also test if that ipc message is being sent correctly. I realize it's quite complex, so let me know if you'd like me to illustrate this better.

badsketch:
Thanks for the detailed response!

If I understand this method correctly, won't the refactor impact many components, especially if I add the same treatment with `sendPlay`? `sendPlay` alone is used in 3 components and is also dispatched in 2 thunks, which themselves are in separate components. I would then need to recompose each of those affected components. Would this be considered obfuscating the code just to allow this test to run?

nukeop:
If you want you can only do this for `sendPaused` to reduce the complexity of your PR, since we don't necessarily need to do this for other methods right now. Dependency injection is a good and common pattern for making code testable, and as a side effect enables us to easily replace the ipc module with something else should we wish so in the future. 

If nuclear used `redux-saga` we could even simply `put` another action and run a saga in response to that action which would then call ipc, decoupling these two things.

badsketch:
going to make new PR to master when ready. thanks for all the help!

Issue #557, opened 2019-11-26
scroll up when click music card


sglee487:
I found that new music content's scroll is maintained when clicked other music card.
So I add an action to reset scroll when clicked.
 
![Record_2019_11_27_01_14_45_712 mp4_20191127_013214](https://user-images.githubusercontent.com/33611003/69652965-ce085900-10b5-11ea-8c36-c8ee78abfb47.gif)

nukeop:
Thanks for reporting. Did you finish working on this?

sglee487:
Yes. Could I create pull request?

nukeop:
Sure, feel free. In the future, don't even create an issue if you fix something, just create a pull request directly.

PR #558, opened 2019-11-27
add scroll reset action

sglee487:
It's related #557 issue.
add scroll reset code in onclick event


nukeop:
Thanks for solving this, it will work for now but I'll make a note to reference the DOM node in a more robust way in case we change the class name in the future.

PR #564, opened 2019-12-10
(#536) Clear progress bar when queue is cleared

badsketch:
Follow up PR to comments in #554 about treating `sendPaused` as an injectable. 

So I tried refactoring it like the way you suggested and was able to move the test along in terms of mocking the ipc.send out successfully. 🙂

The part I'm stuck at now is I'm unable to dispatch my action when I simulate click. Hoping for some more feedback/guidance if you have time.

### DIFF
--- a/packages/app/app/components/PlayQueue/QueueMenu/QueueMenuMore/index.js
+++ b/packages/app/app/components/PlayQueue/QueueMenu/QueueMenuMore/index.js
@@ -18,6 +18,7 @@ const addTrackToPlaylist = (updatePlaylist, playlist, track) => {
 const QueueMenuMore = ({
   disabled,
   clearQueue,
+  resetPlayer,
   savePlaylistDialog,
   addFavoriteTrack,
   updatePlaylist,
@@ -33,7 +34,7 @@ const QueueMenuMore = ({
     <Dropdown item icon='ellipsis vertical' className={styles.queue_menu_more} disabled={disabled}>
       <Dropdown.Menu>
         <Dropdown.Header>{t('header')}</Dropdown.Header>
-        <Dropdown.Item onClick={clearQueue}>
+        <Dropdown.Item onClick={() => {clearQueue();resetPlayer();}}>
           <Icon name='trash' />
           {t('clear')}
         </Dropdown.Item>
--- a/packages/app/app/components/PlayQueue/QueueMenu/QueueMenuMore/index.js
+++ b/packages/app/app/components/PlayQueue/QueueMenu/QueueMenuMore/index.js
@@ -17,16 +17,13 @@ const addTrackToPlaylist = (updatePlaylist, playlist, track) => {
 
 export const QueueMenuMore = ({
   disabled,
-  clearQueue,
-  resetPlayer,
   savePlaylistDialog,
-  addFavoriteTrack,
   updatePlaylist,
   playlists,
   currentItem,
-  sendPaused,
   handleAddToDownloads,
-  handleAddFavoriteTrack
+  handleAddFavoriteTrack,
+  handleClearClick
 }) => {
   const { t } = useTranslation('queue');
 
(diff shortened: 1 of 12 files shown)


packages/app/app/components/PlayQueue/QueueMenu/QueueMenuMore/index.js
```diff
@@ -115,5 +118,8 @@ export default compose(
         });
       }
     }
+  }),
+  withProps({
+    sendPaused: sendPaused
```
nukeop: Use `withProps({ sendPaused })`

packages/app/app/components/PlayQueue/QueueMenu/QueueMenuMore/index.js
```diff
@@ -33,7 +34,7 @@ const QueueMenuMore = ({
     <Dropdown item icon='ellipsis vertical' className={styles.queue_menu_more} disabled={disabled}>
       <Dropdown.Menu>
         <Dropdown.Header>{t('header')}</Dropdown.Header>
-        <Dropdown.Item onClick={clearQueue}>
+        <Dropdown.Item onClick={() => {clearQueue();resetPlayer(sendPaused);}}>
```
nukeop: You can use `withHandlers` in `compose` for this, e.g.

```javascript
handleClearClick: ({ clearQueue, resetPlayer, sendPaused }) => () => {
  clearQueue();
  resetPlayer(sendPaused);
}
```

Just remember to move your `withProps` to the top inside the compose function so that the handler has access to `sendPaused`.

nukeop:
Thanks, your code is very nice. I'll try debugging this and get back to you.

packages/app/test/integration/queueplayer.test.js
```diff
@@ -0,0 +1,80 @@
...
+        
+        const store = createStore(
+            PlayerReducer,
+            initialState,
+            applyMiddleware(thunk)
+        )
+
+        return store;
+    }
+    it.only('resets the player if the queue is cleared', () => {
```
nukeop: Don't forget to remove `only` before yourcommit

nukeop:
Ok, so it's fairly easy. You have to inject the spy using `withProps`, just like you inject the actual implementation in the composed component. I put my edit in a gist here:
https://gist.github.com/nukeop/276d1beef16f8295ac5e48fe0a7c56ae

To see what's been changed, you can download the gist and diff with your file, or just compare visually.

badsketch:
The feedback really helped and I really appreciate the gist upload!
I was even available to finish up the second integration test on my own and I can see the light at the end of the tunnel. 😌

One last issue is that, for `<QueueMenuMore />` I made this [requested change](https://github.com/nukeop/nuclear/pull/564#discussion_r356334418) and added to `withHandlers`. However, this then means that I can't import the unaugmented `<QueueMenuMore />` alone anymore. And I can't export the compose separately to manually augment it in the test, (I've exported it as `enhance` and it's what I've done for my second integration test ), because I won't be able to mock `sendPaused`. Do you know how I can approach this, like an easy way to override existing props or perhaps import/export the `withHandler` or `withProps` individually?

Again, thanks for all the guidance. Also happy holidays! 🎉

nukeop:
You can use `withProps` and `withHandlers` again to override existing props if you use the same names.

badsketch:
So I tested some things and I don't think `withProps` is overridable, for example in [this issue](https://github.com/emotion-js/emotion/issues/547).

I also created [this sandbox](https://codesandbox.io/s/zen-ishizaka-py3d6).

Using `withProps` and composing with another `withProps` will only use the original. 
So then, I tried exporting the `compose` so that it's applied second, but `withHandlers` will try to
use the original props??? So in my test, yes, the spy successfully is used, but `withHandlers` errors out because the original props are undefined. 

That's why I decided to change `withProps` in `<QueueMenuMore/>` to `defaultProps` and now both tests pass.

Thoughts?

nukeop:
You can separate the hocs into composable parts. You can try the same approach. I modified your sandbox to illustrate this: https://codesandbox.io/s/cranky-bash-67t5t

nukeop:
Btw, is it ready to merge now?

badsketch:
I wanted to do some final smoke testing, but after I recently pulled from upstream, I can't get dev to run. It'll sit at these messages and hang. Any idea where I can start troubleshooting this? =/

Screen Shot 2019-12-18 at 6 24 04 PM

Also did you want me to change the `defaultProps` in my latest commit into separated/composable HOC's? Otherwise, I think it's ready to merge. 👍

nukeop:
I think this is where the window is supposed to open. Try running `npm start` in `app` and `main` packages instead.

Issue #157, opened 2019-01-20
Playing songs is counter intuitive


paperbenni:
When you click on a song in an album, you expect it to play, that's how it works everywhere else and there's no reason to use something else. I would suggest to move the context menu that pops up instead to the right mouse button which does nothing at the moment. Also I am able to stack multiple context menus by clicking on multiple songs, that shouldn't happen either. 
https://imgur.com/a/DhOcJUp
I'm using the version from the latest master commit on archlinux. 

By the way I think this app has incredible potential, keep up the amazing work :)

nukeop:
If I remember correctly the track number becomes a play button when you hover over the track, and clicking it does what you want.

paperbenni:
No it doesn't. And remapping the menu would still be nice.

nukeop:
If it doesn't then it's a good idea to make it do so then, maybe I implemented this before I rewrote it from scratch and forgot. I planned replacing right clicks with a custom context menu at some point too.

trekiteasy:
I do agree I would suggest : 
- to harmonize all track interactions (in dashboard / artist view / playlist view / album view / tag view)
- to replace the number button (or the track icon) by a play button on hover
- a click on the play button would : add the track to the current queue just after the current song & play this new song right over
- move the context menu to the right click action on a track
- add a shortcut : double-click to add the track at the end of the queue
What do you think about it ?

nukeop:
The previous version had two extra buttons on hover, one "play now" which clears the queue, adds the selected song, and starts playing it, and a plus sign button which pushes the song to the end of the queue, similar to how "play now" and "add to queue" work elsewhere. I agree that the popup for tracks should be unified and given its own component (maybe make it accept a list of additional buttons in case we want to add some extra actions in some places in the future).

I remember trying to use right clicks for the context menu but I think React handled that poorly.

charjac:
it is actually possible to handle the right click with react
```jsx
<p>onContextMenu={console.log}>right click me</p>
```
and the popup from react-semantic-ui can have a controlled state:
https://react.semantic-ui.com/modules/popup/#usage-controlled

swrenst:
Is there still interest in making the track number a play button on hover? I have it running and currently, nothing happens when I hover over "A#", but I think this would improve the user experience. Would this be something I could try implementing?

nukeop:
Definitely, if you want to do it I would appreciate it.

PR #565, opened 2019-12-10
Issue #157: Implemented track # changes to play button on hover, click to play song

swrenst:
When looking at a track row, the user can now hover over the track number and the number will change to a play button icon. The user can click the icon and it will play that track.

Issue: https://github.com/nukeop/nuclear/issues/157


packages/app/app/components/TrackRow/index.js
```diff
@@ -25,6 +25,10 @@ class TrackRow extends React.Component {
     );
   }
 
+  componentWillMount () {
```
nukeop: `componentWillMount` is deprecated now. Could you move this state initialization to the constructor?
Or better yet, if you want to, you could use `withState` hoc from recompose.

packages/app/app/components/TrackRow/index.js
```diff
@@ -58,6 +62,11 @@ class TrackRow extends React.Component {
     }) < 0;
   }
 
+  changeNumberIcon(hovering) {
+    this.setState({ hoveringOnTrackNum: hovering });
+    console.log("changing state: ", hovering);
```
nukeop: please remove debug statements

packages/ui/test/enzyme.card.test.js
```diff
@@ -38,3 +41,21 @@ test('Should display given content', t => {
   const wrapper = shallow(<Card content='TEXT' />);
   t.is(wrapper.find('p').text(), 'TEXT');
 });
+
+test('Test track number hover shows play icon', t => {
```
nukeop: This is a file for card tests. Could you please move this test to the file for TrackRow component?

Note that this package (ui) has its own TrackRow, which will be merged with the one from the app package in the future. So you'll want to move your tests here: https://github.com/nukeop/nuclear/tree/master/packages/app/test

You could name it `enzyme.trackrow.test.js` or something similar.

nukeop:
Hi, thanks for your contribution. I like the idea and your implementation, and I made some comments so we can improve the quality of the code.

Btw, I know it's confusing but there are two separate TrackRow components, one in `app` and one in `ui`. You can safely ignore any changes in the ui package, and maybe even drop them altogether to keep the PR smaller. Please also remove the changes to the binary files if you can (I know they are being marked as modified but you don't need them).

PR #573, opened 2019-12-17
Refactor/app log flood

charjac:
This PR remove all the annoying logs at startup (exept the Buffer one didnt find its origin).
I integrate the range because of an annoying warning about componentWillReceiveProps.
react-smooth-image-loader wasnt maintained anymore and was responsible for problem with styled-component.

I fix the electron web security issue we could resolve (node integration is mandatory due to ipc, logger and plugins), and disable the warning (we should reactivate the warnings if we update electron).


nukeop:
Great, thanks. I'm going try fixing some bugs this week too.

PR #577, opened 2019-12-23
Fix crash & album display

Zalaxx:
Album display:

| Before |  |
| - | - |
| **After** |  |

App crash:

When trying to add a track to the queue, the function "search" is run on the local-library. This function call "byArtist", which get all the local track and try to find the wanted one. But if there is no local track, the function "byArtist" return null and make the app crash. A simple ternary operator returning an empty object fix that.


nukeop:
Very nice, thanks for fixing this.

PR #580, opened 2019-12-27
style(lint): run eslint in all packages and add husky + lint-staged

charjac:
lint all the files in packages, and add husky (local git hook) and lint-staged in order to prevent commit unlinted file. 
I can add prettier if you want.


charjac:
the push hook is maybe a little bit overkill.

nukeop:
>the push hook is maybe a little bit overkill.
It's useful but many people will disable eslint instead of fixing their code so they can push, and then we'll miss any potential problems.

charjac:
ok, do you want me to remove husky and eslint ?

PR #582, opened 2019-12-30
Fix/download

### LINKED ISSUE #572: Strange error when downloading
2019-12-16-223639_545x181_scrot

Some time after I start downloading a track, this error appears in a dialog box which reopens (or doesn't close) when I click OK. When this occurs the main window is no longer usable. After closing the application the file it's been downloading exists, but is cut off at the end. This doesn't appear to be localized to a particular track, but I haven't tested very many. I'm using the latest version from the `master` branch as of a few hours ago.

charjac:
fix #572 
Looks like if BrowserWindow is used as a super class, the window is not correctly registered by electron.
I remove the inheritence and add a private browserWindow property in the window service.


packages/main/src/controllers/settings.ts
```diff
@@ -42,11 +40,7 @@ class SettingsIpcCtrl {
 
   @ipcEvent('maximize')
   onMaximize() {
-    if (this.platform.isMac()) {
-      this.window.isFullScreen() ? this.window.setFullScreen(false) : this.window.setFullScreen(true);
-    } else {
-      this.window.isMaximized() ? this.window.unmaximize() : this.window.maximize();
-    }
+    this.window.maximize();
```
nukeop: Did you change the behavior for macs?

nukeop:
Great, thanks

PR #584, opened 2020-01-01
Feat/fileplayer

charjac:
in main package
 - add the ability to play a local file by passing its path as arg, and add mimeType association
```
nuclear myfile.mp3
```
 - refactor window event sender
 - add Tray service

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     "@typescript-eslint/eslint-plugin": "^2.13.0",
     "babel-eslint": "^10.0.3",
     "electron": "^4.2.11",
-    "electron-builder": "^21.2.0",
+    "electron-builder": "^20.44.4",
     "eslint": "^6.7.2",
     "eslint-plugin-node": "^9.1.0",
     "eslint-plugin-react": "^7.12.4",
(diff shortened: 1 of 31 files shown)


charjac:
I was also thinking about drag and drop on the queue to add tracks, what do you think ?

nukeop:
As in, dragging files from the outside into the player? That's a really good idea.

package.json
```diff
@@ -38,7 +38,7 @@
   "devDependencies": {
     "babel-eslint": "^10.0.3",
     "electron": "^4.2.11",
-    "electron-builder": "^21.2.0",
+    "electron-builder": "^20.44.4",
```
nukeop: why the downgrade?

charjac: better error log when trying to build locally, i can rollback this now

charjac:
I had a few things: 
- dragging files and/or folders in the queue add tracks to queue and their folder and metadata to the store
- minimal design ('-_-) for the queue when dragHover with files
- redirect on library page when a file in run at startup
- prevent multiple instance of electron being run

charjac:
For me its ok

nukeop:
Looks good to me, fantastic work as always.

nukeop:
I'll change a few things (droppable box styles and state added via HOC not class component) though

nukeop:
I also found a bug where dragging the same song twice causes the second copy to not load the source correctly

charjac:
ok i'll check this bug today :)

PR #588, opened 2020-01-03
Fix/locallibrary

charjac:
- automatically add tracks meta when adding a new local folder
- automatically remove tracks meta when removing a local folder
- improve logging (especially ipc)

I think this should be part of the next release, if you prefer i can PR in another branch.


nukeop:
Yeah I agree it should go to master

PR #590, opened 2020-01-05
Feat/offline

charjac:
- add offline support
  - restrict access to dashboard when no connection and redirect to library
  - hide the search bar when no connection
- simplify localfile management
- fix a visual bug when dragging a track over the queue
- fix run the builded app without args (typo in main.ts)

Maybe we could hide search and dashboard menu item when no connection ?


nukeop:
>Maybe we could hide search and dashboard menu item when no connection ?
Yep, we can focus on this more in the next release, let's keep changes minimal for now to minimize the risk.

Issue #581, opened 2019-12-27
typescript in other packages ?


charjac:
if all packages were in ts we could simplify dev and test env.
@nukeop What do you think ?

charjac:
build is failling on travis because of this (i remove the builded files from common package)

nukeop:
I'm all for it. If you see a good place to use typescript, feel free to do it whenever you contribute.

charjac:
typescript is cool everywhere :) (way better than proptypes for react)

nukeop:
Yeah, we can get rid of proptypes. I think we can replace them with a generic interface for each component.

charjac:
- [ ] app
- [x] common
- [ ] core
- [x] main
- [ ] ui

charjac:
@nukeop What about merging common and core package, i am really sure there is a difference between them.

charjac:
I start converting ui package. Do you have wip on it ?

nukeop:
>@nukeop What about merging common and core package, i am really sure there is a difference between them.
Yes, what's in common could have started as core modules really.

>I start converting ui package. Do you have wip on it ?
Not right now, but I wouldn't consider it a priority to convert everything right away. I want to release a new version soon so I'm focusing on fixing bugs and making small commits to master.

charjac:
ok with that. I will just merge core and common packages.
If you have in minds some bugs i could help with, don't hesitate :)

nukeop:
I'm just clicking through the whole program trying to find and remove stuff that looks wrong, got slightly broken for some reason, or is just embarrassing and not hard to fix. Things that could discourage first time users after the release.

nukeop:
Also one big thing I plan to finally get rid of is duplicate favorites - I don't have time to do that today, but I'll definitely do it this weekend and then I think we'll be ready for release. If you think you can be faster than me then you could take that too.

Btw, if you want to develop new features or do any major refactors, let's merge to a "dev" branch temporarily and merge it back to master after releasing.

charjac:
I fail to reproduce it locally.
How did you do to get duplicate favorites ?

nukeop:
I guess this can be closed now

Venryx:
> Not right now, but I wouldn't consider it a priority to convert everything right away. I want to release a new version soon so I'm focusing on fixing bugs and making small commits to master.

At what point would it be good to convert more of the code-base to typescript?

[...]

PR #592, opened 2020-01-06
fix(app): fix duplicate favorites

charjac:
#581


nukeop:
yess this is exactly what caused it, thanks so much!

Issue #595, opened 2020-01-07
Queue context popups should be displayed on the left


nukeop:
The popups currently obscure the queue, they should be displayed to the left of the clicked track. And there's a small issue with the arrow where the border is drawn inside it.

charjac:
This one is pretty thought.
I tried a few things, it looks like the semantic-ui Popup is fucked up. (not completely sure about it)
The only things that works is a dirty negative margin ..... and a top override (!important).

nukeop:
I used a `target` prop in `AlbumPreview`: 
https://github.com/nukeop/nuclear/blob/a95f3b454097672b1664e77e58281849016649d1/packages/ui/lib/components/AlbumPreview/index.js

Maybe we could do something like this. I want the popups to always be in the same place, exactly to the left of the queue item.

charjac:
thx !!! I will give this a try

PR #604, opened 2020-01-12
fix(ui): fix queue popup behaviour

charjac:
=== UI ===
 - move the QueuePopup component to the ui package.
 - fix positioning using getBoundaryClientRect.
 - handle stream change but it break in redux action.
 - don't display the popup when loading (didn't achieve to fix positioning when loading ...)


charjac:
fix #595

charjac:
I will add tests and a story

packages/ui/lib/components/QueuePopup/index.js
```diff
@@ -0,0 +1,102 @@
...
+
+  return (
+    <Popup
+      className={styles.queue_popup}
+      trigger={<div onClick={handleTriggerClick}>{trigger}</div>}
+      open={isOpen}
+      onClose={handleClose}
+      onOpen={handleOpen}
+      hideOnScroll
+      position='right center'
```
nukeop: shouldn't this be left center?

nukeop:
I can't reroll the stream with this change.

Btw, we could make the popup appear on right click, since double clicking moves the queue to the clicked track, and the popup interferes with that.

packages/ui/lib/components/QueuePopup/index.js
charjac: when i put "left center", it works fine on story book but fail on the app :/
when "right center", its the opposite :scream: :scream: :scream:

charjac:
Ok, will fix that.
No problem for the right click.

packages/ui/lib/components/QueuePopup/index.js
nukeop: Lol, this is horrible. I'll need to replace these popups with something better at some point, but let's stick to these for the time being.

charjac: I think the problem come from the app, on the storybook the popup act as intended without the target hack.
Maybe a css conflict ?

charjac:
I am not proud of this :/ but it works fine on the app, cant figure where does this position problem come ....
I still need to fix the stream change

charjac:
what about preventing other popup when one is displayed ?

nukeop:
I think this was already handled? If it's broken again, it'll be important to fix that.

charjac:
ok will fix this

nukeop:
Thanks, I appreciate this. After this I think we'll be ready for release.

nukeop:
Ok, awesome. I need to fix the broken build soon.

Issue #583, opened 2019-12-31
Show top contributors in help screen


nukeop:
There's a short "about" section that displays some basic info.
It would be nice if we could get a list of contributors from Github and display the top 10 (maybe more) most prominent contributors underneath as a way of saying thanks. This could include avatars, usernames, and links to profiles that open in the browser, plus maybe how many commits/LOC a person contributed.

charjac:
I remember a node script written by kentcdodds who where doing this for Readme.md. I will try to find it on github.

mahimshridhar:
hey @nukeop can i work on this ?

nukeop:
Sure, feel free to.

Burzo:
Hello there, my bad, I'm a bit late to the party. I started working on this yesterday, but didn't know I'd need to reserve it like this - just started with open source contribution. Should I continue and submit my pull request @nukeop, once finished? Mostly styling left, most of the logic is implemented.

Sorry if I'm causing confusion...

nukeop:
@mahimshridhar , have you started working on this? Ultimately you'll have to work out what parts each of you can do, if any.

@Burzo , you can create a pull request now, and mark it as `do not merge` until you're finished, this way it can be seen you're working on this (and you can also reference this issue by adding `#583` to the pull request thread so it shows up here).

mahimshridhar:
I have not started working on this. and @Burzo, you can continue if you want to.

Burzo:
Thank you both, @nukeop  and @mahimshridhar . I'll just create a pull request once its finished then, as to not confuse everything further.

charjac:
I know this has already been implemented, but i just find this, and it looks interesting, so i put it here:
https://github.com/all-contributors/all-contributors

nukeop:
I know about this thing. We keep all assets in the repo though, so people who contribute translations, docs, and graphics, are all de facto contributors, so no need to acknowledge them separately.

PR #606, opened 2020-01-14
Add GitHub contributors #583

Burzo:
Hey,

I created a seperate Redux reducer and action to handle the fetch logic and update the global state accordingly. The contributors are only fetched once for now (at the initial app render), since I didn't see the need to update them later on.

I also added some components to handle the rendering logic and used the Semantic UI to display the Github user's information.

Hope it looks good. Let me know what you think.


nukeop:
Thanks for contributing. I'll merge your changes to the dev branch, and we'll merge dev into master after the release (so your changes won't be a part of the next release, but they'll be in the one after that).

PR #614, opened 2020-01-19
feat(main): add support for discord rich presence

### LINKED ISSUE #607: Discord rich presence support
It could be nice for Discord users to show to their friends that they're using Nuclear. There are some easy to use node libraries for this.

charjac:
=== MAIN ===

the DISCORD_CLIENT_ID env var is simply a discord app clientId i just create.
You can simply replace this with one created with your account.
If you do this, you will have to upload an assets on your app and call it logo.

fix #607


nukeop:
Wow you work fast

nukeop:
Merge from master when you have the time so we can see if CI passes.

PR #619, opened 2020-01-20
fix(local): use event-stream package to handle fpcalc child process

### LINKED ISSUE #618: Fpcalc is broken
When Fpcalc fails, it throws an uncatched error.

charjac:
I put back the fpcalc code as it was before the refacto (with event-stream).
What i did fails sometimes and trigger a blocking alert popup.

fix #618


nukeop:
Nice, thanks for resolving this quickly.

Issue #600, opened 2020-01-11
Local library index optimization


nukeop:
The index is huge even if we only have a couple of albums, and it looks like most of that is taken by the album art, which is replicated in each individual track. This will be the lowest hanging fruit to be handled when optimizing the index.

tracker1:
Should probably consider sqlite, and for the album art, referencing a hash of the image, so the same reference can be used.

Also, not sure if child directories are added when a directory is added, I tried adding my audio directory (thousands if files, several gb of music) and the app blew up.

nukeop:
Yeah, that's probably because of the index performance.

charjac:
Hi @tracker1, is it possible for you to build the branch fix/local-images, and test if the app still crash when parsing your library please ? (I have not enough mp3 to do "stress test", and i am too lazy to dl the whole pirate bay for my test).

PR #625, opened 2020-01-20
Fix/freeze

### LINKED ISSUE #597: Weird loading screen behavior
I think this is caused by `preload.js`.

The initial loader is not animating, then progresses directly to the program without the fade in that used to play over 1s.

If you comment out `render(App)` in `index.js`, you can notice that the logo is initially larger, then becomes smaller and starts animating.

charjac:
=== App ===
dont scan local folders and extract meta each startup, simply use the
stored state.

=== Main ===
dont scan all folder and subfolder when playing a single file, only get
its meta.
same when drop files on the queue.

fix #597


nukeop:
Awesome, this is a good step towards #600 as well

PR #626, opened 2020-01-21
Fix/local images

### LINKED ISSUE #600: Local library index optimization

charjac:
I am not sure at all, if it is the good solution, but it greatly improve perfs.

=== Main ===
store local files image on file system instead as in a json file as base64 string
delete unused thumbnail when app quit and when related folder is removed

fix #600

### DIFF
--- a/packages/app/webpack.config.js
+++ b/packages/app/webpack.config.js
@@ -52,7 +52,7 @@ module.exports = (env) => {
       ignore: [/node_modules/]
     }
   };
-  const contentSecurity = 'connect-src *; style-src \'unsafe-inline\' https:; font-src https: data:; img-src https: data:;';
+  const contentSecurity = 'connect-src *; style-src \'unsafe-inline\' https:; font-src https: data:; img-src https: data: file:;';
   const plugins = [
     new HtmlWebpackPlugin({
       meta: {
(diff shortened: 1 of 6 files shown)


nukeop:
Much better. Will that duplicate the covers? 

I was thinking we could use a hash generated from the cover as the filename, didn't have the time to read the code yet (work).

charjac:
the cover are not duplicated anymore.
For the filename i use the album name, but i dont handle the case where there is an image but no album name, (in this case the hash could be the solution)

nukeop:
I'll compare the changes with what we have now.

charjac:
Maybe we should integrate sqlite in this PR ?

nukeop:
If the scope of the changes isn't too great, feel free to include it in this PR

nukeop:
Should I merge this? Or do you want to add that feature to this?

charjac:
You can merge it, i will not have enough time today to implement it. Will work on it next week .

nukeop:
Thanks for your effort. It reduced the index size from 77 mB to 193 kB for me. Amazing work.

nukeop:
This also seems to have fixed that bug with duplicated tracks in local library.

PR #628, opened 2020-01-22
Feat/invidious

### LINKED ISSUE #627: invidious api
What about using invidious api =>  https://github.com/omarroth/invidious
and use youtube plugin as a fallback when something fail.

https://github.com/FreeTubeApp/FreeTube use it and it works just fine.

This could be configurable in the settings:
 - invidous instance url
 - enable/disable

charjac:
=== App ===
 - move  all env variable to root .env file
 - add invidious url in youtube settings section with default to https://invidio.us
 - add invdious plugin
 - if invidious fail, fallback with youtube plugin
 - use invidious by default

fix #627


packages/app/app/containers/HttpApiUrl/index.js
```diff
@@ -15,7 +15,6 @@ const HttpApiUrl = ({ settings }) => {
   return (
     <a
       className={styles.http_api_url}
-      href='javascript:void(0)'
```
nukeop: I think this href was needed for some cosmetic reason, did you notice anything different when removing it?

packages/app/app/containers/SoundContainer/index.js
```diff
@@ -130,6 +137,14 @@ class SoundContainer extends React.Component {
     });
   }
 
+  handleError() {
```
nukeop: hacky, maybe it'd be better to store a fallback/default plugin in redux (like we store the selected one) and use it when errors occur?

packages/app/app/locales/en.json
```diff
@@ -191,6 +191,7 @@
     "user": "User:",
     "youtube": "Youtube",
     "yt-api-key": "Youtube API Key",
+    "invidious-url": "Invidious instance url",
```
nukeop: This is ok for now but I'm making a mental note to introduce per-plugin settings at some point and configure stuff like this in the plugins section

packages/app/app/plugins/stream/InvidiousPlugin.js
```diff
@@ -0,0 +1,69 @@
+import logger from 'electron-timber';
+
+import StreamProviderPlugin from '../streamProvider';
+import * as Invidious from '../../rest/Invidious';
+import YoutubePlugin from './YoutubePlugin';
+
+class InvidiousPlugin extends StreamProviderPlugin {
```
nukeop: I think I'll define these things properly in Typescript soon

packages/app/app/locales/en.json
charjac: sound great

packages/app/app/containers/SoundContainer/index.js
charjac: sure, one more select in the plugins view ? or something hard coded ?

packages/app/app/containers/HttpApiUrl/index.js
charjac: i replace it with css style 'cursor: pointer;'. it remove the annoying warning in console

packages/app/app/plugins/stream/InvidiousPlugin.js
charjac: I was also thinking about moving reducer and action constants in @nuclear/core, without any electron reference (mobile app in mind).
What do you think about it ?

packages/app/app/containers/HttpApiUrl/index.js
charjac: I replace it with css style
```css
cursor: pointer;
```
it remove the annoying warning

packages/app/app/plugins/stream/InvidiousPlugin.js
nukeop: We can still use reducers on mobile (I think) but decoupling electron dependencies is a step in a good direction

packages/app/app/containers/SoundContainer/index.js
nukeop: Can be hardcoded somewhere, I just don't like depending on comparing string names

charjac:
I add a few things :) : 
 - queue items have two more optionnal property:
   - failed: bool
   - selectedStream: string
 - new property fallback on StreamProviders plugins
 - compute the stream depending on these 3 params and global selected streamProviders
 - fix the queue Popup stream selector (without breaking the alternate stream this time)

nukeop:
What happens if a queue item is failed?

charjac:
it switch to the fallback define in the used StreamProviders plugin

nukeop:
Ok, this is a very good solution

nukeop:
Do you know why the tests are failing?

charjac:
no will check locally

charjac:
same error locally, i will fix it

charjac:
queureducer now import an actions from plugins, and plugins actions require something from electron package. Mocha didnt like it. electron-mocha seems to be ok.

charjac:
ok its fucked on ci .... will find something

nukeop:
Maybe remove the electron dependency and inject at runtime.

charjac:
does plugins still need to access the store from window ?

nukeop:
It's for the arbitrary plugins loaded by users. They get the app instance, the store, and the React api. But all this can be passed to the action as arguments and kept out of the action definitions.

charjac:
finally i just mock electron related modules and test passes

PR #629, opened 2020-01-22
Refactor/i18n

### LINKED ISSUE #520: Extract localization-related code as a separate package @nuclear/locales
As the title says. Reminder for myself.

charjac:
=== App ===
move i18n related code to its own package
add a script which run at each commit, sync the json locale files and edit docs/i18n.md with the missings translation per language. 

it will help contributors a lot.

fix #520


nukeop:
This is great. Translations really helped popularize Nuclear in many countries, especially ones that don't use Latin alphabet (like China).

I think it would also be useful to automate editing the readme section with translator credits in the future.

PR #635, opened 2020-01-24
'Play All' button for FavoriteTracksView

kangwritescode:
The button has the same styles and markup as the 'play' button in AlbumView. The functionality is also patterned similarly as the 'play' button in AlbumView. Please let me know if any changes are needed!


nukeop:
Thanks, looks very good.

PR #639, opened 2020-01-27
fix(windows): use hash for all local thumbnail

### LINKED ISSUE #636: Improperly escaped thumbnail paths cause local library scan to fail
- **Nuclear version**: 0.6.2 commit b2dff07b25808da9a1658eddd9ec716d4c60b042
- **NodeJS version**: 12.14.0
---

Since #626 was merged and thumbnails get stored on disk instead of JSON, certain album names can cause the local library scan to throw an error and fail. I believe this can happen due to album names containing quotes. This is the error that gets thrown once it hits such a file: 
```bash
   ipc api ›  ==> get-localfolders
   ipc api ›  ==> get-metas
   ipc api › once ==> started
  http api › nuclear http api available on port 3000
   ipc api ›  ==> connectivity : boolean
   ipc api ›  ==> set-localfolders : Array<string>
   ipc api › error in event set-localfolders => ENOENT: no such file or directory, open 'C:\Users\XXXXX\Desktop\nuclear\packages\main\build\thumbnails\House set of "Subterranean Animism".jpeg'
   ipc api ›  ==> close
```

As you can see the file it can't find has quotes in the album name. I have attached the problematic file so you can verify this bug.

Tagging @charjac as he authored the pull request.
[sample.zip](https://github.com/nukeop/nuclear/files/4112937/sample.zip)

charjac:
fix #636

### DIFF
--- a/packages/main/src/services/local-library/index.ts
+++ b/packages/main/src/services/local-library/index.ts
@@ -67,7 +67,7 @@ class LocalLibrary {
     if (common.picture && common.album) {
       imagePath = await this.persistCover(
         common.picture[0].data,
-        common.album || crypto.createHash('md5').update(path.basename(filePath)).digest('hex'),
+        crypto.createHash('md5').update(path.basename(common.album || filePath)).digest('hex'),
         common.picture[0].format
       );
     }


docs/i18n.md
```diff
@@ -1064,4 +1064,4 @@
 
 
 
-generated the 25/1/2020
\ No newline at end of file
+generated the 27/1/2020
```
charjac: this is annoying ... will find something

charjac: #640 fix this

PR #641, opened 2020-01-29
added last fm favorite importing

### LINKED ISSUE #631: Import favorites from last.fm
It was suggested on Mastodon by https://mstdn.io/@bueno that it would be useful to be able to import favorites from Last.fm. We already allow connecting accounts for scrobbling so it shouldn't require too much additional work.

o-ali:
Added feature to import Last.Fm favorites into the application.
On the settings screen, when the user is logged in a button appears that can be clicked to fetch the "LovedTracks" from last.fm api and saves them to the local favorites storage.
https://github.com/nukeop/nuclear/issues/631
This fixes #631


packages/app/app/actions/importfavs.js
```diff
@@ -0,0 +1,20 @@
+import { LastFmApi } from '@nuclear/core';
+import { store } from '../persistence/store';
+import globals from '../globals';
+const lastfm = new LastFmApi(globals.lastfmApiKey, globals.lastfmApiSecret);
+// first function is used to extract the total number of loved tracks by sending a request to the
+// last.fm api for 1 song. (total number of loved tracks is included in  the response fields)
+// that number is then used to fetch all the songs in the second function which also adds them to the favorites
+export function ImportFmFavorites() {
```
nukeop: This function isn't needed, since you can use fetchAllFmFavorites with totalTracks=1

packages/app/app/actions/scrobbling.js
```diff
@@ -127,3 +136,38 @@ export function updateNowPlayingAction(artist, track, session) {
...
+      dispatch({
+        type: LASTFM_ENABLE_IMPORT,
+        payload: null
+      });
+    };
+  }
+
+}
+
+export function FmFavUpdateMsg(msg) {
```
nukeop: We don't need to store this message persistently, or in redux, since it can be generated in the component.

nukeop:
Hi, thanks for the contribution, very nice work. I left some comments suggesting how to improve the code and refactor it into a simpler form. If anything is unclear, or if you need any guidance, just let me know and I'll be very happy to help.

o-ali:
Hi,
Thanks for the feedback. I've made the changes and pushed them as requested. Let me know if you need anything else changed, or if I didn't make the changes as you intended.

Sidenote: I noticed an issue where the user stays logged in even after logging out. Reproduce by logging in -> logging out -> close app -> open app -> (user is logged back in). It seems to occur because the log out function doesn't reset the storage variables for the name, sessionkey etc.. It's a simple fix. Do you want me to add it to this pull request or create another issue and branch ?

packages/app/app/components/Settings/index.js
```diff
@@ -19,15 +19,19 @@ import GithubSettings from './GithubSettings';
...
 
 const volumeSliderColors = {
   fillColor: { r: 248, g: 248, b: 242, a: 1 },
   trackColor: { r: 68, g: 71, b: 90, a: 1 },
   thumbColor: { r: 248, g: 248, b: 242, a: 1 }
 };
-
 @withTranslation('settings')
 class Settings extends React.Component {
+  state = {
```
nukeop: State in components should never be used.
It's better to use this instead:

https://github.com/acdlite/recompose/blob/master/docs/API.md#withstate

But in this case, we don't need this variable at all. If you follow the pattern I described below, we'll have access to loading/success/error state of this data, and we'll be able to choose a message according to that.

packages/app/app/components/Settings/index.js
```diff
@@ -77,10 +81,87 @@ class Settings extends React.Component {
...
+        <span>
+          {this.props.t('fmfav-msg')}
+        </span>
+        <Spacer />
+        <span>
+          <strong>{this.state.lastFmFavImportMessage}</strong>
+        </span>
+        <Spacer />
+        {(
+          <Button disabled={!lastFmFavImportStatus} loading={!lastFmFavImportStatus} onClick={() => {
```
nukeop: This is way too complex for a handler. You should move all this logic to an action, but refactored to be simpler.

You can follow the pattern we have in many other places, with actions ending in `_START`, `_SUCCESS`, and `_ERROR` to update the redux state with data accordingly. You can avoid most of this branching logic this way. And then you can join these simple actions into a sequence using a thunk.

Let me know if I should make this clearer or if you need an example.

nukeop:
Nice, almost there. I think we can simplify the handler that downloads the data and updates the message.

o-ali:
Thanks for the continued feedback! 
I've updated the code to make it better match the existing format for other similar actions and used the _SUCCESS _ERROR actions as you suggested. It's definitely looking better than it was before 😄 . I was too sure what you meant by thunk so if you have more comments/suggestions on more changes to make please let me know.

packages/app/app/actions/importfavs.js
```diff
@@ -0,0 +1,99 @@
...
+        lastFmFavImportMessage: ''
+      }
+    });
+  };
+}
+
+function FmFavError() {
+  return {
+    type: LASTFM_FAV_IMPORT_ERROR,
+    payload: null
```
nukeop: You don't have to specify a payload if you don't want to pass anything.
This could be turned into a single line:
```javascript
const fmFavError = () => ({ type: LASTFM_FAV_IMPORT_ERROR })
```

packages/app/app/actions/importfavs.js
```diff
@@ -0,0 +1,99 @@
...
+const lastfm = new LastFmApi(globals.lastfmApiKey, globals.lastfmApiSecret);
+
+export const FAV_IMPORT_INIT = 'FAV_IMPORT_INIT';
+
+export const LASTFM_FAV_IMPORT_START = 'LASTFM_FAV_IMPORT_START';
+export const LASTFM_FAV_IMPORT_SUCCESS_1 = 'LASTFM_FAV_IMPORT_SUCCESS_1';
+export const LASTFM_FAV_IMPORT_SUCCESS_FINAL = 'LASTFM_FAV_IMPORT_SUCCESS_FINAL';
+export const LASTFM_FAV_IMPORT_ERROR = 'LASTFM_FAV_IMPORT_ERROR';
+
+export function FavImportInit() {
```
nukeop: If you only want one action, no need to use dispatch, just return the action object directly.

packages/app/app/actions/importfavs.js
```diff
@@ -0,0 +1,99 @@
...
+      lastfm.getNumberOfLovedTracks(storage.lastFmName, 1)
+        .then((resp) => resp.json())
+        .then(req => {
+          if (!req.lovedtracks) {
+            dispatch(FmFavError());
+            return;
+          }
+          let totalLovedTracks = req.lovedtracks['@attr'].total;
+          dispatch(FmSuccess1(totalLovedTracks));
+          lastfm.getNumberOfLovedTracks(storage.lastFmName, totalLovedTracks)
```
nukeop: Don't nest promises. You can return a promise, and keep the `.then()` chain going.

This way you also only have to use one `catch()` callback.

nukeop:
Very good, the code is much better now.

A thunk is a composite action that uses redux-thunk: https://github.com/reduxjs/redux-thunk

So basically what you wrote. I made some comments about how to make your actions simpler and shorter, if you'd like to commit some more.

o-ali:
Made the suggested changes, let me know how it's looking now.

packages/app/app/actions/importfavs.js
```diff
@@ -0,0 +1,73 @@
...
+          return lastfm.getNumberOfLovedTracks(storage.lastFmName, totalLovedTracks);
+        })
+        .then((resp) => resp.json())
+        .then((req) => {
+          let counter = 0;
+          req.lovedtracks.track.forEach(favtrack => {
+            FavoritesActions.addFavoriteTrack(favtrack);
+            counter += 1;
+          });
+          dispatch(FmSuccessFinal(counter));
```
nukeop: Why not use `req.lovedtracks.track.length` instead of the counter?

nukeop:
Cool, this is pretty much in a state that can be merged. I think we could remove the counter variable and use the array length instead, and we'll be done.

packages/app/app/components/Settings/index.js
```diff
@@ -77,10 +76,46 @@ class Settings extends React.Component {
...
+        <span>
+          {this.props.t('fmfav-msg')}
+        </span>
+        <Spacer />
+        <span> {(lastFmFavImportMessage && (
+          <strong>{lastFmFavImportMessage}</strong>
+        ))}
+        </span>
+        <Spacer />
+        {(
```
nukeop: No need for this `{(`

packages/app/app/actions/importfavs.js
o-ali: Good point 😅

packages/app/app/actions/importfavs.js
```diff
@@ -0,0 +1,71 @@
...
+      lastfm.getNumberOfLovedTracks(storage.lastFmName, 1)
+        .then((resp) => resp.json())
+        .then(req => {
+          let totalLovedTracks = req.lovedtracks['@attr'].total;
+          dispatch(FmSuccess1(totalLovedTracks));
+          return lastfm.getNumberOfLovedTracks(storage.lastFmName, totalLovedTracks);
+        })
+        .then((resp) => resp.json())
+        .then((req) => {
+          req.lovedtracks.track.forEach(favtrack => {
```
nukeop: This fails if the user has no favorites, because there's no `track` property. You can either use lodash `_.get`, or check if `lovedtracks` is undefined.

nukeop:
I found some more issues. Works as intended in the happy path case (more than 0 favorites).

Btw, as for the messages that are displaying when the tracks are importing, we'll want them to be translatable eventually, but I will handle that after your PR is merged.

o-ali:
I'll work on that. While looking in to this I noticed the api also has a limit of 1000 songs per page (so per request). Assuming some people might have more than that I can add logic to loop over and request several 'pages' which will let us get all the loved tracks.

nukeop:
Let's keep it simple for now, we'll see if there are any cases where somebody wants to import more than a 1000 and then we can implement it.

o-ali:
Ok np. I've added some error checks for 0 or more than 1000 song as well as general checks for validating the object.

nukeop:
Thanks for doing a few iterations on this code, we've got a really nice feature now thanks to you.

o-ali:
It was a learning experience for me too 😄 . Glad I could help.

PR #646, opened 2020-02-01
Added Swedish translation

PalleKarlsson:
This is a part of #345.

I've translated all the strings available in packages/i18n/src/locales/en.json and the README to Swedish.


nukeop:
Fantastic, thank you very much for your effort.

Issue #647, opened 2020-02-01
Wrap all text in i18next functions


PalleKarlsson:
Hi!

This is somewhat related to #345 

While adding the Swedish translation I noticed that there are several texts that are hard-coded and not going through i18next. I'm not sure if you have other plans for handling translations, or if it just is low-priority in relation to other issues. But I wanted to fix it, and just continued coding after finishing the translation. :)

I have a PR that I hoped of getting feedback on. You can find it [here](https://github.com/nukeop/nuclear/pull/648).

nukeop:
Nice catch, I probably didn't notice. We've integrated translations some time ago and over time we probably didn't wrap all strings properly. Thanks for fixing this.

odiaz95012:
Hello, is it okay if I work on this issue? It's my first time working on an open-source and it seems like a good place to start.

nukeop:
Ok, you can try hunting down all the strings that are still not translated. They should be added to `en.json` and you should use the hook provided by our i18n library to get them. Once they're in `en.json`, they will propagate to other languages.

PR #648, opened 2020-02-01
Added more strings to the translation engine

PalleKarlsson:
Related to #647

So for the most part I think I followed the same pattern as the already existing translations but there are a few places that I could use some feedback.


packages/app/app/reducers/equalizer.js
```diff
@@ -72,7 +63,7 @@ const getSelected = (selected, presets) => {
   if (selected.toLowerCase() === selected) {
     return selected;
   } 
-  const legacySelected = presets.find(preset => preset.label === selected);  
+  const legacySelected = presets.find(preset => preset.id === selected);  
   return legacySelected ? legacySelected.id : presets[0].id;
 };
```
PalleKarlsson: So from what I could see the label property is only used here, and is passed as the text to display in the EqualizerPresetList.

The id's follow the same naming conventions as the translation keys, so I removed the label property and used the id property as a translation key instead. You can see this in EqualizerPresetList.

But the legacySelected function scares me a bit. All the tests pass, and I couldn't find any errors while manually testing, but I'm new to this code so I might have missed something. What do you think?

packages/app/app/containers/TrackPopupContainer/index.js
```diff
@@ -17,6 +17,7 @@ const TrackPopupContainer = props => {
     trigger,
     track,
     artist,
+    t,
```
PalleKarlsson: Since this is a container I couldn't use the import { withTranslation } that is used in other places (or at least I don't think so).

So I added the translation-function in the parent component and passed it as a prop to here. Is there a cleaner way of doing it?

nukeop: Why not?

You can import `withTranslation` and `compose` from recompose, and then do
```javascript
export default compose(
  connect(
    mapStateToProps,
    mapDispatchToProps
  ),
  withTranslation('something')
)(TrackPopupContainer);
```

packages/app/app/reducers/equalizer.js
nukeop: I don't know if equalizer code is properly covered by tests. You can try testing it manually, there's only a handful of presets. I think your solution is pretty good in any case.

packages/app/app/containers/TrackPopupContainer/index.js
PalleKarlsson: I'm still kind of new to React so I haven't learned how everything works yet, so thanks for this suggestion! I'll implement it and then I'll take a look at all translation keys, I've mostly just kebab-cased the actual strings, but most other translation keys are more descriptive of their function. So I'll try to make them more consistent with the other keys!

PR #651, opened 2020-02-03
feat(locallib): use sqlite for local meta

charjac:
=== Main ===
sqlite and typeorm
model adaptation
new fpcalc binaries (the old ones were broken)

the orm is overkill ... but its clean.
there is still a few things to do, and it need to be strongly tested on all os.

Convert all thumb to webp with low quality settings using sharp. (need to be tested on windows and mac).
before this my 20 thumbs were 20Mo after only 8ko.

clean all unused thumb every 30 days.

- [x] Linux (arch)
- [x] mac os
- [x] windows


packages/app/app/components/LibraryView/LibraryAlbumGrid/index.js
```diff
@@ -23,7 +23,7 @@ export default compose(
           .thru(_.head)
           .value(),
         thumb: _(group)
-          .map('image[0][\'#text\']')
```
nukeop: Don't change this, I'm unifying and overhauling search and this will be replaced anyway.

packages/main/src/services/local-library/interfaces.ts
```diff
@@ -0,0 +1,22 @@
+import { NuclearStream } from '@nuclear/core';
+import LocalFolder from './model/LocalFolder';
+
+export interface NuclearLocalMeta {
```
nukeop: Can you try to make it compatible with the types I'm working on for the search plugins?

i. e. https://github.com/nukeop/nuclear/blob/feature/search-providers/packages/core/src/plugins/plugins.types.ts

charjac: I see you just merge your branch :) :) :). I will refactor this stuff this week.

packages/app/app/components/LibraryView/LibraryAlbumGrid/index.js
charjac: ok

packages/main/src/services/local-library/interfaces.ts
nukeop: Nice, i tried to move a lot of functionality to typescript. Some tests are broken because of this, not sure why. We're at 17% TS now

charjac: I am still a little bit confuse with the model.
for example in the queue the thumbnail has to be named thumb, but in the library view it has to be thumbnail.
artist is sometimes a string sometimes an object.

packages/app/app/components/LibraryView/LibraryAlbumGrid/index.js
charjac: actually this component will only be used with local files, its way more easier to manage a thumbnail property than an array with a single object with a #text property.

packages/main/src/services/local-library/interfaces.ts
nukeop: This is a consequence of the different apis returning different things. I was trying to standardize what they return while retaining as much information as possible, and preserving naming etc. If you have a better idea how to organize these structures, by all means improve them.

charjac: Ok, i will keep it simple for now.

What do you think of improving a bit the Track struct. We could use getter to access thumbnail or artist for example. (in another PR)

charjac:
Should i fix the lint errors ?

nukeop:
If it's not too much work. I don't know why they pass on Linux but not on Windows.

charjac:
I remove the lint fixes, they should be done in another PR.

nukeop:
Ok, ready for review/merge?

charjac:
We should probably test it on mac os and windows before merge

charjac:
It's ok on windows

charjac:
I finally achieve to test it on macos catalina and it works just fine.
Its ok for review.

charjac:
I will rebase this when #710 is merged.

nukeop:
Great news, I will merge it today or tomorrow.

charjac:
its rebased and still working.
The new local library ui is awesome !!!

I only tested it with 50 albums.
We should test it with huge mp3 library, but i don't have this and too lazy to download the whole piratebay :).
If someone could help ?

PR #656, opened 2020-02-19
Change Discord Rich Presence Details Text

nathanwfranke:
Before, the details text said "listening to" which took up half the space, so you would get part of the artist's name and the rest would be cut off.

**New behaviour**:
image

Also, added a `-` to separate the artist and title


nukeop:
Nice, thank you for fixing this.

PR #664, opened 2020-03-05
Feature/queue item context menu

### LINKED ISSUE #612: Change queue popup into a context menu
Queue popups currently display stream info and allow rerolling streams.

I would like to change these into a regular context menu (similar visually to the one at the top of the play queue) that lets users interact with queue items more freely and download them, add them to favorites, playlists, etc.

mr00z:
This resolves #612 

Had to change positioning of the context menu - it's placed under or at the top of the given queue item. This is because after addition of the action buttons popup became a lot higher and was overflowing the app's window when triggered by the first or last item in the queue. 
Also changed thumbnail's size in the `StreamInfo` to fixed values for the Popper.js to work properly - it was calculating whether to flip popup or not when the image has not been loaded yet which was leading to strange behaviors.


nukeop:
Great, thank you for this.
Do you have any screenshots of your work?

mr00z:
Sure.

nukeop:
Oh, this is awesome. Maybe we could move the popup to the left of the queue, so it never obscures the other queue items?

mr00z:
Yeah, I tried that in the first place, but the popup is a bit taller now and due to that when triggered by the first or the last item in the queue its pointing arrow is misplaced. See pictures (ignore color of the arrow):
left
left2
The setting for its placement is being set in `:before` pseudo-element in Semantic's CSS which I am not able to change on React's side. I was thinking about using `calc()` function in CSS, but haven't come up with any solution. Maybe you have any idea on how to do it?

nukeop:
You should be able to move the arrow using css and top/bottom properties. The popup is positioned using the target prop, you could use something similar.

The alternative is to turn the popup into a "pusher" tab sliding out from the right, from under the queue panel, dimming the rest of the window. Here's an example of what I'm talking about: https://react.semantic-ui.com/modules/sidebar/#states-dimmed

Maybe this will be easier/more fun to implement for you. Of course if you don't feel like doing this, we could merge what you wrote right away and I'll handle further adjustments myself.

mr00z:
Ok, I’ll try some more and see if I can do something about the arrow’s placement and get back to you.

mr00z:
Alright, I think there is no clear way to deal with the arrow's position as of now. It seems that [there is a bug](https://github.com/Semantic-Org/Semantic-UI-React/issues/3618) in Semantic UI React library regarding Popper.js. I tried to manually assign `top` value to the arrow in case when the popup is too close to the top or bottom, but it's still messing up in case of compact queue items. 

You can merge my changes if you want to, because I don't see any valid solution to positioning the popup on the left of the queue with those additional buttons.

nukeop:
Thanks for looking into this. I'll merge your changes, and later I will incorporate what you wrote into a sliding panel.

Issue #659, opened 2020-02-23
Allow using back/forward media keys to navigate


nukeop:
We have navigation buttons, would be good if using the keys that some keyboards and mice have also worked.

JohnRawlins:
Can I try working on this feature?

nukeop:
Closed by #665, thank you @JohnRawlins

nukeop:
Not closed yet actually, we still need support for navigating pages back and forward.

JohnRawlins:
![nuclear](https://user-images.githubusercontent.com/42707277/76172417-fcdda300-616b-11ea-8234-f2ee653c1bc7.png)

By navigating pages, do you mean using the forward and back media keys as shortcuts for these buttons?

nukeop:
Yep, exactly!

JohnRawlins:
@nukeop I was thinking that we can use the [alt] + [left arrow] and [alt] + [right arrow] keys, instead of the media keys to control the navigation between pages. This is similar to the shortcuts used in Chrome. What do you think about this solution?

nukeop:
This needs some rethinking, so thanks for raising this. I realized there are no dedicated back/next keys or buttons, they're actually for controlling music players. So the plan would be:

- Alt + arrows like you suggested for previous/forward navigation. This can be defined with Mouse trap here: https://github.com/nukeop/nuclear/blob/master/packages/app/app/containers/ShortcutsContainer/index.js
- Media keys "next track" and "previous track" for switching to the next/previous track in the queue. This can be implemented the way you did this previously. Additionally, we could try using the globalShortcut Electron API to make the program respond to those globally: https://www.electronjs.org/docs/api/global-shortcut

alizeh12:
Hi, I would like to work on this if the issue hasn't been resolved already @nukeop

nukeop:
Sorry, it's been already implemented, I just forgot to close the issue.

allanlaal:
mouse back/forward keys do nothing in:

          ID: org.js.nuclear.Nuclear
         Ref: app/org.js.nuclear.Nuclear/x86_64/stable
        Arch: x86_64
      Branch: stable
     Version: 2dd909
     License: GPL-3.0+
      Origin: flathub
  Collection: org.flathub.Stable
Installation: system
   Installed: 319,5 MB
     Runtime: org.freedesktop.Platform/x86_64/20.08
         Sdk: org.freedesktop.Sdk/x86_64/20.08

[...]

PR #665, opened 2020-03-05
Added the ability to use media control keys with music playback

JohnRawlins:
Please review this PR relating to issue #659. Thank you.


nukeop:
Not a bad place to add this. I was thinking if maybe we should move this to ShortcutsContainer, but that one handles mainly mousestrap bindings, so I think this can stay here.

Thank you for contributing, very good work.

PR #669, opened 2020-03-11
Added the ability to use keyboard shortcuts for navigating between pages

JohnRawlins:
Relating to issue #659

I decided to use [alt], along with the [left] and [right] keys to navigate between pages. Is this solution a good alternative to the media keys? If not, let me know and I will make the necessary changes. Thank you.


nukeop:
I replied in the issue, and then noticed that you already implemented this. Great job, this is exactly how I imagined it.

PR #674, opened 2020-03-17
Fix local tracks in playlists

### LINKED ISSUE #655: Local tracks in playlist have no playback source
- **Nuclear version**: 0.6.3 commit 5c4bd7020ba5a170d55c69941a00d8fdf8586362
- **NodeJS version**: 13.9.0
---
When trying to play a local track from a playlist Nuclear does not seem to pass on the fact that it is supposed to be a local track. The track therefore loses all playback sources and can not be played. 

Adding a simple 
```js
console.log(item);
```
before the return statement here https://github.com/nukeop/nuclear/blob/5c4bd7020ba5a170d55c69941a00d8fdf8586362/packages/app/app/actions/queue.js#L19-L20
allows printing out the data Nuclear is sending. Below are two examples of these.

This is when the track is played from the local library: 

This is when the track is played from a playlist after being added to it via the queue after playing it from the local library: 

As you can see the track played from the playlist loses the `local` and `streams` entries while the `loading` state is set to true instead.

Curiously using the button to add all tracks of a playlist to the queue works as expected. Additionally the playlist stored in `config.json` does contain the 
```js
local: true
```
entry but it seems to get ignored.

I believe this might be due to the different versions of the TrackRow component being used.

Mongogamer:
These commits fix #655.
Playlists will now pass on the `streams` and `local` key of tracks so local tracks can be played again. I don't believe passing either of those keys on will cause any issues with remote tracks. I have tested with multiple songs from Invidious and a really old data sample from YouTube and haven't encountered any issues so far.

### DIFF
--- a/packages/app/app/components/TrackRow/index.js
+++ b/packages/app/app/components/TrackRow/index.js
@@ -40,7 +40,9 @@ class TrackRow extends React.Component {
     this.props.actions.playTrack(this.props.streamProviders, {
       artist: this.props.track.artist.name,
       name: this.props.track.name,
-      thumbnail: this.getTrackThumbnail()
+      thumbnail: this.getTrackThumbnail(),
+	  local: this.props.track.local,
+	  streams: this.props.track.streams
     });
   }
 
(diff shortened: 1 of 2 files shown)


nukeop:
Old youtube streams will not play. The same is true for stream links generated on a different computer. Now that we use invidious, I think it's safe to use though.

PR #678, opened 2020-03-28
Search providers

nukeop:
This creates a dropdown in the search box that lets you choose between different search providers (Discogs and Musicbrainz for now, Bandcamp later).

Also includes a ton of all around improvements, and a lot of stuff has been rewritten in Typescript. It's now possible to write UI components in Typescript too.


nukeop:
Before merging, I need to improve the new searchbox (keyboard operation), and fix the build.

Issue #645, opened 2020-01-30
Show last few searches in a dropdown


nukeop:
Something like Stremio:

![ArchLabs-43-393x321](https://user-images.githubusercontent.com/12746779/73490994-8ea40480-43a5-11ea-8086-e3a67cb126f9.png)

In the future we could add more stuff to it, so it should be kept as generic as possible.

grimkor:
Morning @nukeop,

mind if I take a pop at this?

Thorbecke:
unsubsribe

Thorbecke:
Please, unsubscribe, Thanks!

nukeop:
Sure @grimkor !

Thorbecke, if you want to unsubscribe, you'll need to stop watching the repo, or click the unsubscribe button in this issue (not sure how you subscribed).

grimkor:
@nukeop I've got the functionality of the dropdown complete but with #630 pending nothing is displayed in the dropdown. Not sure if you want the functionality now and the content later once other PRs are ready to give it life.

https://github.com/grimkor/nuclear/commit/d8969fe223997cca49a71851ec9866434ab326f1

nukeop:
Great, it's fine to open a PR now, I'll take a look as soon as possible. Thank you!

nukeop:
Done

PR #680, opened 2020-04-01
feat: Search dropdown

grimkor:
Related to #645 

Been a while since I worked on Redux so if there's any feedback or reworking required I'm all ears.


grimkor:
I'll take a look at this later this week when I have time again.

PR #686, opened 2020-04-09
Fix build script not running on Windows

### LINKED ISSUE #683: "build" script cannot run on Windows
The root `build` script fails on Windows, due to it using the `rm` command (which Windows does not have).

Same thing for the `start` script in `packages/main/package.json`, called by the root `start` script (which I've been trying to run).

Solution: add shx as a dev-dependency, and have the script call `shx rm` instead: https://github.com/shelljs/shx

I've used `shx` in multiple projects, and it's worked well for providing cross-platform equivalents of the linux `rm`, `cp`, `mkdir`, etc. commands (list of commands [here](https://github.com/shelljs/shelljs#command-reference)).

EDIT: I can confirm that after inserting `shx ` before the two `rm` calls mentioned above, running `npm run start` works as expected now on Windows. (launches program without errors)

Venryx:
* Fixed build script not running on Windows, by using the cross-platform "shx" equivalent. (resolves #683)

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
   "scripts": {
     "postinstall": "lerna bootstrap --no-ci",
     "start": "lerna run start --stream",
-    "build": "rm -rf dist && lerna run build && npm run pack",
+    "build": "shx rm -rf dist && lerna run build && npm run pack",
     "test": "lerna run build && lerna run test",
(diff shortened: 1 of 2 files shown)


Venryx:
Oops, wrong branch. Will close this and recreate.

PR #689, opened 2020-04-14
Uncomment app.quit()

tghgg:
Should allow the app the terminate properly now instead of lingering in the background.


nukeop:
Thank you!

PR #690, opened 2020-04-15
645 searchdropdown

grimkor:
This is for #645.

Unfortunately I haven't had much time to work on this so for now I have tied it off with a working dropdown that will display when a user clicks inside the search box and closes when either ESC is pressed or the user clicks away into the main window. 

Currently since there is no content inside the search dropdown nothing is displayed but the functionality is there.

Hopefully the structure is correct for the project; I'm unfamiliar with the setup so let me know if there's anything you want moved around or done in another way and any feedback is appreciated.


packages/app/app/actions/search.js
```diff
@@ -298,3 +300,25 @@ export const albumInfoSearchByName = (albumName, history) => async (dispatch, ge
     ));
   }
 };
+
```
nukeop, line 303: I have a suggestion about these actions that will make these more concise:

```javascript
export const setSearchDropdownVisibility =  displayState => ({
  type: EARCH_DROPDOWN_DISPLAY_CHANGE,
  payload: displayState
})


```

No need to use a thunk (the async function with dispatch) since it's not a complex action.

nukeop:
Thanks, I tested it and it works ok, it leaves us in a nice place to start work on search history. Thank you for contributing.

Issue #660, opened 2020-02-24
Discord Rich Presence counts elapsed even if song is paused


nathanfranke:
You can't actually pause the start time in the Discord API so instead it should just say paused. When you play it the start time is set to the current time minus the position so it shows at the right spot

nukeop:
Thanks for catching this, needs to get fixed.

nathanwfranke:
I am not too familiar with how this project works but this is the code that should work
```js
let activity = {
  details: `${track.artist} - ${track.name}`,
  largeImageKey: 'logo'
}
if (paused) {
  activity["details"] += "\nPaused";
} else {
  activity["startTimestamp"] = Date.now() - elapsed;
}
await this.rpc.setActivity(activity);
```

advaithm:
Fast forwarding or rewinding need to be included to.

charjac:
Could i close this ?

PR #694, opened 2020-04-18
feat(discord): better play/pause support

charjac:
related to #660 

better elapsed time in discord.
Its probably too hacky, i did not find any better way to do it, i really dont like this interval.
What do you think ?

### DIFF
--- a/packages/main/src/services/discord/index.ts
+++ b/packages/main/src/services/discord/index.ts
@@ -26,14 +26,14 @@ class Discord {
   private async sendActivity() {
     try {
       await this.rpc.setActivity(this.activity);
-      this.logger.log('discord > update activity');
+      this.logger.log('update discord activity');
     } catch (err) {
       this.logger.error('error trying to set discord activity');
     }
   }
 
   async pause() {
-    if (this.isReady) {
+    if (this.isReady && this.activity) {
       this.pauseStart = Date.now();
   
       this.activity.details += '\nPaused';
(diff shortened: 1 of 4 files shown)


nukeop:
Would it be possible to create a redux subscriber that listens to play/pause/fast forward/rewind events, and tells the Discord part what to do? Or does Discord require refreshing the state every few seconds?

charjac:
discord control the timer, the only thing we can do is refresh the start timestamp.
We could refresh it when paused, store the paused elapsed time and then refresh it with some basic math when played again.

charjac:
The discord activity is refreshed:
 - when playing track change
 - when paused
 - when play again
 - when track-list is cleared
 - when a track is removed from a track-list of 1 track

charjac:
I remove the interval thing, i think its not perfect but better like this

charjac:
> Would it be possible to create a redux subscriber that listens to play/pause/fast forward/rewind events, and tells the Discord part what to do? Or does Discord require refreshing the state every few seconds?

We should definitely do this for discord and mpris event to avoid using ipcRenderer inside redux actions. 
Will try something in another PR.

nukeop:
Great, this is very nice now, thanks for improving this.

packages/main/src/services/discord/index.ts
```diff
@@ -10,29 +9,56 @@ import Logger, { $mainLogger } from '../logger';
...
+      this.activity.startTimestamp = this.pauseStart;
+      return this.sendActivity();
+    }
+  }
+
+  async play() {
+    if (this.isReady && this.activity) {
+      this.pausedTotal += Date.now() - this.pauseStart;
+      if (this.activity) {
+        this.activity.details = this.activity.details.substr(0, this.activity.details.length - 8);
```
nathanfranke, line 49:
```suggestion
        this.activity.details = this.activity.details.substr(0, this.activity.details.length - 7);
```
This is badly programmed. Would be fixed after https://github.com/nukeop/nuclear/pull/761 which doesn't hard code the length of `\nPaused`

Issue #696, opened 2020-04-19
Update DevBuildProcess in Contributing Wiki Page


primeapple:
Hi everyone,

I suggest to update the "Contributing" WikiPage ( see https://github.com/nukeop/nuclear/wiki/Contributing ). The last update was over 1 year ago and since this you seem to have moved more towards docker! Please update and also include the docker instructions from https://github.com/nukeop/nuclear/blob/master/README.md

Also I found, that the building is not working correctly? With docker I get the error:
```
dev_1      | lerna ERR! Error: ENOENT: no such file or directory, rename '/nuclear/packages/main/package.json.lerna_backup' -> '/nuclear/packages/main/package.json'
dev_1      | lerna ERR!     at Object.renameSync (fs.js:754:3)
dev_1      | lerna ERR!     at cleanup (/usr/local/lib/node_modules/lerna/node_modules/@lerna/npm-install/npm-install.js:70:10)
dev_1      | lerna ERR!     at done (/usr/local/lib/node_modules/lerna/node_modules/@lerna/npm-install/npm-install.js:78:7)
dev_1      | lerna ERR!     at /usr/local/lib/node_modules/lerna/node_modules/@lerna/npm-install/npm-install.js:94:19
dev_1      | lerna ERR! lerna ENOENT: no such file or directory, rename '/nuclear/packages/main/package.json.lerna_backup' -> '/nuclear/packages/main/package.json'
```
However, after this, the starting continues until 
```
dev_1      | + lerna run start
dev_1      | lerna notice cli v3.20.2
dev_1      | lerna success run No packages found with the lifecycle script 'start'
```

With the manual method by using lerna, I can build nuclear successfull, but when starting with `npm start`, I also get: 
```
lerna success run No packages found with the lifecycle script 'start'
```

However, `npm run build` gives me 
```
sh: shx: Kommando nicht gefunden.
npm ERR! code ELIFECYCLE
npm ERR! syscall spawn
npm ERR! file sh
npm ERR! errno ENOENT
npm ERR! nuclear@0.6.3 build: `shx rm -rf dist && lerna run build && npm run pack`
npm ERR! spawn ENOENT
npm ERR! 
npm ERR! Failed at the nuclear@0.6.3 build script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
npm WARN Local package.json exists, but node_modules missing, did you mean to install?
```

Do you have any ideas?

Venryx:
[...]

PR #698, opened 2020-04-20
Fix development process instructions in readme (#696)

primeapple:
This is a pull request to improve the documentation (see issue #696 )

I think this would make it easier to develop nuclear for new users. I also plan on improving the wikipage.

Another thing I would like to change would be the docker config. However, I guess a good way to do this is in a new issue.

Please tell me if you would like to change something.


nukeop:
Thanks!

Issue #679, opened 2020-03-29
Volume not saved after restart


nathanwfranke:
It always changes to max volume after restarting

nukeop:
Right, thank you for reporting. This needs to be saved in the store.

dejesusj:
If no one's working on this, I'd also like to give this a try.

nukeop:
Thanks. Settings are defined in this file: https://github.com/nukeop/nuclear/blob/master/packages/core/src/settings/index.ts

So you'd need to add a new option somewhere in there, then make sure this action saves the current volume in it: 

https://github.com/nukeop/nuclear/blob/20772c637d70e28bd1a27ebe6099f6931e7968ba/packages/app/app/actions/player.js#L64

There are also a few places that read volume; those will need to use the new setting. Let me know if you have any problems.

zephy20:
Is this still on pending? If yes, I'd like to give a try to this.

nathanwfranke:
Go ahead since it's been 11 days since someone else wanted to fix it

nukeop:
Yep, go ahead. What I wrote earlier still applies.

dejesusj:
Could someone point me to the places that use Volume, specifically upon restart?

nukeop:
Currently volume is initialized by redux, using an initial state:
https://github.com/nukeop/nuclear/blob/master/packages/app/app/reducers/player.js#L25

PR #700, opened 2020-04-21
Saves volume upon restart

dejesusj:
This is a pull request that would resolve https://github.com/nukeop/nuclear/issues/679
I’ve made the following changes https://github.com/nukeop/nuclear/pull/700/files:
- Defined a new setting for volume
- Updates the setting every time volume is updated
- Sets initial volume state to previously saved volume


dejesusj:
Would you know how to fix the issue causing this check to fail by any chance? I noticed a similar issue in a previous pull request.

nukeop:
Yes, this is a problem with some tests after a refactor, don't worry about it. Thanks for contributing.

Issue #677, opened 2020-03-23
Playback stops if shuffle randomly reaches the last song


nathanfranke:
If I queue 10 songs (named 1-10 for example)
shuffle might go from song 5, to 3, to 6, to 10 but once it reaches 10 and 10 finishes playing, the song will stop playing completely

nukeop:
Right, the queue end check should also check if this setting is enabled.
This will also cause issues with single track looping: https://github.com/nukeop/nuclear/issues/475

dejesusj:
I would like to work on this!

nukeop:
Great, thank you. The best place to start would be here: https://github.com/nukeop/nuclear/blob/master/packages/app/app/containers/SoundContainer/index.js#L69

This functions controls what happens after a song has finished playing.

dejesusj:
@nukeop Apologies for the delayed response, I was unexpectedly busy recently. Based on my understanding of how queues work on most music streaming platforms, I have a number of questions.
- Once a song is played, should it be removed from the queue?
    - If so, then could we just remove each played song from queueItems, and keep playing from the queue until queueItems is empty?
- If songs should not be removed from the queue after they are played, then how should the queue function? 
    - Currently it seems to function as a temporary playlist of some sort.
        - If that is the case, then should all songs be played at least once, and then once all songs have been played, either stop playback or continue looping the queue based on settings?
            - If Loop is toggled, then keep playing from the queue
            - If Loop is not toggled, and Autoradio is not toggled, then stop playback

nukeop:
No, songs should not be removed from the queue automatically. The only change for this issue should be that we continue to the next random track regardless of which one we're currently at. Autoradio should not add new tracks, and the player should continue selecting new tracks randomly until user action.

nukeop:
Fixed, thank you @dejesusj !

Issue #475, opened 2019-09-20
[Feature request] Single track looping


farahnur42:
As it stands right now the loop button isn't looping the track(s) and is stopping after a single play each. Not sure if this is a bug or intended behaviour of the looping functionality but it'd help a lot for someone like me who tends to keep songs on repeat for extended periods of time.

mahimshridhar:
Can i take this ?

nukeop:
Sure! I haven't started working on it yet. If you have any questions, feel free to ask me here.

mahimshridhar:
Any suggestions on how do I solve this?

nukeop:
What happens after a song is finished playing is controlled here:
https://github.com/nukeop/nuclear/blob/b9feeb853b5c5644a36442e33d21fca3075a3d01/packages/app/app/containers/SoundContainer/index.js#L62

There is already a check that handles looping the queue after we're done playing the last song. You could look at this code, introduce a new setting, and reset the current song based on that.

mahimshridhar:
nuclear_install_1 is up-to-date
Starting nuclear_dev_1 ... error

ERROR: for nuclear_dev_1  Cannot start service dev: error gathering device information while adding custom device "/dev/snd": no such file or directory

ERROR: for dev  Cannot start service dev: error gathering device information while adding custom device "/dev/snd": no such file or directory
ERROR: Encountered errors while bringing up the project.

I got this error while running docker-compose up dev and im not able to resolve this, can you please help me?

nukeop:
What system is this? And what sound api?

mahimshridhar:
im on macOS 10.14.5

nukeop:
Ok, as a temporary workaround you can try using the old method of running the dev environment:

```shell
$ npx i -g lerna
$ lerna bootstrap
$ lerna run start
```

mahimshridhar:
Im sorry, should i modify package.json for this to work?  "lerna run start" wont work, it gives me: 
lerna notice cli v3.16.4
lerna success run No packages found with the lifecycle script 'start'

nukeop:
In this case, you can cd to `packages/app` and run `npm start`. I have no idea why lerna would fail to see the script in package.json.

mahimshridhar:
ok now it is working but some reason it is stuck at loading,
[...]

PR #701, opened 2020-04-21
Fixed conditional to handle after song has been played

dejesusj:
I've changed the second conditional in handleFinishedPlaying to continue to select songs as long as the queue is not empty, rather than until the last song in the queue is selected. This fixes the issue mentioned in https://github.com/nukeop/nuclear/issues/677, however if shuffle is toggled and the song that is selected randomly is the one that just played, playback is paused. I believe this is related to the issue mentioned in https://github.com/nukeop/nuclear/issues/475.

### DIFF
--- a/packages/app/app/containers/SoundContainer/index.js
+++ b/packages/app/app/containers/SoundContainer/index.js
@@ -82,7 +82,7 @@ class SoundContainer extends React.Component {
     }
 
     if (
-      this.props.queue.currentSong < this.props.queue.queueItems.length - 1 ||
+      this.props.queue.queueItems.length !== 0 ||
       this.props.settings.loopAfterQueueEnd
     ) {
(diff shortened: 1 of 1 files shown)


nukeop:
This will cause the queue to keep playing even if we're not shuffling, which should be enabled only if we use the setting `loopAfterQueueEnd`.

dejesusj:
> This will cause the queue to keep playing even if we're not shuffling, which should be enabled only if we use the setting `loopAfterQueueEnd`.

Ah, I see. So if we're shuffling, then we want to keep picking songs at random from the queue indefinitely? And if we are not shuffling, then we want to pause playback once the last song in the queue is played?

nukeop:
Yes, exactly.

nukeop:
Thank you, works well now!

PR #702, opened 2020-04-22
Feat/ipc redux

charjac:
add a ipc middleware to redux store, its role is to send ipc events. This way, redux actions are independant of electron (we should probably do something similar for electron-store).

This PR also optimize ipc exchange for volume and playpause events.
before: 
mpris => main => renderer => main
now:
mpris => main => renderer

I also remove mpris.js in core package.
all ipcRenderer related code is now located in ipcContainer and ipcMiddleware.

All events name are defined in an enum in core package.


nukeop:
Wow what a fantastic contribution. It was a long time coming. It's so neat and ordered now it makes it way easier to maintain and use.

Issue #699, opened 2020-04-21
Docker-Compose rework


primeapple:
Hi again.

I would like to talk about the docker-compose config a bit. The reason is, that we then could move directly to docker and won't even need npm installed on the local machine anymore!

There are several things that seem a bit confusing to me.

1. In the docker-compose.yml you have to services. However, the `install` service seems to be not used? Can it be removed?
2. Why do you have to run `lerna exec npm rebuild node-sass`? Why is sass needing to be rebuild?
3. Why do you run `lerna run start`? Wouldn't it be better to run `lerna run build` and then execute the built app?

Overall I would like to restructure the docker setup like this: https://dev.to/alex_barashkov/using-docker-for-nodejs-in-development-and-production-3cgp

What are your thoughts about this?

EDIT: Also I am curious about the [this Dockerfile](https://github.com/nukeop/nuclear/blob/master/Dockerfile). Is it still used anymore? What is with the docker scripts in the [package.json](https://github.com/nukeop/nuclear/blob/master/package.json)? Are they used anymore?

nukeop:
`install` is used in the `dev` service (depends_on list).
`node-sass` uses native modules that need to be rebuilt in case of different node versions, if they're compatible it won't rebuild them.
`lerna run start` starts the program in dev mode, which is the purpose of this config. `lerna run build` would build the program in production mode.

That article you've linked is about pure node apps, most of it won't apply to this use case, for a number of reasons (e.g. Nuclear has a different build process than what's described there, because it has multiple packages that need to be built, in both dev and prod environments, so they all need dev dependencies, we also don't want to copy the code inside the container).

The dockerfile and package.json script at the bottom of your post exists because somebody wanted a way to build and run the program inside a container, which they contributed.

primeapple:
Thank you so much for your answers!

Since `install` uses the same Docker file as `dev` it does exactly the same as dev. So the `install` container is build and also up, but is not used at all from my understanding.

The thing with the current docker container is, that we actually DO copy everything in the repo into the docker-container (see last line in `Dockerfile_dev`). That is not how to use docker, you would want to exclude any unused stuff (node_modules mainly) and built everything from scratch in Docker. This way it makes sure that the project works the same way on any computer, no matter which node_modules were installed before.

In the end this should work on any pc without trouble:
```sh
# clone new instance of nuclear somewhere
git clone https://github.com/nukeop/nuclear.git
# start docker-compose, rebuild it if you
[...]

PR #705, opened 2020-04-26
reworked docker dev, added docker-compose builder

primeapple:
A first idea on #699 
@nukeop , please check if you like the changes. Also please check if you are able to bring docker up. I have some troubles with XHOST and get stuck when trying to launch the app with `docker-compose up`.

The instructions how to run the docker are in the README changes.

Inspiration is from https://hackernoon.com/a-better-way-to-develop-node-js-with-docker-cd29d3a0093

### DIFF
--- a/docker-compose.builder.yml
+++ b/docker-compose.builder.yml
@@ -1,0 +1,12 @@
+# inspired from https://hackernoon.com/a-better-way-to-develop-node-js-with-docker-cd29d3a0093
+version: '3.7'
+
+services:
+  install:
+    image: node:14
+    volumes:
+      - .:/nuclear/
+    working_dir: /nuclear/
+    command: bash -xc "
+        npm install;
+        npx lerna bootstrap"
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -1,8 +1,3 @@
-FROM node
+FROM node:14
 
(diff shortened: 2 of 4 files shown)


docker-compose.builder.yml
```diff
@@ -0,0 +1,12 @@
...
+
+services:
+  install:
+    image: node:14
+    volumes:
+      - .:/nuclear/
+    working_dir: /nuclear/
+    command: bash -xc "
+        npm install;
+        npx lerna bootstrap"
```
nukeop, line 12: `lerna bootstrap` will run as a postinstall step, so all we need is `npm install`.

Btw, I read that article some time ago, and again now, and I still don't see what we gain by splitting this part into a separate file, as opposed to a service in the same docker-compose file as the rest.

nukeop:
Hey, I finally had some time to take a look at this. I made one comment; otherwise it's fine.

primeapple:
Hey, thank you so much! :-)

After working and trying on docker some more, I am not so happy for this... Docker seems awesome for production, but for development... Well, I am not so sure about this.
I will look into it a bit more, but maybe it is the best, to just leave it as it is in master currently and flag it as "experimental" in the readme?

nukeop:
I'm fine with either, I guess the default build process should be `npm install` then `npm start` and we can leave docker out of it.

PR #706, opened 2020-04-26
fix(plugins): fix invidious plugins

charjac:
Looks like the first result of some search (a lot) is fucked up ...


nukeop:
Maybe this is because the first search result tends to be an 'official' video, which are often geoblocked? I didn't notice that many streams failing.

charjac:
I have a lot of problem with the actual top album "fiona apple". (not all tracks)
Did you experience same problem ?

nukeop:
Yeah, looks like I have the same problem with this album.

PR #707, opened 2020-04-26
feat(devtools): add shortcut and settings for devtools

### LINKED ISSUE #684: [feature request] Add button to open Developer Tools in production builds
Related to: https://github.com/nukeop/nuclear/issues/418

I'm starting work on the plugin I wanted to make months ago (an advanced playlist-construction/shuffling system, with weightings per folder to control the relative distributions).

However, it's hard to debug, since there doesn't seem to be a way to open the Developer Tools panel from the UI, in the production version of Nuclear. (I'm testing the plugin there since it starts up a lot quicker for everyday use than the `npm run start`, etc. commands)

I did find a work-around for now, adding this to the onLoad function of my plugin:
```
onLoad: api => {
...
	const listener = (event, webContents)=> {
		if (!webContents.isDevToolsOpened()) {
			webContents.openDevTools({mode: "bottom"});
		}
		api.app.removeListener("browser-window-blur", listener);
		api.app.removeListener("browser-window-focus", listener);
	};
	api.app.on("browser-window-blur", listener);
	api.app.on("browser-window-focus", listener);
	api.app.focus();
...
}
```

And then clicking outside of the window to trigger the "browser-window-blur" event. (this is required to get access to the [webContents](https://www.electronjs.org/docs/api/web-contents) instance, since only the `app` part of the electron api is currently exposed)

However, it would be nice to have a button built in to the UI for this -- for example, this would let the user copy-paste in code snippets to perform some basic tasks, without having to set up a plugin folder.

**EDIT:** Found a better way to open the dev-tools programmatically:
```
const webContents = require("electron").remote.BrowserWindow.getFocusedWindow().webContents;
if (!webContents.isDevToolsOpened()) {
	webContents.openDevTools({mode: "bottom"});
}
```

Would still be nice to have a "Dev-tools" button somewhere, though. (perhaps to the right of the "Add a plugin" button on the Plugins page)

charjac:
fix #684

### DIFF
--- a/packages/app/app/containers/ShortcutsContainer/index.js
+++ b/packages/app/app/containers/ShortcutsContainer/index.js
@@ -146,24 +146,22 @@ class Shortcuts extends React.Component {
       'ctrl+top',
       'command+top',
       'ctrl+down',
-      'command+down'
+      'command+down',
+      'f12',
+      'command+i'
     ]);
   }
 
-  componentWillUpdate() {
-
+  componentDidUpdate() {
+    if (this.props.settings.devtools) {
+      Mousetrap.bind(['f12', 'command+i'], this.props.actions.openDevtools);
+    } else {
+      Mousetrap.unbind(['f12', 'command+i']);
+    }
   }
 
   shouldComponentUpdate(nextProps) {
-    if (nextProps.settings.devmode !== this.props.settings.devmode) {
-      if (nextProps.settings.devmode) {
-        Mousetrap.bind(['f12', 'command+i'], this.props.actions.openDevtools);
-      } else {
-        Mousetrap.unbind(['f12', 'command+i']);
-      }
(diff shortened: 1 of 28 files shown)


packages/app/app/containers/ShortcutsContainer/index.js
```diff
@@ -149,7 +150,19 @@ class Shortcuts extends React.Component {
     ]);
   }
 
-  shouldComponentUpdate() {
+  componentWillUpdate() {
+
+  }
+
+  shouldComponentUpdate(nextProps) {
```
nukeop, line 157: Shouldn't this be in `componentDidUpdate`?

charjac: I changed it

PR #710, opened 2020-05-07
Adding tree view for local library (plus other fixes)

### LINKED ISSUE #568: Scrolling issue with new simple list in local library
- **Nuclear version**: 0.5.0 commit 331521a7ba41c898ced62e97365b60295a05cf6e
- **NodeJS version**: 12.12.0
---
**Description of the issue**: Since the new album layout has been added to the local library there have been some scrolling issues with the simple list. The scrollbar constantly jitters around and completely resets if dragged too low.
Here is a short video showcasing the issue:
[https://streamable.com/357vq](https://streamable.com/357vq)

### LINKED ISSUE #685: [feature request] Tree-view for local library
Most of the long-term music players I've used in the past have had a way to display one's local library within a tree view, matching the folder/file structure.

For example, in foobar2000:

It would be nice to have an option to display one's local library like this in Nuclear as well. By displaying the actual folder/file structure, it allows you to categorize and subcategorize by your own criteria (for example, the `@Games` folder above), without it all just getting "flattened" as in the regular track/album views. Those views are good for many people, but some prefer the full control of having the raw directory tree represented.

I'm willing to work on this (as I also need it for [my plugin](https://github.com/Venryx/nuclear-vplugin) anyway), but would like to start the discussion here first to know if/how @nukeop would like me to approach it.

Venryx:
Tree-view-related changes:
* Added folder-tree view to local library page. (resolves #685)
* Greatly improved ui responsiveness, when track was playing, and the folder-tree view was open. (by using react-hooks useMemo in LibraryView component to prevent the "tracks" prop from changing, which was causing a new treeConfig to get sent to react-redux-grid each tick, causing it to rerender its internals -- specifically, deep-comparing the data tree)
* Added new `local.expandedFolders` store field, which remembers the expanded rows of the local-library folder tree-view. (and gets persisted to disk on program close)
* As required for persistence of `local.expandedFolders`, added support for nested fields to syncStorage enhancer.

General changes: (mostly fixes)
* Fixed that local tracks would not play, due to incorrect stream selection.
* Fixed that local-library folder-paths were not getting normalized (eg. from `\\` to `/`, on Windows) on initial selection. (fixes inconsistency, since gets normalized after closing and relaunching -- and consistency needed by new track-population code)
* React-list (for simple-list view of local library) was correctly setting the `height` value for the table, but for some reason it was being ignored by the browser renderer with the default `display: block` style. Switched the display-type to `table` to fix this. (resolves #568)
* Fixed error when opening the "Add local folder" dialog, then pressing "Cancel".
* Renamed NuclearBrutMeta.pos to NuclearBrutMeta.position, to match with all other "position" fields in the project. Also fixes React warning in Album view of local library, which was looking for "position" field. (TypeScript integration throughout the codebase should really help with this sort of thing.)
* Fixed that AlbumPreview component was still having React warnings, due to `track.position` apparently having clashes / not always being unique. (now just uses index as key)
* Corrected ui/Card.PropTypes.image to match the actual data. (thus solving React warning)
* Fixed that FavoriteTracksView was using minHeight instead of height, causing "no favorite tracks" text to not be centered. (and for scrolling, improved styling by having scrolling occur in table specifically, not for entire container, as that would cause darker background to not scroll with)
* Fixed PropType definition for TrackRow.track.artist, in ui/TrackRow/index.js. (fixing warning)
* Fixed React warning in ui/TrackRow, about `withAddToDownloads` not being valid attribute for "tr" element.

Screenshot of new folder-tree view:


Venryx:
Hmm, the CI build is failing: https://travis-ci.org/github/nukeop/nuclear/jobs/684086426

However, on reading the logs, it appears its caused by this:
```
⨯ cannot resolve https://github.com/electron-userland/electron-builder-binaries/releases/download/nsis-resources-3.4.1/nsis-resources-3.4.1.7z: status code 429
```

I'm assuming it's just a network error? If so, how to restart the CI build process?

nukeop:
I restarted your build on Travis, let's see how it goes this time.

nukeop:
Nice, it passed.

packages/app/app/components/LibraryView/LibraryFolderTree/index.js
```diff
@@ -0,0 +1,166 @@
...
+          id: path,
+          path,
+          name: path.split('/').slice(-1)[0],
+          children: []
+        };
+  
+        if (localFolders.includes(path)) {
+          newEntry.parentId = -1;
+        } else if (path.includes('/')) {
+          /* if (!localFolders.find(folderPath => path.startsWith(folderPath))) {
```
nukeop, line 27: Please remove the commented out code

packages/app/app/components/LibraryView/LibraryFolderTree/index.js
```diff
@@ -0,0 +1,166 @@
...
+        const newEntry = {
+          id: path,
+          path,
+          name: path.split('/').slice(-1)[0],
+          children: []
+        };
+  
+        if (localFolders.includes(path)) {
+          newEntry.parentId = -1;
+        } else if (path.includes('/')) {
```
nukeop, line 26: I think that the best course of action for platform compatibility would be to use `path.sep` instead of `/`

packages/app/app/components/LibraryView/LibraryFolderTree/index.js
```diff
@@ -0,0 +1,166 @@
...
+  );
+};
+
+LibraryFolderTree.propTypes = {
+  tracks: PropTypes.array,
+  localFolders: PropTypes.arrayOf(PropTypes.string),
+  expandedFolders: PropTypes.arrayOf(PropTypes.string)
+};
+
+export default compose(
```
nukeop, line 164: Let's forget about hocs in this component and use `useTranslation` instead, compose is also unneeded.

packages/app/app/components/LibraryView/LibraryFolderTree/styles.scss
```diff
@@ -0,0 +1,8 @@
+$header-background-color: transparent;
+$background-color: transparent;
+//$table-prefix: BaseTable !default;
+
+$border: none;
+$row-hovered-background-color: hsla(0, 0%, 100%, .2);
+
+@import '~react-base-table/es/_BaseTable.scss';
```
nukeop, line 8: Can this be moved to the top of the file?

packages/app/app/components/LibraryView/LibrarySimpleList/index.js
```diff
@@ -23,6 +23,12 @@ const LibrarySimpleList = ({
   t
 }) => (
   <ReactList
+    ref={comp => {
```
nukeop, line 26: Does this have to use a ref? Why not use css?

packages/app/app/components/LibraryView/index.js
```diff
@@ -1,35 +1,56 @@
-import React, { useCallback } from 'react';
-import PropTypes from 'prop-types';
+import React, { useCallback, useMemo } from 'react';
+import { useSelector, useDispatch } from 'react-redux';
```
nukeop, line 2: State subscriptions and action creator bindings should be kept in the corresponding container, and passed down as needed. Here they're split between the container and the component.

packages/app/app/store/enhancers/syncStorage.js
```diff
@@ -8,11 +8,16 @@ export default function(paths) {
       initialState = undefined;
     }
 
-    let persistedState, finalInitialState;
+    // retrieve initial states from reducers, to apply prior to merging electron-store persisted-data
```
nukeop, line 11: Why is this needed?

We already have a mechanism for applying default values to settings that weren't touched, unless this is needed for a different reason.

nukeop:
Thanks for fixing so many things at once, and for a detailed changelog. I made a few comments. Very solid work.

packages/app/app/components/LibraryView/LibraryFolderTree/index.js
Venryx: Will do.

packages/app/app/components/LibraryView/LibraryFolderTree/styles.scss
Venryx: I'm not sure of the details of how scss variable-insertion works, but I'll try it.

packages/app/app/components/LibraryView/LibrarySimpleList/index.js
Venryx: Good point, I'll try that. (I basically only use js-inline styles in my projects, so it didn't come to mind at first)

packages/app/app/components/LibraryView/index.js
Venryx: Right; the action creators and state subscriptions should be moved to the container for consistency.

I got a bit confused, subconsciously thinking the `useSelector` calls were needed in `LibraryView`, since I extracted them from the `useMemo` calls (which do need to be in `LibraryView`). I'll move them to the container.

packages/app/app/store/enhancers/syncStorage.js
Venryx: From what I remember, without the merging of `reducerInitialStates` beforehand, the supplying of data for `state.local` within the `persistedState` object, was causing the "default value" for the `state` parameter in the `local` reducer to be ignored. (since the reducer function's default-value is only used when `undefined` is passed for that `state` parameter)

I'll double-check this though.

packages/app/app/components/LibraryView/LibraryFolderTree/index.js
Venryx: Will do.

packages/app/app/components/LibraryView/LibraryFolderTree/index.js
Venryx: Will do.

packages/app/app/components/LibraryView/LibraryFolderTree/index.js
Venryx: Will do.

Venryx: Hmm, but making this change would require that the track metadata be created differently per platform. Isn't it best to standardize on "/" within the persistent metadata, rather than change it per platform? For example, using path.sep for the `track.path` property would mean you couldn't directly transfer the metadata between different-os computers.

packages/app/app/components/LibraryView/LibraryFolderTree/styles.scss
Venryx: Okay yeah, I tried placing the import at the top, and it makes it ignore the variable assignments. (the variables have to be set before the import, since the imported scss makes use of those variables, and the variable assignments and reads apparently "run" sequentially)

packages/app/app/components/LibraryView/LibrarySimpleList/index.js
Venryx: There doesn't seem to be any way to "identify" the particular element in question, other than by using the `ref` property. Here is the react-list source: https://github.com/coderiety/react-list/blob/master/react-list.es6

Note that there is no `className` or generic props passthrough for the root container. (which is what we need to style)

I could add the style under `.tracks_container div:nth-child(2)`, except then it would apply the styling to the album and folder-tree views as well.

Venryx: One "hack" is to add an extra container around the react-list component, with a unique class. Then I could add the styling for `.library_simple_list_container > div`.

Should I do that, or stick with the ref approach?

packages/app/app/store/enhancers/syncStorage.js
Venryx: Yep, the description above is correct; without that `reducerInitialState` merging, the default values for the fields in `state.local` (in `reducers/local.js`) are excluded on program start, due to there being data for that root `state.local` field from the `persistedState` source.

packages/app/app/components/LibraryView/LibraryFolderTree/index.js
Venryx: Fixed.

Venryx: Fixed.

packages/app/app/components/LibraryView/index.js
Venryx: Fixed.

Venryx:
Added most of the changes requested -- except for two which are awaiting comments/confirmation.

packages/app/app/components/LibraryView/LibraryFolderTree/index.js
nukeop: The persisted paths don't need to be uniform across platforms, since the directory structures won't match on different machines anyway (you won't be able to transfer the library data to a machine running a different OS). I think you're right though, let's stick with your initial idea, it's simpler.

packages/app/app/components/LibraryView/LibrarySimpleList/index.js
nukeop: Ok, that's too bad that the library doesn't expose a way to get to that container.

You could import the styles from the album view and tree view and use the `:not` selector but that's too much code for little benefit. Instead of directly setting the `display` property using the ref, could you please set a class on the ref and use that class in scss? This way if we want to apply more styling to that component in the future it will be easier.

Venryx: Done.

Venryx:
Okay, finished resolving the last change (regarding simple-list-container styling).

nukeop:
Excellent, thank you so much for contributing this.

PR #713, opened 2020-05-13
TS in app

nukeop:
I refactored the build process to enable TS in the app package (so we can use it everywhere now).

Core/ui are not built with tsc anymore, only main is (because it needs to be built separately). App imports core and ui directly, and then webpack transpiles it with `babel-loader` in dev mode (it's fast but it ignores all TS typings), and with `ts-loader` in prod mode (slower but checks typings). Tests etc work too.

The only remaining problem is that it times out on Windows without any output, and I can't figure out why. @charjac , have you seen something like this before and maybe you know off the top of your head what could be happening here?


charjac:
@nukeop no clue, i am not a big fan of windows ;)

nukeop:
I tried debugging this on windows and I can't get it to work no matter what. It builds correctly but lerna doesn't detect that webpack has finished, so it's frozen. I'm gonna merge it anyway, since it gives us huge benefits. I can always build windows version on my VM manually.

PR #727, opened 2020-06-25
Hidden options in settings menu

### LINKED ISSUE #725: Regression - Volume shows up in settings with no name or description
This shouldn't show up in the first place
I believe this is a regression from #700 @dejesusj 

image

nathanfranke:
Fix #725

Adds the ability to hide certain options from the settings menu. Used for the volume control setting.

:smiley:


nukeop:
Great solution, thanks for this.

Issue #722, opened 2020-06-04
[Feature Request] Organize/Sort songs in Favorite Tracks Page


skyawmyint:
**Location of Improvement**
In the favorite tracks page
**Description of Improvement**
Currently, there is no way to organize how the list of songs in the favorite tracks page is laid out. The songs are added in there based on how recent it was added. I think eventually, there needs to be a way to organize it also by artist or title too. Having a way to organize it such as by dragging or clicking options and sorting would be beneficial as well.

kacperkoza34:
Hello, I would like to work on this one if possible.

WildLeons:
Hi @kacperkoza34,
Are you still working on that feature or not?

kacperkoza34:
Hi @WildLeons
I'm still working on this feature

Blakelist7:
Hello, I would like to contribute to this project as a beginner so could someone help me?
Please.

anontyro:
hey, this looks interesting as there seems to be not much movement here I'll start to play around with this and see what I can do :)

Edu93Jer:
Hi, @skyawmyint is this issue still open?
I am newbie to this project and would like to contribute to it.
I'd like to work on this issue if no one is.

ahmedtarabia:
Hey @nukeop , I would like to look into this if no one else is.

nukeop:
Ok, sure. Do you want me to assign you?
Please remember that we require each new functionality to be covered by tests to a sensible degree.

ahmedtarabia:
Yes, please. I will report back if I run into any issues. 
Thanks

mashazyu:
Hey @nukeop ,

I'd like to work on this issue, if no one else is working on it at the moment. Therefore,
- Does anyone actively work on this issue, or I can go ahead and give it a try?
- If(when) I have further questions, can I just write you directly in Discord?

nukeop:
You can give it a try, I'll assign you. Feel free to contact me on Discord with anything.

nukeop:
Please note that this will be best implemented by adding an ability to sort tracks in the TrackTable component, which will also enable it everywhere this component is used. This should be toggleable with a prop, since we won't want to be able to sort some lists, like albums.

PR #729, opened 2020-06-26
[Feature Request] Organize/Sort songs in Favorite Tracks Page #722

kacperkoza34:
Add posibility to organize songs by titels, artist's names
and in your way

### DIFF
--- a/packages/app/app/actions/favorites.js
+++ b/packages/app/app/actions/favorites.js
@@ -18,6 +18,65 @@ export function readFavorites() {
   };
 }
 
+export function replaceFavoriteTrack(i, moveUp, length) {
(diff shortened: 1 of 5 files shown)


packages/app/app/actions/favorites.js
```diff
@@ -18,6 +18,65 @@ export function readFavorites() {
   };
 }
 
+export function replaceFavoriteTrack(i, moveUp, length) {
```
nukeop, line 21: I don't think this and sorting should be done in the actions. It would be better to handle this functionality with a hook that creates callbacks for sorting arbitrary track lists, and use this hook in a container.

PR #731, opened 2020-07-03
working flatpak.

advaithm:
**we have flatpaks people. please note that this only builds the latest release.**


advaithm:
the travis ci is failing because rpmbuild isn't installed on the system @nukeop

PR #740, opened 2020-07-16
Fix genre images out of sync

MSandeep96:
Decrease font size of genre's title as it was overflowing. Match genre's names in lower case to reduce cases.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ env:
      - GITHASH=${TRAVIS_COMMIT::6}
 
 before_install:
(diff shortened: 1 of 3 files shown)


nukeop:
Do you have a screenshot of what it looks like now? Whether it overflows or not depends on the resolution.

MSandeep96:
image
What it looked like before ^
image
What it looks like with this PR^

I agree that on higher resolutions this might look really small. Should I try a different approach?

nukeop:
Looks alright, thanks for changing this.

PR #743, opened 2020-07-22
fix(build): add icon dir path to electron-builder linux config

charjac:
see simon brandner comment https://github.com/electron-userland/electron-builder/issues/2577
fix#737

I have no debian at my hand so i didn't test it, if someone could do before merge :)

### DIFF
--- a/package.json
+++ b/package.json
@@ -135,6 +135,7 @@
       }
     ],
     "linux": {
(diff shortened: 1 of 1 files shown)


nukeop:
Looks like the icon is packaged in the deb correctly now

PR #750, opened 2020-07-30
Fixed bug for the \Add All\ button

WildLeons:
Hello,
I've been testing stuff to fix this bug and it seems to be working.

I replaced line 94: ".slice(0, expanded ? 15 : 5)"
by ".slice(0, tracks.length > 15 ? 15 : tracks.length)"

The old line required to first scroll down the list before adding the trackss to get all of them otherwise we only have the first 5.
Now it is not necessary to do this anymore and I checked for artists who have less than 15 tracks to be sure than it will work, I tested with artists like "GRIGRI" for example.


nukeop:
Hey, thanks for fixing this. It looks like you're 100% correct and we were assuming that we'd always get at least 15, doesn't seem to be the case though.

PR #752, opened 2020-08-01
automation of flatpak.json updating

advaithm:
the file gen_new.py should be run with like so with the following flag
```shell
$ python gen_new.py --url URL
```
replace the url with the new url


nukeop:
Can you tidy it up a little and remove unrelated commits so it's easier to see what's been changed?

advaithm:
okay

PR #764, opened 2020-08-19
Adding of a button to export lyrics into PDF file

WildLeons:
I place a button under the lyrics part translated into all available languages.

For the PDF, it's save with the name : [Name of the track]_[Name of the artist]_Lyrics.pdf
The file looks like this:
[Name of the track]

[Name of the artist]

[Lyrics]


nukeop:
Great work, thanks for contributing! Do you have an example of what these pdfs look like?

WildLeons:
Yes here is one.
[To The End Of The World_Alestorm_Lyrics.pdf](https://github.com/nukeop/nuclear/files/5107016/To.The.End.Of.The.World_Alestorm_Lyrics.pdf)

WildLeons:
Can you wait before merged ? I am going to fix something important that i didn't notice : there is only a part of the lyrics not all because there is only one page into the PDF file and it isn't enough

WildLeons:
It's fixed and completely functionnal now

WildLeons:
Here is a file created by the corrected function if you want to see
[To The End Of The World_Alestorm_Lyrics.pdf](https://github.com/nukeop/nuclear/files/5114765/To.The.End.Of.The.World_Alestorm_Lyrics.pdf)

WildLeons:
Add a little change for optimization

WildLeons:
I promise you this is the last time, I won't touch it again.

packages/app/app/components/LyricsView/index.js
```diff
@@ -44,6 +70,63 @@ export class LyricsView extends React.Component {
       </div>
     );
   }
+  
+  //Addition of WildLeons
```
nukeop, line 74: This function should be kept somewhere separate from the component and supplied as a prop. Ideally the component needs just an event handler. All this complex logic should be isolated (maybe in the core package), we might want to use it somewhere else in the future (e.g. on mobile).

packages/app/app/components/LyricsView/index.js
```diff
@@ -44,6 +70,63 @@ export class LyricsView extends React.Component {
...
+	  //Margins for shaping
+	  let margins = {
+		  top: 10,
+		  left: 10
+	  };
+	  
+	  //Get the name of the song and its artist and add them at the beginnig of the file
+	  let name = this.props.track.name;
+	  let artist = this.props.track.artist;
+	  pdf.text(name + " - " + artist, margins.left, margins.top);
```
nukeop, line 88: You can use string templates for this `${name} - ${artist}`

packages/app/app/components/LyricsView/index.js
```diff
@@ -18,21 +24,41 @@ export class LyricsView extends React.Component {
...
       lyricsStr = this.props.t('not-found');
-    }
-    return (
-      <div className={styles.lyrics_text}>
-        {lyricsStr}
-      </div>);
+	}
+	
+	//Changes by WildLeons
+	if(lyricsStr === this.props.t('not-found')) {
```
nukeop, line 30: Checking if lyrics are empty should be done in a more generic way than comparing it to the internationalised string

nukeop: Also instead of using if/else, rendering can use && or the ternary operator.
```typescript
{ condition && <Component /> }
{ condition ? <Component/> : <DifferentComponent/> }
```

packages/app/app/components/LyricsView/index.js
```diff
@@ -44,6 +70,63 @@ export class LyricsView extends React.Component {
...
+	  //Retrieve the lyrics only from the props.lyrics like in renderLyrics()
+	  let lyrics = _.get(this.props.lyrics, 'lyricsSearchResults', '');
+	  lyrics = _.get(lyrics, 'type', '');
+	  
+	  //Add variables to store the lyrics that will be on a page
+	  //and count the number of lines already passed before going to a next page
+	  let lyricsOfThePage = "";
+	  let nbrLineBreak = 0;
+	  
+	  for(let i = 0; i < lyrics.length; i++){
```
nukeop, line 99: These pages can be generated much more cleanly e.g. by using this function: https://lodash.com/docs#chunk

nukeop:
I left some comments for you. Could you also run the linter before committing the code (and ideally add some tests)?

WildLeons:
Okay, I'll work quietly on it.

Issue #784, opened 2020-10-29
Small screen size (<= 767px) makes tracks in 'Local Library' disappear


spongechameleon:
Hi all, I am looking for help finding the correct files to apply a fix.

[Quick video of bug and fix](https://i.imgur.com/pLL2AQH.mp4)

## The bug

It appears that at screen widths <= 767px, several styles are applied to the tracks in Local Library that cause them to basically disappear.

## The fix

I've found that by negating three style properties on two elements, the view returns to normal.

First element:
`<thead class="thead">`
Property to be removed:
```
// As viewed in dev tools "Elements" tab

@media only screen and (max-width: 767px)
.ui.table:not(.unstackable) thead {
    display: block;
}
```
Second element:
`<tbody class="tbody">`
Properties to be removed:
```
// As viewed in dev tools "Elements" tab

@media only screen and (max-width: 767px)
.ui.table:not(.unstackable) tbody, .ui.table:not(.unstackable) tr,
.ui.table:not(.unstackable) tr>td, .ui.table:not(.unstackable) tr>th {
    width: auto !important;
    display: block !important;
}
```
## The appropriate file

It seems the problem elements are rendered from `app/app/components/LibraryView/LibrarySimpleList/index.js`, however, I cannot find any media queries in the corresponding `styles.scss` file (same directory), so I'm not too sure how to fix this.

Some direction would be much appreciated.

nukeop:
Thanks for the report. I think these styles are coming from semantic-ui-css:
https://github.com/Semantic-Org/Semantic-UI-CSS/blob/fc9f8bd7e4f5934756ec60c0423f77d1c7be7c6f/components/table.css#L151

PR #786, opened 2020-10-30
Make tracks in Local Library visible at small screen widths

spongechameleon:
[Link to issue](https://github.com/nukeop/nuclear/issues/784)

## To test:

- Open up dev tools
- Go to 'Local Library' (icon at the very bottom of the left pane)
- Expand the dev tools such that nuclear's screen width is <767px

### Before:

- Observe that track information 'disappears' and becomes unusable

### Now:

- Observe that track information is visible in the same format as when the screen is >767px

P.S. thanks to nuke here's a [link](https://github.com/Semantic-Org/Semantic-UI-CSS/blob/fc9f8bd7e4f5934756ec60c0423f77d1c7be7c6f/components/table.css#L151) to the Semantic UI styles that have been overridden

### DIFF
--- a/packages/app/app/components/LibraryView/LibrarySimpleList/styles.scss
+++ b/packages/app/app/components/LibraryView/LibrarySimpleList/styles.scss
@@ -49,3 +49,42 @@
     }
   }
 }
+
+/* Override Semantic UI small screen styles */
+@media only screen and (max-width: 767px) {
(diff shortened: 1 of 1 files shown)


nukeop:
Thanks for fixing this. I don't like the `!important` everywhere but I guess it's one way of overriding these styles.

PR #788, opened 2020-10-31
Scale down bottom player bar <=767px screen width

spongechameleon:
### What and why

I've made the bottom player bar with the track info & playback controls shrink at screen widths <= 767px (which is when the Semantic UI 'small' breakpoint hits). This makes it easier to use the controls at smaller screen widths.

### What's been tested

- The `useEffect` hook in `VolumeSlider` only runs once at mount :100:
- Performance feels the same

### Potential Issues

- New prop on the `Range` component (`.../ui/lib/components/Range`)

Added:
```
<div ... ref={rangeRef}>
```
Reason to do so is to prevent `VolumeSlider` from re-rendering unnecessarily (`resize` event listener in `VolumeSlider` uses the DOM state from this `ref` directly instead of having to go through a `useState` in `VolumeSlider`, which would involve removing and setting new `resize` event listeners whenever the `useState` state changed).

Anyways, not sure if the default prop `rangeRef: null` will work. Not sure if anything other than `VolumeSlider` renders `Range`, though.

If other components do render `Range`, init'ing and passing 
```
const dummyRef = useState(null);
...
    rangeRef={dummyRef}
``` 
would prevent any errors, if `null` turns out to not work.

If this is too sketch I could just remove the new prop from `Range` and keep everything local to `VolumeSlider`. That would mean `useEffect` would run more often but honestly people probably aren't re-sizing the application very often so I'm guessing the performance won't really be affected either way. Everything seemed smooth to me while testing.

### DIFF
--- a/packages/ui/lib/components/Cover/styles.scss
+++ b/packages/ui/lib/components/Cover/styles.scss
@@ -11,3 +11,11 @@
   width: 76px;
   height: 76px;
 }
+
+@media (max-width: 767px) {
+  .nuclear.cover_container,
+  .nuclear.cover_container img {
+    width: 50px;
+    height: 50px;
+  }
+}
--- a/packages/ui/lib/components/PlayerBar/styles.scss
+++ b/packages/ui/lib/components/PlayerBar/styles.scss
@@ -37,3 +37,11 @@
     pointer-events: none;
   }
 }
+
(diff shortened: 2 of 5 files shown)


nukeop:
Can't we resize the volume slider with css instead of these complex hooks?

Issue #720, opened 2020-06-04
Local music from added folder not being added to Local Library


skyawmyint:
**Version of nuclear (written under the main logo)**
0.6.3

**Your operating system**
Windows 10 Home

**Reproducible steps (1... 2... 3...) that cause the issue**
Step 1: Go to the Local Library tab from the left menu
Step 2: Click on add folders button
Step 3: Go to a folder that has music in it and click select folder
Step 4: Click the refresh button next to add folder

**What you expected to see, versus what you actually saw**
When adding a folder from the local desktop, the directory is made path, but none of the songs from that folder is being brought onto the application and the Local Library remains empty.  There is also no button to add songs to the Local Library.
Expected the songs from the folder that was specified in the add folders should be in the library now. The actual results were that the songs from the folder that was specified in the add folders was not added to the library. The library shows the it is still empty.

Images, animations, or a link to a video showing the issue occurring
![LocalLibrary_Example1](https://user-images.githubusercontent.com/44180667/83703021-3125fb80-a5c3-11ea-9ef3-bb1b7b70c2f8.jpg)

**A code snippet that demonstrates the issue or a link to a code repository the developers can easily pull down to recreate the issue locally**
N/A

nukeop:
What formats of files are inside your library? Could you please give me an example of an album folder?

masoud-msk:
It happened for me when adding a large directory containing 4000 files. I narrowed the problem down to a single file which caused the error `ipc api › error in event set-localfolders => pictureType should be defined`.
Note that *VLC* can successfully play the file with album art.

Venryx:
I'm getting an issue, possibly the same, where Nuclear gets stuck scanning/refreshing the files in my library:
![](https://i.imgur.com/Hw8UAQJ.png)

I searched Process Hacker 2 for the list of commands running under the Nuclear root folder, and it showed the many `fpcalc` instances as it attempted to scan the files. When the scanning stopped in Nuclear's UI, I checked that list again, and no `fpcalc` commands. So it's not fpcalc hanging, it's something higher level.

It's worth noting that _some_ files from my library folder did end up showing up in Nuclear after the scan-freeze. But only those from the "Foreign" folder, not the other two.

It's odd because Nuclear used to be able to scan and present the file from my entire library. So some change in the last half year or so has caused it to choke on certain files, apparently.

Venryx:
I looked at the console, and found this:
```
[...]

PR #795, opened 2020-12-01
Fix for track-recognition error breaking library scan

Venryx:
Added a fix for the issue mentioned here: https://github.com/nukeop/nuclear/issues/720#issuecomment-689836722

This pull request adds a try-catch to the code processing the metadata returned by acoustic-id for a set of tracks.

Previously, if an error was encountered during this metadata processing (eg. "Empty fingerprint", as I hit), it would cause the entire local-library scanning process to fail. This was terrible for the user-experience, as it made Nuclear as a whole unusable, even if only a couple files in the library had the track-recognition issue.

### DIFF
--- a/packages/main/src/services/local-library/index.ts
+++ b/packages/main/src/services/local-library/index.ts
@@ -76,10 +76,14 @@ class LocalLibrary {
     const formattedMetas = await Promise.all(filesPath.map((file, i) => this.formatMeta(metas[i], file)));
 
     if (this.config.isConnected) {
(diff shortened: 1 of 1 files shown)


Venryx:
Closing this, as the fetchAcousticIdBatch function already has a try-catch inside it; but it was failing to catch the error, because the error was just outside of the try-catch block. (due to the code assuming that if a result was returned, the result would have an artist field, but apparently that's not always the case)

Added a pull-request for that source-issue fix instead: https://github.com/nukeop/nuclear/pull/796

PR #805, opened 2021-01-16
Update Swedish translation

jony0008:
Just a note: Usually 'sv' is used an abbreviation of Svenska (Swedish) and not 'se'. I don't have much skill in programming so I don't know if that matters here.


nukeop:
You're correct that the appropriate language code would be sv in ISO 639-1, but the library we're using (i18next) and its language detector use `se`.

PR #819, opened 2021-02-08
selectable text in lyric view

HaiDang666:
It'll be nice if users can also select `title` and `artist name`


nukeop:
Great, thanks!

Issue #811, opened 2021-01-28
Add tooltips to the icons in the player bar


nukeop:
Some of the icons are not self-explanatory, they need tooltips saying what they do

KashyapPavra:
@nukeop , I would love to work on this feature. Could you brief where and what you would like to name the tooltips?

nukeop:
Ok, so to render the tooltips we should use this component: https://react.semantic-ui.com/modules/popup/

![image](https://user-images.githubusercontent.com/12746779/106177449-6dc8e780-6190-11eb-952c-ac064772a555.png)

These are the buttons that need them. Left to right they are: `repeat`, `shuffle`, `autoradio`, `miniplayer`. Of course these strings need to be placed in en.json and properly localized. For now we can only include the english names and worry about other languages later.

KashyapPavra:
@nukeop  Great, I'll have a look at it this weekend. Can you assign me this issue?

KashyapPavra:
@nukeop been trying to start the application using ``npm start`` on MacOS 11.1 but have been getting this error:

```
@nuclear/main: /Users/kashyappavra/Desktop/OpenSource/nuclear/packages/main/node_modules/electron/dist/Electron.app/Contents/MacOS/Electron exited with signal SIGABRT
@nuclear/main: npm ERR! code 1
@nuclear/main: npm ERR! path /Users/kashyappavra/Desktop/OpenSource/nuclear/packages/main
@nuclear/main: npm ERR! command failed
@nuclear/main: npm ERR! command sh -c shx rm -rf build && webpack --env.NODE_ENV=development && electron build/main.js
```

Any idea what's causing this?

nukeop:
Nope, but you can try following the process we use in CI which is guaranteed to work on the latest Mac OS (since we run it after every push to master): https://github.com/nukeop/nuclear/blob/master/.github/workflows/build.yml#L36

KashyapPavra:
@nukeop still not able to get it running. Maybe you can assign this issue to someone else.

HaiDang666:
I can take this if @KashyapPavra wants

KashyapPavra:
> I can take this if @KashyapPavra wants

@HaiDang666 Sure I don't mind.

HaiDang666:
fixed in https://github.com/nukeop/nuclear/pull/822

PR #822, opened 2021-02-09
#811 tooltips for playerbar icon

HaiDang666:
image


nukeop:
Thanks, looks pretty great.

Issue #820, opened 2021-02-08
Add links to artist's albums/merchandise when using Bandcamp


nukeop:
This would mesh with: https://github.com/nukeop/nuclear/issues/131

The bandcamp scraper gives us links to relevant pages. It would be a nice way to support the artists if the links were displayed on artist's and album's pages in a prominent place (up to implementer where exactly). In the future we could promote interested artists by displaying their bandcamp links somewhere in the dashboard.

Yahyakiani:
@nukeop Can I have this issue? First timer here thanks

nukeop:
Sure, I'll assign it to you.

5usana:
Hello! If this is still an open issue i'd love a shot at tackling it. May I?

nukeop:
> Hello! If this is still an open issue i'd love a shot at tackling it. May I?

Sure, no problem. This still needs some improvements in the bandcamp scraper to provide links to merchandise.

jreiser1:
Hello, I've got a working prototype for this issue. Can I be assigned to it? I'm hoping to open a pull request in the next day or so. 

Right now I've implemented an Artist Info Button on the ArtistView header that will open up the artist's Bandcamp/iTunes/Discogs/Audius/Musicbrainz page depending on which metadata provider the user has selected. Audius, Discogs, and MusicBrainz will currently only open a search page with the artist's name provided, but I think that's a good start at least. Bandcamp and iTunes will open directly to the artist's page.

![image](https://user-images.githubusercontent.com/47801590/141892845-52040946-56c9-4809-91ae-c46e584e4eb8.png)

nukeop:
Thanks, this looks good. Could you please open a pull request?

jreiser1:
Sure, will open one in a few minutes

PR #827, opened 2021-02-10
Artist links

Yahyakiani:
Resolves Issues #820 .
Current Outlook of Link Designs is like so on Artist and Album Page:
v1
v2

### DIFF
--- a/packages/app/app/components/AlbumView/index.js
+++ b/packages/app/app/components/AlbumView/index.js
@@ -12,6 +12,7 @@ import { safeAddUuid } from '../../actions/helpers';
 
 import styles from './styles.scss';
 import artPlaceholder from '../../../resources/media/art_placeholder.png';
+import ArtistLink from '../ArtistView/ArtistLink';
 
 @withTranslation('album')
 class AlbumView extends React.Component {
@@ -154,6 +155,7 @@ class AlbumView extends React.Component {
           <div className={styles.album_buttons}>
             {this.renderPlayAllButton(album)}
             {this.renderOptionsButtons(album)}
+            <ArtistLink artistLink={this.props.album.uri} circular={true} size={'small'} />
           </div>
         </div>
         <a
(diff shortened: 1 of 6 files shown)


packages/app/app/components/ArtistView/ArtistLink/index.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import { Button } from 'semantic-ui-react';
+import FontAwesome from 'react-fontawesome';
+
+const shell = require('electron').shell;
+
+const ArtistLink = ({ artistLink, size, label='Visit', circular=false}) => (
+  <Button
+    primary
+    onClick={() => shell.openExternal(artistLink)}
```
nukeop, line 10: Can you put this in `useCallback`?

packages/app/app/components/ArtistView/ArtistLink/index.js
```diff
@@ -0,0 +1,19 @@
...
+    size={size}
+    circular={circular}
+    aria-label={'Visit Artist Profile'}
+  >
+    <FontAwesome name='link' /> {label}
```
nukeop, line 15: This label should be localized, can you use the localization API and create a string in en.json?

packages/app/app/components/AlbumView/index.js
```diff
@@ -154,6 +155,7 @@ class AlbumView extends React.Component {
           <div className={styles.album_buttons}>
             {this.renderPlayAllButton(album)}
             {this.renderOptionsButtons(album)}
+            <ArtistLink artistLink={this.props.album.uri} circular={true} size={'small'} />
```
nukeop, line 158: This is not exactly what I meant here. We need to check if this data is coming from Bandcamp, and use relevant links for merch/albums. We'll want an album link here (which looks ok) and the merchandise link on the artist page (which might have to be passed down from the metadata plugin in case it's not there yet).

packages/app/app/components/ArtistView/ArtistLink/index.js
Yahyakiani: @nukeop Do you mean Memoize the functional component and usecallback for onclick handler right ? Because Memoize is usually used for efficient rendering.Do correct me if I am wrong.

nukeop: No need to memoize the whole component, just put the onclick handler in a `useCallback` hook so we don't re-create this function on each render.

PR #838, opened 2021-02-19
SponsorBlock API

HaiDang666:
It works with both PlayerBar and MiniPlayer.

### DIFF
--- a/packages/core/src/rest/SponsorBlock.ts
+++ b/packages/core/src/rest/SponsorBlock.ts
@@ -25,8 +25,16 @@ export async function getSegments (videoID: VideoID, categories?: Category[]): P
           return r.json();
         })
         .then(j => {
+          if (!j) {
+            return resolve([]);
+          }
+
           let segments = j.map(({category, segment}) => {
-            return {startTime: Math.round(segment[0]), endTime: Math.round(segment[1]), category};
+            return {
+              category,
+              startTime: Math.round(segment[0]), 
+              endTime: Math.round(segment[1])
+            };
           });
 
           segments = _.sortBy(segments, ['startTime']);
--- a/packages/core/src/rest/SponsorBlock.ts
+++ b/packages/core/src/rest/SponsorBlock.ts
@@ -1,5 +1,6 @@
 import { VideoID, Category, Segment } from './SponsorBlock.types';
 import _ from 'lodash';
+import logger from 'electron-timber';
 
 const BASE_URL = 'https://sponsor.ajay.app';
 export const ALL_CATEGORIES = ['sponsor', 'intro', 'outro', 'interaction', 'selfpromo', 'music_offtopic'];
@@ -9,11 +10,9 @@ export async function getSegments (videoID: VideoID, categories?: Category[]): P
 
   if (!categories) {
     query += `&categories=${JSON.stringify(ALL_CATEGORIES)}`;
-  } else if (categories.length > 1) {
+  } else if (categories.length) {
     query += `&categories=${JSON.stringify(categories)}`;
-  } else if (categories.length === 1) {
-    query += `&category=${categories[0]}`;
-  } 
+  }
 
   return new Promise((resolve) => {
     try {
@@ -38,9 +37,8 @@ export async function getSegments (videoID: VideoID, categories?: Category[]): P
           });
 
           segments = _.sortBy(segments, ['startTime']);
-          const temp = _.concat(segments);
-          segments = _.filter(_.uniq(segments), (v: Segment) => {
-            for (const s of temp) {
+          segments = _.filter(segments, (v: Segment) => {
+            for (const s of segments) {
               if (s === v) {
                 continue;
               }
@@ -54,7 +52,9 @@ export async function getSegments (videoID: VideoID, categories?: Category[]): P
           resolve(segments);
         });
     } catch (error) {
-      resolve([]);
+      logger.error(`An error when getting skipsegment from SponsorBlock, videoID: ${videoID}`);
+      logger.error(error);
+      return [];
     }
   });
 }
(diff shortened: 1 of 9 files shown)


packages/core/src/rest/SponsorBlock.ts
```diff
@@ -0,0 +1,60 @@
...
+
+  if (!categories) {
+    query += `&categories=${JSON.stringify(ALL_CATEGORIES)}`;
+  } else if (categories.length > 1) {
+    query += `&categories=${JSON.stringify(categories)}`;
+  } else if (categories.length === 1) {
+    query += `&category=${categories[0]}`;
+  } 
+
+  return new Promise((resolve) => {
```
nukeop, line 18: This can be simplified like this:
```typescript
try {
    const result = await fetch(`${BASE_URL}/api/skipSegments${query}`);
    const json = await result.json();

    //rest of the processing
} catch(e) {
    // Log error
} finally {
    return [];
}
```

nukeop, line 18: This can be simplified like this:
```typescript
try {
    const result = await fetch(`${BASE_URL}/api/skipSegments${query}`);
    const json = await result.json();

    //rest of the processing
} catch(e) {
    // Log error
} finally {
    return [];
}
```

nukeop, line 18: This can be simplified like this:
```typescript
try {
    const result = await fetch(`${BASE_URL}/api/skipSegments${query}`);
    const json = await result.json();

    //rest of the processing
} catch(e) {
    // Log error
} finally {
    return [];
}
```

packages/core/src/rest/Youtube.ts
```diff
@@ -112,14 +113,16 @@ export async function trackSearchByString(query: string, omitStreamId?: string,
...
   return {
     source: sourceName,
     id: topTrack.id,
     stream: formatInfo.url,
     duration: parseInt(topTrackInfo.videoDetails.lengthSeconds),
     title: topTrackInfo.videoDetails.title,
     thumbnail: topTrack.bestThumbnail.url,
-    format: formatInfo.container
+    format: formatInfo.container,
+    skipSegment: segments
```
nukeop, line 126: Name it `skipSegments` (plural)

packages/core/src/rest/SponsorBlock.ts
```diff
@@ -0,0 +1,60 @@
...
+          let segments = j.map(({category, segment}) => {
+            return {
+              category,
+              startTime: Math.round(segment[0]), 
+              endTime: Math.round(segment[1])
+            };
+          });
+
+          segments = _.sortBy(segments, ['startTime']);
+          const temp = _.concat(segments);
```
nukeop, line 41: This line won't do anything.

packages/core/src/rest/SponsorBlock.ts
```diff
@@ -0,0 +1,60 @@
...
+          segments = _.filter(_.uniq(segments), (v: Segment) => {
```
nukeop, line 42: What segments are you trying to filter out here? Can you write a couple of tests to make sure it's doing what you intended?

packages/ui/lib/components/Seekbar/index.tsx
```diff
@@ -7,22 +7,45 @@ import styles from './styles.scss';
...
   children,
   fill,
   seek,
   queue,
-  height
+  height,
+  skipSegment,
+  timePlayed
 }) => {
+  const hasMounted = useRef(false);
```
nukeop, line 35: Shouldn't this be a `useState`?

packages/core/src/rest/SponsorBlock.ts
HaiDang666: The api doesn't sort those segments in order of `startTime` so it will be not easy when comparing with `timePlayed`.
The purpose of filter is removing segment that is sub in another, like below:
Screen Shot 2021-02-26 at 09 42 14
I gonna add some tests

packages/core/test/sponsorblock.test.ts
```diff
@@ -0,0 +1,56 @@
...
+
+mock('electron-store', MockStore);
+mock('electron-timber', MockLogger);
+mock('electron', electron);
+
+import { rest } from '../src';
+
+test('has empty result', async (t) => {
+  const videoID = 'Ri7-vnrJD3k';
+  const segments = await rest.SponsorBlock.getSegments(videoID);
```
nukeop, line 16: Can you replace this call with some static data formatted the same way as an actual api response? The reason for this is that we don't want to needlessly strain sponsorblock's api, since it's being offered for free (and it's good practice to forgo external dependencies in tests).

nukeop:
Thanks for your changes, I have one more suggestion and then we're ready to go.

packages/core/test/sponsorblock.test.ts
HaiDang666: Uhm i don't get it. You mean remove the `await rest.SponsorBlock.getSegments(videoID);` by an empty array or like as below:
Screen Shot 2021-03-02 at 11 51 41

nukeop: What you posted in the screenshot looks fine, you can either use different pieces of static data for all the tests, or create a builder function that will create mock data with the characteristics you need for each test.

HaiDang666:
added fake response from SponsorBlock server to test formatResponse

nukeop:
Great, thank you for this.

ajayyy:
Hey, could you add a note to the readme with a link to SponsorBlock? It would probably make most sense in the features list.

HaiDang666:
> Hey, could you add a note to the readme with a link to SponsorBlock? It would probably make most sense in the features list.

added in https://github.com/nukeop/nuclear/pull/851

PR #843, opened 2021-02-26
update eslint script

HaiDang666:
It seems that the current `eslint` scripts in `package.json` cover all file for me. After modifying, it checks all of the file I want.


nukeop:
Great, thank you

PR #851, opened 2021-03-04
ref SponsorBlock in readme

HaiDang666:


nukeop:
Thanks for adding this so fast!

PR #871, opened 2021-03-09
broken icon in tablehead of playlist 

HaiDang666:
before:
Screen Shot 2021-03-09 at 10 02 40

after:
Screen Shot 2021-03-09 at 09 56 48


nukeop:
Yeah I noticed this.  A long term solution is to use localIdentName in css-loader.

Issue #863, opened 2021-03-07
Full album queued instead of title track 


azumukupoe:
I have a problem with: 
Mike Oldfield/Platinum/Platinum
Mike Oldfield/QE2/QE2
Mike Oldfield/Crises/Crises
Mike Oldfield/Discovery/Discovery ...
(same should happen with albums that have tracks with the name of album title)

It happens when: Hitting play in album page

My OS is: Windows

I'm using Nuclear version:0.6.10

HaiDang666:
When you click play button in an album page, the app will add all album's tracks to the queue in order to play the whole album. This is expected behavior.

azumukupoe:
I don't think you understand me...
An entire album gets queued as a track because some people upload full album as one file on YouTube

HaiDang666:
I get it now. The album track list is from Discogs. It has an 0 duration track and the app still search for it on youtube
<img width="843" alt="Screen Shot 2021-03-09 at 09 26 23" src="https://user-images.githubusercontent.com/12780733/110409281-9b6f4d80-80b9-11eb-9e4c-93586f739080.png">

HaiDang666:
I'm working on this.

PR #874, opened 2021-03-11
add album in youtube search terms

HaiDang666:
this support https://github.com/nukeop/nuclear/issues/863 when songs title the same as album title

### DIFF
--- a/packages/app/app/actions/queue.ts
+++ b/packages/app/app/actions/queue.ts
@@ -51,7 +51,8 @@ export const addToQueue = (item) => {
       try {
         const streamData = await selectedStreamProvider.search({
           artist: item.artist,
-          track: item.name
+          track: item.name,
+          album: item.album
         });
         dispatch(updateQueueItem({
           ...item,
--- a/packages/app/app/components/TrackRow/index.js
+++ b/packages/app/app/components/TrackRow/index.js
@@ -42,7 +42,8 @@ class TrackRow extends React.Component {
       name: this.props.track.name,
       thumbnail: this.getTrackThumbnail(),
       local: this.props.track.local,
-      streams: this.props.track.streams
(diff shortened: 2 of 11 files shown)


nukeop:
In the case of that Mike Oldfield album, we search for `Mike Oldfield - Platinum Platinum`, which results in the same top result: https://www.youtube.com/results?search_query=mike+oldfield+-+platinum+platinum

Instead of matching the album name, maybe it's better to take the duration into account?

HaiDang666:
> In the case of that Mike Oldfield album, we search for `Mike Oldfield - Platinum Platinum`, which results in the same top result: https://www.youtube.com/results?search_query=mike+oldfield+-+platinum+platinum
> 
> Instead of matching the album name, maybe it's better to take the duration into account?

I was thinking about using duration but still not sure. 

FYI: the case is `Mike Oldfield - part 1: airbone` and `Mike Oldfield - part 1: airbone platinum`. Cos we already handled subtracks.
result without album: https://www.youtube.com/results?search_query=mike+oldfield+part+1%3A+airborne
with album: https://www.youtube.com/results?search_query=mike+oldfield+part+1%3A+airborne+platinum

nukeop:
We could use a heuristic checking if the top stream has a duration longer than e.g. 60s than the track we're looking for, and if so, reject it (since streams sometimes have varying duration).

HaiDang666:
> We could use a heuristic checking if the top stream has a duration longer than e.g. 60s than the track we're looking for, and if so, reject it (since streams sometimes have varying duration).

To get duration from stream it takes one request. So if the top stream duration doesn't match the condition we have to request for second or third stream. Imo adding album in the search term will provide more correct result for the top stream so it can reduce one request, particularly this case.

Best solution for now is we do both: adding album title and checking duration.

PR #879, opened 2021-03-17
Discogs returns heading in tracklist array

HaiDang666:
before:
Screen Shot 2021-03-17 at 17 21 19

after:
Screen Shot 2021-03-17 at 17 20 34

discord api: https://api.discogs.com/masters/1334523


HaiDang666:
I think it should display the heading but not clickable, so the position still makes sense? what do you think @nukeop ?

nukeop:
> I think it should display the heading but not clickable, so the position still makes sense? what do you think @nukeop ?

Yes, this is a very good idea. If you can, try designing the row for this. Spotify/Discogs should already be handling this so maybe we should draw inspiration from them.

HaiDang666:
look better
Screen Shot 2021-03-18 at 11 09 00

PR #882, opened 2021-03-18
Sonarcloud bugs

HaiDang666:
It's annoying me

### DIFF
--- a/packages/app/app/components/FavoriteTracksView/index.js
+++ b/packages/app/app/components/FavoriteTracksView/index.js
@@ -47,7 +47,6 @@ const FavoriteTracksView = ({
   };
 
   const addTracksToQueue = () => {
-
     tracks.map(track => {
       const trackImage = getTrackimage(track);
       addToQueue({
@@ -58,13 +57,12 @@ const FavoriteTracksView = ({
     });
   };
 
-  const playAll = (tracks) => {
+  const playAll = () => {
     clearQueue();
-    addTracksToQueue(tracks);
+    addTracksToQueue();
     selectSong(0);
     startPlayback();
   };
-
 
   const renderPlayAllButton = () => {
     return (
@@ -74,7 +72,6 @@ const FavoriteTracksView = ({
     );
   };
 
-  
   return (
     <div className={styles.favorite_tracks_view}>
       {
--- a/packages/app/app/components/SearchResults/TracksResults/index.js
+++ b/packages/app/app/components/SearchResults/TracksResults/index.js
@@ -21,7 +21,7 @@ const TracksResults = ({ tracks, limit }) => {
             {_.take(collection, limit).map((track, index) => {
               if (
                 track &&
-                _.hasIn(track, 'name') & _.hasIn(track, 'image') &&
+                _.hasIn(track, 'name') && _.hasIn(track, 'image') &&
                 _.hasIn(track, 'artist')
               ) {
                 const newTrack = _.cloneDeep(track);
--- a/packages/app/app/utils.js
+++ b/packages/app/app/utils.js
@@ -16,7 +16,7 @@ export function formatDuration(duration) {
     seconds = '0' + seconds;
   }
 
-  if (hours === '00') {
+  if (String(hours) === '00') {
     return minutes + ':' + seconds;
   } else {
     return hours + ':' + minutes + ':' + seconds;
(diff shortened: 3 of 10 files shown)


packages/main/src/services/http/server/schema.ts
```diff
@@ -11,6 +11,7 @@ export const getSettingsSchema = {
       option: {
         type: 'string',
         enum: settingsConfig
+          // I'm not sure about this RESTRICTED_SETTINGS is updated in somewhere
           .filter(({ name }) => !RESTRICTED_SETTINGS.includes(name as never))
```
HaiDang666, line 15: I need help with this.

nukeop:
Thanks, I meant to take a look at sonarcloud after fixing the plugins issue. I'm sure there's a lot of stuff that needs to be fine tuned/excluded for it to be more useful.

packages/main/src/services/http/server/schema.ts
nukeop: This, as well as READONLY_SETTINGS, was made so that we're safe in the future - currently has no effect but we should keep it.

HaiDang666:
> Thanks, I meant to take a look at sonarcloud after fixing the plugins issue. I'm sure there's a lot of stuff that needs to be fine tuned/excluded for it to be more useful.

I just touched the `Reliability` section which is simple. `Security` and `Code Smell` looking hard for me now.

packages/app/app/utils.js
```diff
@@ -16,7 +16,7 @@ export function formatDuration(duration) {
     seconds = '0' + seconds;
   }
 
-  if (hours === '00') {
+  if (String(hours) === '00') {
```
nukeop, line 19: This will need to be reviewed in details since we sometimes use numbers (if you just do a `String(0)` this will never be true)

packages/app/app/components/FavoriteTracksView/index.js
```diff
@@ -58,9 +58,9 @@ const FavoriteTracksView = ({
     });
   };
 
-  const playAll = (tracks) => {
+  const playAll = () => {
     clearQueue();
-    addTracksToQueue(tracks);
+    addTracksToQueue();
```
nukeop, line 63: Can you cover this with a test to make sure it actually plays the tracks?
You can do it similarly to the existing playlist view tests.

packages/app/app/components/SearchResults/TracksResults/index.js
```diff
@@ -21,7 +21,7 @@ const TracksResults = ({ tracks, limit }) => {
             {_.take(collection, limit).map((track, index) => {
               if (
                 track &&
-                _.hasIn(track, 'name') & _.hasIn(track, 'image') &&
+                _.hasIn(track, 'name') && _.hasIn(track, 'image') &&
```
nukeop, line 24: Has this even worked before?

packages/app/app/utils.js
HaiDang666: In the code, `hours` was checked for less than `10` so it cover the case `hours = 0` 
```js
  let hours = Math.floor(secNum / 3600);

  if (hours < 10) {
    hours = '0' + hours;
  }

  if (String(hours) === '00') {
    return minutes + ':' + seconds;
  }
```

packages/app/app/components/FavoriteTracksView/index.js
HaiDang666: In the original code, the `addTracksToQueue` doesn't use `tracks` from parameter (it uses `tracks` from `props`), also when `playAll` called in `onClick` it doesn't pass any parameter. 
```js
  const addTracksToQueue = () => {
    tracks.map(track => {
      const trackImage = getTrackimage(track);
      addToQueue({
        artist: track.artist.name,
        name: track.name,
        thumbnail: trackImage
      });
    });
  };

  const playAll = (tracks) => {
    clearQueue();
    addTracksToQueue(tracks);
    selectSong(0);
    startPlayback();
  };

  const renderPlayAllButton = () => {
    return (
      <a href='#' className={styles.play_button} onClick={playAll}>
        <Icon name='play'/> Play
      </a>
    );
  };
```

packages/app/app/components/SearchResults/TracksResults/index.js
HaiDang666: Yes it's worked well,  `_.hasIn(track, 'name') & _.hasIn(track, 'image')` return only `1` or `0` cos `hasIn` return boolean and only case for it `return 1` is both give `true`.  As a whole expression only return true, 0 or false which perfectly works with this `if`. 

This change just make the expression to be consistent return `false` instead of `0`.

nukeop:
Ok, great! Thanks for fixing all this.

PR #885, opened 2021-03-22
Refactor data format from plugin api

HaiDang666:
Metadata
- [x] Discogs
- [x] Audius
- [x] Musicbrainz
- [x] Bandcamp

Stream providers
- [x] Youtube
- [x] Audius
- [x] Bandcamp
- [x] Invidious
- [x] Soundcloud
- [x] Jamendo

### DIFF
--- a/packages/core/src/plugins/meta/discogs.ts
+++ b/packages/core/src/plugins/meta/discogs.ts
@@ -91,17 +91,17 @@ class DiscogsMetaProvider extends MetaProvider {
     });
 
     return {
-      ...release,
-      tracklist,
-      resourceUrl: release.resource_url,
       id: `${release.id}`,
       artist: _.head(release.artists).name,
+      title: release.title,
       thumb: coverImage,
       coverImage,
       images: _.map(release.images, 'resource_url'),
       genres: [..._.map(release.genres), ..._.map(release.styles)],
+      year: `${release.year}`,
       type: releaseType,
-      year: `${release.year}`
+      tracklist,
+      resourceUrl: release.resource_url
     };
   }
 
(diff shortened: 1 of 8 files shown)


packages/core/src/plugins/meta/discogs.ts
```diff
@@ -146,10 +150,22 @@ class DiscogsMetaProvider extends MetaProvider {
...
-      .then(json => json.map(result => ({
-        ...result,
-        source: SearchResultsSource.Discogs
-      })));
+      .then(json => {
+        const artists = [];
+        const releases = [];
+
+        if (json.results) {
+          json.results.map(item => {
```
nukeop, line 158: instead of pushing inside map, just assign the maps to consts (it's fine to map twice)

HaiDang666: How to do it? I haven't done it before.

HaiDang666: how ab this? cos  map return "undefined" if we don't return any
```js
  if (json.results) {
    const artists = json.results.flatMap(item => 
      (item.type === 'artist') ?
        [this.discogsArtistSearchResultToGeneric(item)] : []
    );

    const releases = json.results.flatMap(item =>
      (item.type === 'master' || item.type === 'release' ) ?
        [this.discogsReleaseSearchResultToGeneric(item)] : []
    );

    return Promise.resolve({ artists, releases, tracks: [] });
  }

  return Promise.resolve({ artists: [], releases: [], tracks: [] });
```

nukeop: Yes, this is great.

In some situations you can use this: https://lodash.com/docs/4.17.15#compact

But your approach is better here

PR #895, opened 2021-03-29
Allow add track to playlist in search result

HaiDang666:


HaiDang666:
Linux test failed not related to the changes

PR #897, opened 2021-03-31
Add ArtistView test

HaiDang666:


nukeop:
Btw, it's good to have tests for this section, as it will be the first step towards refactoring it.
I designed this whole flow wrong. When you go to an artist page, whatever link leads to it initiates a search, then after an id is retrieved, goes to `artist/[id]`. It should be the other way around - first go to ArtistView, then ArtistView itself fetches the data it needs to display. I will refactor it to follow this pattern once we get the tests out of the way so we can be reasonably sure it's implemented correctly this time.

HaiDang666:
> Btw, it's good to have tests for this section, as it will be the first step towards refactoring it.
> I designed this whole flow wrong. When you go to an artist page, whatever link leads to it initiates a search, then after an id is retrieved, goes to `artist/[id]`. It should be the other way around - first go to ArtistView, then ArtistView itself fetches the data it needs to display. I will refactor it to follow this pattern once we get the tests out of the way so we can be reasonably sure it's implemented correctly this time.

you are correct, when making the test I found that this problem but not directly to what you said. It was the similar artists. When clicking on artist name, it triggers `searchArtistByName` and redirects to `artist/[id]` after receiving the artist data for `artistDetails`. In the mean time, the UI's still at the previous artist and doesn't have any signal that similar artist is loading and users will think that nothing happens and they keep clicking on other section. Thus I made a simple solution that doesn't involve many changes by using a flag. I comment on the code changes below.

packages/app/app/actions/search.ts
```diff
@@ -215,6 +215,8 @@ export const artistInfoSearchByName = (artistName, history) => async (dispatch,
   const selectedProvider = getSelectedMetaProvider(getState);
   const { settings } = getState();
   try {
+    _.invoke(history, 'push', '/artist/loading');
+    dispatch(artistInfoStart('loading'));
```
HaiDang666, line 219: loading flag for when clicking a similar artist

packages/app/app/containers/ArtistViewContainer/index.tsx
```diff
@@ -0,0 +1,28 @@
...
+const ArtistViewContainer: React.FC = () => {
+  const props = useArtistViewProps();
+  const { artistId } = useParams<{ artistId: string }>();
+  const dispatch = useDispatch();
+  const artistReleasesSearch = useCallback(async (artistId) => {
+    dispatch(SearchActions.artistReleasesSearch(artistId));
+  }, [dispatch]);
+
+  useEffect(() => {
+    if (artistId !== 'loading') {
```
HaiDang666, line 18: loading flag for when clicking a similar artist

nukeop:
>  In the mean time, the UI's still at the previous artist and doesn't have any signal that similar artist is loading and users will think that nothing happens and they keep clicking on other section. Thus I made a simple solution that doesn't involve many changes by using a flag. I will comment on the code changes below.

Great, this is exactly the problem that was bothering me. I was thinking of having two different behaviors for artist view:
- If we supply an id, it uses the currently selected metadata provider to grab everything and load the page
- If we only supply a name, it firsts try to search for that name, then proceeds to the id page if it gets a match

nukeop:
But your solution is also good and pretty similar to what I had in mind.

packages/app/app/containers/ArtistViewContainer/ArtistViewContainer.test.tsx
```diff
@@ -0,0 +1,167 @@
...
+    expect(history.location.pathname).toBe('/artist/test-artist-id');
+    await waitFor(() => component.getByText(/#tag1/i).click());
+    expect(history.location.pathname).toBe('/tag/tag1');
+  });
+
+  it('should search for similar artist after clicking his name', async () => {
+    const { component, history, store } = mountComponent();
+    expect(history.location.pathname).toBe('/artist/test-artist-id');
+    await waitFor(() => component.getByText(/artist-similar-1/i).click());
+    const state = store.getState();
```
nukeop, line 47: Maybe also check that we went to `artist-similar-1`'s page by checking the current path in history?

packages/app/app/containers/ArtistViewContainer/ArtistViewContainer.test.tsx
```diff
@@ -0,0 +1,167 @@
...
+    expect(state.search.artistDetails).toEqual(expect.objectContaining({
+      loading: expect.objectContaining({loading: true})
```
nukeop, line 49: Not sure if this is right, we would expect `state.search.artistDetails` to match this:
```typescript
{ loading: { loading: true } }
```

PR #900, opened 2021-04-08
Enhance playlist view

HaiDang666:


packages/app/app/App.js
```diff
@@ -179,18 +178,10 @@ class App extends React.PureComponent {
               this.renderMenuCategory('collection', [
                 { name: 'favorite-albums', path: 'favorites/albums', icon: 'dot circle'},
                 { name: 'favorite-tracks', path: 'favorites/tracks', icon: 'music'},
-                { name: 'library', path: 'library', icon: 'file audio outline' }
+                { name: 'library', path: 'library', icon: 'file audio outline' },
+                { name: 'playlists', path: 'playlists', icon: 'list alternate outline'}
```
nukeop, line 182: Can we still go to individual playlists from this?

HaiDang666: I remove individual playlist from the menu, cos it doesn't make sense when users have many playlist and also when using with compact menu style.

if we keep individual playlist in menu should only show like top3 and implement set playlist order function

PR #902, opened 2021-04-10
Fix when title is not available by music-metadata

XutaxKamay:
Sometimes, common.title is not defined by music-metadata library.
So it ends up not being even inserted into the database at all, even though it is supported audio.

I feel like this is a bug from the library but for now, this workaround works.

### DIFF
--- a/packages/main/src/services/local-library/index.ts
+++ b/packages/main/src/services/local-library/index.ts
@@ -62,7 +62,12 @@ class LocalLibrary {
       path: file,
       folder,
       duration: format.duration,
-      name: common.title,
+      name: function() {
+        if (!_.isNil(common.title)) {
+          return common.title;
+        }
+        return path.basename(file, path.extname(file));
+      }(),
       position: common.track.no,
       album: common.album,
       artist: common.artist || 'unknown',


packages/main/src/services/local-library/index.ts
```diff
@@ -62,7 +62,12 @@ class LocalLibrary {
       path: file,
       folder,
       duration: format.duration,
-      name: common.title,
+      name: function() {
+        if (common.title !== undefined && common.title != null) {
```
nukeop, line 66: can you replace this condition with `_.isNil(common.title)`?

XutaxKamay: Of course I can but that would check only for `null`, not when the variable is just `undefined`,
when I replace the whole condition, it does this:
```
@nuclear/main:       main › QueryFailedError: SQLITE_CONSTRAINT: NOT NULL constraint failed: local_track.name
@nuclear/main:     at new QueryFailedError (webpack:///./node_modules/typeorm/error/QueryFailedError.js?:11:28)
@nuclear/main:     at Statement.handler (webpack:///./node_modules/typeorm/driver/sqlite/SqliteQueryRunner.js?:48:38)
```
Maybe you wanted `common.title !== undefined &&  _.isNil(common.title) ` ?

PR #909, opened 2021-04-18
Update README.md

scardracs:
Add gentoo


nukeop:
Could you link directly to the package?

scardracs:
@nukeop here you hare

PR #917, opened 2021-04-22
Added pop-up that contains volume controls when screen too small

schaudry98:
I utilized detection of window size to determine whether it was needed to minimize the bottom bar or not. If the window size was too small upon load, then it will replace the volume controls with a button that triggers a pop-up on hover and will contain the same volume controls. This can also be edited to dynamically resize as well. 

I had some confusion around exactly where the VolumePopUp.tsx file should be placed so that can be moved wherever matches the structure for the project. 

Bottom Bar Minimized

### DIFF
--- a/packages/ui/lib/components/PlayerBar/VolumePopUp.tsx
+++ b/packages/ui/lib/components/PlayerBar/VolumePopUp.tsx
@@ -1,0 +1,27 @@
+import React from 'react';
+
+import VolumeControls, { VolumeControlsProps } from '../VolumeControls';
+
+import { Button, Popup } from 'semantic-ui-react';
+
+export type VolumePopUpProps = VolumeControlsProps;
+
+const VolumePopUp: React.FC<VolumeControlsProps> = ({
+  volume,
+  updateVolume,
+  toggleMute,
+  isMuted,
+  playOptions
+}) => (
+  <Popup trigger={<Button>More</Button>} flowing hoverable basic>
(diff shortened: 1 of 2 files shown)


nukeop:
Pretty good functionality-wise, but I'm not so sure about the colors. Would you like to try redesigning it to match the rest of the player? I could also do it myself after merging your PR.

schaudry98:
If you could design the colors to match the video player, I would really appreciate it! Please let me know if you need anything else from me before merging the PR. This will be my first contribution to open source! :)

nukeop:
Ok, if you're fine with it I'll merge your pull request as it is, then edit the design myself.

Keeping it open for now so I don't forget about it.

schaudry98:
I got a little bit of time and tried inverting the colors on the pop-up and thought it matched more closely with the general theme. If this looks better, feel free to merge the request! 

Screen Shot 2021-04-24 at 3 25 33 PM

HaiDang666:
looking better

nukeop:
There is a problem with this feature - `window.innerWidth` is not getting updated in real time, so it won't apply the compact style to volume controls unless you reload the window. I'm going to add a resize listener to handle the changes.

nukeop:
image

I made it look like this to match other popups, and fixed the resizing issue.
I also replaced the "More" button with an ellipsis icon, and realigned the controls within the player bar.

Issue #841, opened 2021-02-25
On Autoradio mode, playing queue should not stop when it fails to load for next track.


haidang666:
<img width="1074" alt="Screen Shot 2021-02-25 at 14 38 38" src="https://user-images.githubusercontent.com/12780733/109119369-2cb7f900-7777-11eb-9ee1-880b85436f7f.png">

Sometime ytsr fails and then the playing queue stops. In UI, track's name still display with 0:00 duration.

Suggestion: when ytsr try re-do again or load another track.

Jghaddad123:
hey can I take a swing at this bug?

HaiDang666:
sure

HaiDang666:
closed with https://github.com/nukeop/nuclear/pull/938

Xunius:
Hi all, I saw this is closed but I'm seeing the same issue with nightly (downloaded last night).

What happens: I searched for a name, clicked it play, then autoradio added a new one to the queue.
But this stopped at the end of the 2nd song. The third one is in constant loading state (loading circle animation keeps on going forever). See screenshot:

![image](https://user-images.githubusercontent.com/8076931/152913100-ecc74d49-8039-49d3-a309-b9f00d04eaf1.png)

nukeop:
I changed the way loading streams works and the queue should never be stuck again.

PR #938, opened 2021-05-12
Auto radio bug

Jghaddad123:
Issue: https://github.com/nukeop/nuclear/issues/841

solution: when a song is undefined it will be removed from list causing no freeze at time 0:00 issue.
[nuclear.pdf](https://github.com/nukeop/nuclear/files/6463636/nuclear.pdf)


HaiDang666:
Your PR included the changes in `package-lock.json`. Try `reset --soft` to the beginning of your work and make a commit w/o package-lock file then push force.

packages/core/src/rest/Youtube.ts
```diff
@@ -115,19 +116,28 @@ export async function trackSearchByString(query: string, omitStreamId?: string,
...
+      duration: parseInt(topTrackInfo.videoDetails.lengthSeconds),
+      title: topTrackInfo.videoDetails.title,
+      thumbnail: topTrack.bestThumbnail.url,
+      format: formatInfo.container,
+      skipSegments: segments
+    };
+  }
+    catch (e){
+      logger.error(`Warning: topTrack.url is undefined, removing song`);
+      return undefined;
```
HaiDang666, line 138: I think I will better to throw an error instead of `return undefined`

HaiDang666 (changes requested):

packages/core/src/rest/Youtube.ts
nukeop: Yes, you should throw an error here, and then we can catch it in the thunk and dispatch an error action.

PR #939, opened 2021-05-12
add pull request template

HaiDang666:


nukeop:
I'll fix the situation with package-lock manually (probably more people will have the same problem). And the tests are automatically executed on PRs, so no need for a template yet.

PR #940, opened 2021-05-12
youtube live

HaiDang666:


packages/app/app/components/HLSPlayer/index.tsx
```diff
@@ -0,0 +1,171 @@
+import React, {RefObject, ReactElement} from 'react';
+import ReactHlsPlayer from 'react-hls-player';
+
+enum SoundStatus {
+  PAUSED = 'PAUSED',
+  PLAYING = 'PLAYING',
+  STOPPED = 'STOPPED',
+}
+
+export enum SoundErrors {
```
nukeop, line 10: These errors should be localized and kept in the locale files.

packages/app/app/containers/SoundContainer/index.js
```diff
@@ -167,11 +168,24 @@ class SoundContainer extends React.Component {
...
+    return /http.*?\.m3u8/g.test(url);
+  }
+
   render() {
     const { queue, player, equalizer, actions, enableSpectrum, currentStream, location, defaultEqualizer } = this.props;
     const currentTrack = queue.queueItems[queue.currentSong];
     const usedEqualizer = enableSpectrum ? equalizer : defaultEqualizer;
-    return Boolean(currentStream) && (
+    return Boolean(currentStream) && (this.isHlsStream(currentStream.stream) ? (
+      <HlsPlayer
```
nukeop, line 180: Will this work with the visualizer?

packages/core/src/rest/Youtube.ts
```diff
@@ -102,6 +102,26 @@ export function urlSearch(url) {
   }
 }
 
+export async function liveStreamSearch(query: string) {
+  const filters1 = await ytsr.getFilters(query);
```
nukeop, line 106: Can you refactor it like this?
```typescript
const videoFilter = (await ytsr.getFilters(query)).get('Type').get('Video');
const liveFilter = (await ytsr.getFilters(videoFilter.url)).get('Features').get('Live');
```

It's clearer what they do this way

nukeop (changes requested):

packages/app/app/components/HLSPlayer/index.tsx
```diff
@@ -0,0 +1,171 @@
...
+
+  toggleControls = () => {
+    this.playerRef.current.controls = !this.playerRef.current.controls;
+  }
+
+  setAudioMuted = (muted: boolean) => {
+    this.playerRef.current.muted = muted;
+  }
+
+  setAudioVolumn = (volume: number) => {
```
nukeop, line 73: change to `setAudioVolume`

packages/app/app/components/HLSPlayer/index.tsx
HaiDang666: I did the error same as `react-hifi` and also `handleError` doesn't display the error message
```js
  handleError(err) {
    logger.error(err.message);
    this.props.actions.streamFailed();
  }

```

do you want to `toast` the error message it will also work on `Sound` component of `react-hifi`

packages/app/app/containers/SoundContainer/index.js
HaiDang666: currently it doesn't.

packages/app/app/components/HLSPlayer/index.tsx
nukeop: Ok, I guess that's fine then

packages/app/app/components/HLSPlayer/index.tsx
```diff
@@ -0,0 +1,227 @@
...
+      this.setAudioMuted(muted);
+    }
+
+    if (volume && prevProps.volume !== volume) {
+      this.setAudioVolume(volume);
+    }
+  }
+
+  componentDidMount() {
+    // function fireOnVideoStart() {
```
nukeop, line 112: Can you remove the commented out code?

packages/app/app/containers/SoundContainer/index.js
nukeop: Ok, maybe in the future we could think about streaming the sound through the same pipeline as everything else

nukeop (approved):

PR #941, opened 2021-05-12
Auto radio bug

Jghaddad123:
[nuclear.pdf](https://github.com/nukeop/nuclear/files/6464890/nuclear.pdf)


packages/core/src/rest/Youtube.ts
```diff
@@ -134,7 +134,7 @@ export async function trackSearchByString(query: string, omitStreamId?: string,
     };
   }
     catch (e){
-      logger.error(`Warning: topTrack.url is undefined, removing song`);
+      throw new Error(`Warning: topTrack.url is undefined, removing song`);
       return undefined;
```
HaiDang666, line 138: You don't need `return` when u `throw` an error and log the original error is ok. Like this:
```js
    logger.error('youtube track search error');
    logger.error(e);
    throw new Error(`Warning: topTrack.url is undefined, removing song`);
```

HaiDang666 (changes requested):

packages/core/src/rest/Youtube.ts
Jghaddad123: your recommendations have been added

nukeop:
Thanks for fixing this, I tested it manually and it seems to work. I'll think of an automated test later when I have some time, but I won't delay merging any further.

PR #943, opened 2021-05-14
Export lyrics

jtmobs:


HaiDang666:
you install package in wrong directory, It should be in `packages/app/package.json`

jtmobs:
Done, Please check now

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "nuclear",
-  "version": "0.6.13",
```
HaiDang666, line 3: You have to check closely the changes on `package.json`. It changes many packages to lower version. It should only add the new packages that u want to use.

HaiDang666 (changes requested):

package.json
jtmobs: Yes it changed mine to "version": "0.6.11",
should I commit the .11 version or leave it as it is?
jspdf library is the needed package and its there, let me know what exactly you want me to do

nukeop: I don't get it, why did you lower the version number? This should stay unchanged

jtmobs: I had issues with the forked one and we forked that months ago, I made adjustments and everything is fine now, please review

jtmobs: I had issues with the forked one and we forked that months ago, I made adjustments and everything is fine now, please review

nukeop: It looks like those changes are still there, can you merge master into your branch?

PR #959, opened 2021-05-28
Educate people about the free software philosophy

CrazyPython:
This links to a nice video introduction on the free software philosophy, on a GNU MediaGoblin instance. We can hope this helps improve people's awareness of free software.

Without this clarification and link, people may assume "free software" means "software at no monetary charge."

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,4 +128,4 @@ This will be updated as the program evolves.
 
 ## License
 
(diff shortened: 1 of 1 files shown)


nukeop:
This text is copied from the gnu.org document about the GPL (https://www.gnu.org/licenses/gpl-3.0.en.html), I don't want to modify it.

PR #973, opened 2021-06-17
Enhance tray menus

HaiDang666:


packages/main/src/services/trayMenu/index.ts
```diff
@@ -1,40 +1,149 @@
...
+          this.window.send(IpcEvents.NEXT);
+        }
+      });
 
-          app.quit();
+      template.push({
+        label: 'Previous',
+        type: 'normal',
+        click: async () => {
+          this.window.send(IpcEvents.PREVIOUS);
```
HaiDang666, line 104: This one doesn't work as expected, it jumps from the top track to the bottom

packages/main/src/services/trayMenu/index.ts
```diff
@@ -1,40 +1,149 @@
...
+
+    // Control button
+    if (this.playerContext.track) {
+      if (this.playerContext.isPlaying) {
+        template.push({
+          label: 'Pause',
+          type: 'normal',
+          click: async () => {
+            this.window.send(IpcEvents.PAUSE);
+            this.update({isPlaying: false});
```
HaiDang666, line 78: I think `this.update({isPlaying: false});` is unnecessary but

packages/main/src/services/trayMenu/index.ts
nukeop: If you're at the first track, clicking previous should loop to the last track.

Issue #966, opened 2021-06-07
New nuclear feature: podcast feature


Architsakri:
using the apple podcast api, we can import podcasts for free an play them on nuclear

Ivan12273:
Hi, I'm new collaborating on OP projects, is anyone already working on this? I would like to help with this new feature :)

nukeop:
As far as I know, nobody picked it up yet. Would you like to be assigned? Not sure if we need any new components designed?

Ivan12273:
Yes, I would like it. I saw that the API requests are in core > rest. 

![image](https://user-images.githubusercontent.com/42753018/121740120-3b29e680-cac2-11eb-922c-8460f43a27f7.png)

A new item called "podcast" next to the live broadcast could be?

nukeop:
Sure, I think that's a good way to handle it. You are right that it should be placed in core.
Here's some more info on what should go where: https://app.gitbook.com/@nukeop/s/nuclear/developer-resources/architecture

Let me know if you have any questions.

Ivan12273:
Alright, thanks :)

Architsakri:
wait hey, since last.fm's api can pick up most played songs, artists, and all that. We should change the dashboard and add a recommendation engine so we can have recommended artists, most played songs and artists and all that.

nukeop:
There's autoradio already, but we don't keep a history of played tracks.

Architsakri:
I was thinking that if this had a mobile version it would really blow up. Is there any chance of a mobile version?

nukeop:
I tried developing on an Android VM but it's too slow and unreliable. Iphone version is not feasible of course.

Architsakri:
yeah I am in the process of developing with react native. So far I have made a login system, and a dashboard synced with last.fm just to get information about the current trend. Would you want to test the android apk?

Architsakri:
but leaving that aside, can we be expecting a feature where you can cast your audio to speakers right from the app?

nukeop:
Using chromecast?

Architsakri:
yeah

Ivan12273:
Hello, I did not have problems creating a new plugin for search podcast with the iTunes API.
But I had troubles doing the meta provider. I can receive all the podcast: 

![image](https://user-images.githubusercontent.com/42753018/122283257-fa5f1280-ceb1-11eb-8f17-fed0807e70d6.png)

[...]

PR #974, opened 2021-06-18
Podcast feature

Ivan12273:
Context: Adding a new feature for listen to podcast from iTunes API.

Hello! I am working on the issue [#966](https://github.com/nukeop/nuclear/issues/966) I added the next changes:

- **rest/iTunesPodcast.ts:** four functions for obtaining the information from the iTunes API.
- **stream/iTunesPodcastPlugin.ts:** search and getAlternateStream (I needed to do two request for each one, one for obtaining the id of the podcast and the other for the episode, the iTunes API has its limitations for searching)
- **meta/iTunespodcast.ts:** for obtaining the details of the podcast and the episodes. I did it in artist in order to test what I have, I was trying to do it in album but I noticed that the album tab works different that the artist tab.

I solved my problem I mentioned with the id.
I was thinking of creating a new podcast tab or re-use the album tab, but still not sure how the album components work, I would appreciate if someone can help guide me through the parts that I need to modify to add a new tab as album :)


nukeop:
> I was thinking of creating a new podcast tab or re-use the album tab, but still not sure how the album components work, I would appreciate if someone can help guide me through the parts that I need to modify to add a new tab as album :)

I think it would be good to just use the same album grid as albums use.

Search results are provided by this container: https://github.com/nukeop/nuclear/blob/master/packages/app/app/containers/SearchResultsContainer/index.js
And they're passed to this component: https://github.com/nukeop/nuclear/blob/master/packages/app/app/components/SearchResults/index.js

If you format the results correctly, you will be able to add an item under `panes()`, using `renderPane`.
If you're able, please also add a test for this here: https://github.com/nukeop/nuclear/tree/master/packages/app/app/containers/SearchResultsContainer

Ivan12273:
Hello, I made new changes in:

- /components/SearchResults/index.js
- /containers/SearchResultsContainer/index.js

And other files in order to add a new tab for podcast, I re-use the album grid and other album functions for display the podcasts.

Also I added a new abstract function in metaProvider.

image

image

It worked fine, are these changes correct?

packages/app/app/components/AlbumView/index.tsx
```diff
@@ -138,7 +138,7 @@ export const AlbumView: React.FC<AlbumViewProps> = ({
                   <th className={cx(styles.center, styles.position)}>
                     <Icon name='hashtag' />
                   </th>
-                  <th className={styles.left}>Song</th>
+                  <th className={styles.left}>Tracks</th>
```
HaiDang666, line 141: can you use locale in stead of this hard string

packages/core/src/plugins/meta/audius.ts
```diff
@@ -128,10 +129,13 @@ class AudiusMetaProvider extends MetaProvider {
     throw new Error('Method not implemented.');
   }
 
-
   async fetchAlbumDetailsByName(): Promise<AlbumDetails> {
     throw new Error('Method not implemented.');
   }
+
+  searchForPodcast(): Promise<SearchResultsPodcast[]> {
```
HaiDang666, line 136: @nukeop is it possible to make an interface for `MetaProvider` to implement default method to return empty value. So we don't need to do like this in very meta provider when we come ups with new method.

HaiDang666 (changes requested):

packages/app/app/components/AlbumView/index.tsx
Ivan12273: Sorry, what did you mean with locale?

Ivan12273: I changed it to how it was originally

nukeop: These need to be translated according to the selected language, using `react-i18next`.

packages/core/src/plugins/meta/audius.ts
nukeop: @Ivan12273 You can add this method to this abstract class with a default implementation: https://github.com/nukeop/nuclear/blob/6a99b3f85042f170de840d3e9b9381d8cd5fd84b/packages/core/src/plugins/metaProvider.ts

This way you can remove it from child providers.

nukeop:
I agree with @HaiDang666 's comments. Very good work too.

packages/app/app/components/AlbumView/index.tsx
HaiDang666: you can see in the same file that:
```nodejs
const { t } = useTranslation('album');
...
<PopupButton
    onClick={addAlbumToDownloads}
    ariaLabel={t('download')}
    icon='download'
    label={t('download')}
  />
```
and the `download` key is defined in `packages/i18n/src/locales/en.json` as:
```js
{
  ...
  album: {
     ...
     download: 'Download'
  }
}
```

Ivan12273: Hi! I changed it again using `react-i18next` I was only able to translate to Spanish, the others are in English :)

HaiDang666: That good 😌

packages/core/src/plugins/meta/audius.ts
Ivan12273: I added the method with the default implementation and removed it from the others classes. I needed to add` // eslint-disable-next-line @typescript-eslint/no-unused-vars` for the unused var :)

HaiDang666:
I found test snapshot fail and this
Screen Shot 2021-06-20 at 16 39 39

Ivan12273:
> I found test snapshot fail and this
> Screen Shot 2021-06-20 at 16 39 39
> 
> Screen Shot 2021-06-20 at 16 43 50

Hi, I corrected it, it was because I changed the text haha.

image

The first one is because of the file name, in my local project I have it different, I already changed it in the repo too :)

HaiDang666 (approved):
Good work, just wait for @nukeop.

nukeop:
Thanks, this is a really good contribution.

PR #1000, opened 2021-07-03
fix: Don't show empty tab in result after searching

### LINKED ISSUE #997: Don't show empty tab in result after searching
image

rbrishabh:
fixes #997
Hey @HaiDang666!
how does this look?
let me know if this needs any changes! Thanks

### DIFF
--- a/packages/app/app/components/SearchResults/index.js
+++ b/packages/app/app/components/SearchResults/index.js
@@ -1,6 +1,6 @@
 import React from 'react';
 import { Tab } from 'semantic-ui-react';
-import { withTranslation } from 'react-i18next';
+import { withTranslation, useTranslation } from 'react-i18next';
 import { Card } from '@nuclear/ui';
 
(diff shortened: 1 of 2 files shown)


nukeop:
If there are no search results in any tab, the screen will be empty. Can you add some sort of an empty state component in that case? I think we already have something like this (search for `EmptyState`).

nukeop:
Please also write a test for this

rbrishabh:
Hello @nukeop! I have added an empty state. I would be happy to write a test case! Is there any docs around how I can add a new test case? 
Thanks!

HaiDang666:
You can refer from this `packages/app/app/containers/DashboardContainer/DashboardContainer.test.tsx`

rbrishabh:
hi! I have a busy schedule and cant keep up with this now! Please feel free to continue it from here. I can close it for now

PR #1004, opened 2021-07-05
fix github link in HelpModal

HaiDang666:


nukeop:
Can you add a test for it?

HaiDang666:
Test when clicking on button, then check it opens new browser right?

nukeop:
Just mock out the electron module and check if it's called with the correct address

nukeop:
Awesome, great quality

PR #1013, opened 2021-07-11
Convert HelpModal to ts

HaiDang666:


packages/app/app/components/HelpModal/Contributors/index.tsx
```diff
@@ -16,6 +31,8 @@ const Contributors = ({
     return <div className={styles.contributors_error}>There was an error loading contributors from Github</div>;
   }
 
+  const handleGithubClick = (link: string) => link && electron.shell.openExternal(link);
```
nukeop, line 34: Correct me if I'm wrong, but this needs an argument, but will be called with an event argument on click.
This name is misleading, since it opens any link, not just Github links.

packages/app/app/containers/HelpModalContainer/index.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+const HelpModalContainer: React.FC<{}> = () => {
+
+  const githubContrib: ContributorProps = {
+    contributors: useSelector(githubContribSelectors.contributors),
+    loading: useSelector(githubContribSelectors.loading),
+    error: useSelector(githubContribSelectors.error)
+  };
+
+  return (<HelpModal githubContrib={githubContrib} />);
```
nukeop, line 16: It might be useful to move the callbacks opening links and usestate/usecallback calls here as well.
We generally want hook calls to happen in containers. The exception is `useTranslation`, which can be placed in components, unless they're in the `ui` package - those should take strings as props to show text.

nukeop:
I think it shows lowered coverage due to different formatting, and there's one more line than previously. I would not worry about it if it's just 0.01%. The check is there so we can immediately know if a pull request doesn't have tests for the test/bug it implements. In many cases it won't disqualify a PR, even if it fails.

PR #1024, opened 2021-07-19
Remove duplicate entries for Gentoo GURU

1over137:
We don't need two identical entries there for a single package. 
In addition, sad news: scardracs has resigned from Gentoo in general and the package is now unmaintained. It should still keep working in the foreseeable future with minimal changes, though, as long as there is no major change in the way this program is meant to be built.


HaiDang666:
@scardracs can you confirm ?

scardracs:
@HaiDang666 yes I can confirm

Issue #999, opened 2021-07-03
Add a button to create an empty one though on the playlist screen


HaiDang666:
as title

mezeru:
Can you elaborate , I could not understand wym exactly

HaiDang666:
Currently to create a playlist, we need to add tracks to `play queue` then `save as playlist`. This behavior is counterintuitive for many users so we should have a button on the playlist screen to create an empty playlist.

mezeru:
Ah ! I see , I will try to fix it

PR #1029, opened 2021-07-27
Create new playlist button on Playlists screen

ReshetovItsMe:
This PR addresses issue #999 to create playlist from playlists screen

It involves the following changes:

- Translation for new buttons
- PlaylistHeader added dialog and button to create playlist
- Playlist helper in core now can formatPlaylistForStorage with empty tracks
- PlaylistContainer hook pass createPlaylist action
- Updated snapshots

Looks like this:
new playlist


packages/app/app/components/Playlists/PlaylistsHeader/index.tsx
```diff
@@ -5,28 +5,52 @@ import { Button } from '@nuclear/ui';
...
 
-      <Button
-        basic
-        onClick={handleImportFromFile}
-        data-testid='import-from-file'
-      >
-        <Icon name='file text' />
-        {t('import-button')}
-      </Button>
+      <div>
```
nukeop, line 28: You can get rid of this div

packages/app/app/components/Playlists/index.tsx
```diff
@@ -23,9 +23,10 @@ const EmptyState = () => {
 type PlaylistsProps = {
   playlists: Playlist[];
   handleImportFromFile: React.MouseEventHandler;
+  createNew: (name: any) => void;
```
nukeop, line 26: Use a correct type for this (don't use `any`)

ReshetovItsMe: there are no type in action, but ok

packages/app/app/components/Playlists/PlaylistsHeader/index.tsx
ReshetovItsMe: no, I can't because header has flex with space between

packages/app/app/components/Playlists/index.tsx
ReshetovItsMe: fixed

PR #1031, opened 2021-07-31
Add copy orginal track url button in PlayQueue popup

HaiDang666:
It copy the track page url to clipboard (based on stream provider), not the stream url

image

### DIFF
--- a/packages/app/app/components/PlayQueue/QueuePopup/index.js
+++ b/packages/app/app/components/PlayQueue/QueuePopup/index.js
@@ -29,7 +29,7 @@ export const QueuePopup = ({
   copyToClipboard
 }) => {
   const triggerElement = useRef(null);
-
+  
   const handleOpen = useCallback(
     event => {
       event.preventDefault();
@@ -48,7 +48,7 @@ export const QueuePopup = ({
     const selectedStreamProvider = _.find(plugins.plugins.streamProviders, { sourceName: plugins.selected.streamProviders });
     actions.rerollTrack(selectedStreamProvider, selectedStream, track);
   };
-  
+
   const handleSelectStream = ({ track, stream }) => {
     actions.changeTrackStream(track, stream);
   };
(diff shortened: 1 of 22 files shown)


packages/ui/lib/components/StreamInfo/index.js
```diff
@@ -52,6 +53,11 @@ const StreamInfo = ({
         )}
       </div>
       <div className={styles.stream_buttons}>
+        {selectedStream.originalUrl && (
+          <a href='#' data-testid='copy-orignal-url' onClick={handleCopyStreamUrl}>
+            <Icon name='copy' />
```
nukeop, line 58: this icon suggests copying a file, better user `linkify`, and add a tooltip saying it copies the link to your clipboard

packages/app/app/components/PlayQueue/QueuePopup/index.js
```diff
@@ -49,6 +51,11 @@ export const QueuePopup = ({
     actions.changeTrackStream(track, stream);
   };
 
+  const handleCopyStreamUrl = useCallback(() => {
+    clipboard.writeText(selectedStream?.originalUrl?.length ? selectedStream.originalUrl : '');
```
nukeop, line 55: We should avoid writing anything to the clipboard if a stream is not available.

packages/app/app/components/PlayQueue/QueuePopup/index.js
```diff
@@ -49,6 +51,11 @@ export const QueuePopup = ({
     actions.changeTrackStream(track, stream);
   };
 
+  const handleCopyStreamUrl = useCallback(() => {
```
nukeop, line 54: This callback should be defined in a container, we want to avoid complex dependencies like Electron in components.

nukeop:
Trying to fix this

nukeop:
`selectedStream` is undefined, that's why popups don't open in the test environment.

nukeop:
There was no match for stream provider name with the selected track, so it did not open a popup.

nukeop:
Tests should be fine now, however I can't get the popup to show up in the snapshot (even though it exists in the tree).

https://testing-library.com/docs/react-testing-library/api/#baseelement-1
This option could probably fix that if we figure out how to use it here.

Issue #1050, opened 2021-09-11
Drop down menu on right of search bar gets hidden


Hardik-hi:
**Platform:** Windows 

**Nuclear version:** v0.6.16

**Description of the issue:**

![nuclear_ui_error](https://user-images.githubusercontent.com/55993168/132957005-010c6015-4dfa-4873-9569-cd233e6421b5.gif)

When searching in the search bar the drop down menu to the right gets hidden behind the search history pane. It is accessible when the user clicks outside the history pane.

**Possible solution** Decrease the width of the search history pane or maybe move the dropdown menu to the right.

Hardik-hi:
I can try working on this issue though.

Can I proceed? @nukeop @HaiDang666 ?

nukeop:
Thanks for reporting, sure, please take it. You just need to adjust the z-levels of these elements.

Hardik-hi:
@nukeop Gotcha!

PR #1052, opened 2021-09-12
Fix: Drop down menu on right side of search bar gets hidden

Hardik-hi:
Plugin dropdown on right of search bar now appears over the search drop down

#1050

Test result log (successful):
[overlay_output.txt](https://github.com/nukeop/nuclear/files/7149332/overlay_output.txt)


Hardik-hi:
@HaiDang666 Btw why is Codecove showing error in build?

nukeop:
Codecov is passing, since this doesn't change coverage.

PR #1053, opened 2021-09-19
Make discord rich presence toggleable

Memexurer:
makes discord rich presence toggleable


nukeop:
Thank you for the contribution. Could you please write a test for it?

HaiDang666:
You should remove language json files, only need to keep the english one.

PR #1078, opened 2021-10-18
Hide empty search tabs #997

### LINKED ISSUE #997: Don't show empty tab in result after searching

Erveon:
Resolves #997 

First time contributing to a project like this, noticed nothing had happened on the issue so decided to just give it a shot.
This simply hides a tab if it has no results for the category as requested in the issue and its comments.

Using the lodash get for this is inspired by similar functionality in the [SearchResults/AllResults render](https://github.com/nukeop/nuclear/blob/master/packages/app/app/components/SearchResults/AllResults/index.js#L107).

Screenshots:
image
image

From testing I've noticed that if a tab is active and a new search query is made that does not include the selected tab it'll show the empty state without an 'active' tab. I'm not sure exactly what the expected behavior would/should be if this is not desired.

image

### DIFF
--- a/packages/app/app/containers/SearchResultsContainer/SearchResultsContainer.test.tsx
+++ b/packages/app/app/containers/SearchResultsContainer/SearchResultsContainer.test.tsx
@@ -1,0 +1,60 @@
+import React from 'react';
+import { render } from '@testing-library/react';
+import { createMemoryHistory } from 'history';
+
+import { buildStoreState } from '../../../test/storeBuilders';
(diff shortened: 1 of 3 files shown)


nukeop:
Hi, thanks for the contribution. I proposed a small improvement.
Are you able to write a test for this feature?

HaiDang666:
Looking good, I gonna start the CI

nukeop:
Excellent code, thank you for contributing.

PR #1102, opened 2021-11-18
Artist Info button implemented 

jreiser1:
Could resolve https://github.com/nukeop/nuclear/issues/820. The URL opened will depend on what metadata provider the user selects near the search bar.


jreiser1:
Sorry again for the crazy volume of commits, we've had some trouble with our normal master branch and had to rebase some stuff.

packages/app/app/containers/ArtistViewContainer/hooks.ts
```diff
@@ -47,13 +48,41 @@ export const useArtistViewProps = () => {
     dispatch(FavoritesActions.removeFavoriteArtist(artist));
   }, [artist, dispatch]);
 
+  const handleArtistInfoClick = () => {
```
nukeop, line 51: I think it would be better if getting this link was implemented in the metadata provider, this way we could get it depending on the currently selected provider.

Could you also please add a test for this functionality so it's covered?

nukeop:
Thanks, I left a comment regarding implementation details.

Regarding the number of commits, you should start out with the latest master branch commit from this repo. I'm going to squash them anyway, but it looks like there are some unrelated files there (e.g. related to Travis, a CI that we haven't been using for years). The diff looks ok though so in the end it doesn't make a difference.

packages/app/app/containers/ArtistViewContainer/hooks.ts
jreiser1: I'll work on this tonight. Sorry for the delay, the holiday week was more hectic than I imagined. Could you point me towards the right place for the metadata provider? I want to make sure I'm putting stuff in the right place.

nukeop: For example here: https://github.com/nukeop/nuclear/blob/master/packages/core/src/plugins/meta/bandcamp.ts

You should add a method for getting an artist link to the MetaProvider abstract class (https://github.com/nukeop/nuclear/blob/master/packages/core/src/plugins/metaProvider.ts), and implement it in the providers that extend it, or just return null otherwise.

jreiser1: image
I noticed that the metadata plugins are already returning a resourceUrl as part of the SearchResultsArtist enum, which is used to render the artist search results (if I'm understanding the code correctly). Do you think it would make more sense to pull the URL from there instead of writing a separate function to grab the URL once more?

jreiser1: I've added a test for the button similar to the tests for clicking on similar artists or artist tags

PR #1105, opened 2021-11-20
Equalizer and FavouriteAlbumview component to typescript

Yahyakiani:
Converting Components to typescript

### DIFF
--- a/packages/app/app/components/EqualizerPresetList/index.tsx
+++ b/packages/app/app/components/EqualizerPresetList/index.tsx
@@ -1,12 +1,17 @@
 import React from 'react';
-import PropTypes from 'prop-types';
 import { List, Icon } from 'semantic-ui-react';
 import classNames from 'classnames';
 import { useTranslation } from 'react-i18next';
 
 import styles from './styles.scss';
 
-const EqualizerPresetList = ({ presets, onClickItem, selected }) => {
+type EqualizerPresetListProps ={
+  presets: Array<any>;
+  onClickItem: React.MouseEventHandler;
+  selected: string;
+}
+
+const EqualizerPresetList:React.FC<EqualizerPresetListProps> = ({ presets, onClickItem, selected }) => {
   const { t } = useTranslation('equalizer');
   return (
     <div className={styles.preset_list_container}>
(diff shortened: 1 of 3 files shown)


packages/app/app/components/EqualizerPresetList/index.tsx
```diff
@@ -1,12 +1,17 @@
...
-import PropTypes from 'prop-types';
 import { List, Icon } from 'semantic-ui-react';
 import classNames from 'classnames';
 import { useTranslation } from 'react-i18next';
 
 import styles from './styles.scss';
 
-const EqualizerPresetList = ({ presets, onClickItem, selected }) => {
+type EqualizerPresetListProps ={
+  presets: Array<any>;
```
nukeop, line 9: Please don't use any

Yahyakiani: Removed the use of Any from both of the components

packages/app/app/components/FavoriteAlbumsView/index.tsx
```diff
@@ -10,6 +9,31 @@ import Header from '../Header';
...
+  artists: { name:string } [];
+  images: Array<string>;
+  genres: Array<string>;
+  tracklist: Array<Tracklist>
+}
+
+type FavoriteAlbumsViewProps = {
+  albums: Array<Album>;
+  removeFavoriteAlbum: React.MouseEventHandler;
+  albumInfoSearch: (albumId: any, releaseType: any, release: any) => Promise<void>;
```
nukeop, line 33: albumId and release should be strings instead of any, releaseType is a union of string literals (`'master' | 'release'`)

Yahyakiani: Got it. In hindsight, I used this because of another component `ArtistView/index.tsx` and I believe that must be fixed as well along with this. I will make the consequent changes in both. 

```
type ArtistViewProps = {
  artist: Artist & {
    loading?: boolean;
  };
  isFavorite: boolean;
  addTrackToQueue: (item: any) => Promise<void>;
  artistInfoSearchByName: (artistName: any) => Promise<void>;
  albumInfoSearch: (albumId: any, releaseType: any, release: any) => Promise<void>;

```

Issue #1095, opened 2021-11-15
"Display Track Duration Over the Seekbar" setting doesn't work.


JFMugen:
**Platform:*Windows*

**Nuclear version:*v0.6.17*

**Description of the issue:*"Display Track Duration Over the Seekbar" setting doesn't work. Duration is always there.*

![image](https://user-images.githubusercontent.com/61559703/141806888-4ce1242d-b635-4e5c-acd5-6c845ef596b0.png)

nukeop:
2 things are required to fix this:
- Write a test
- In https://github.com/nukeop/nuclear/blob/6a99b3f85042f170de840d3e9b9381d8cd5fd84b/packages/app/app/containers/PlayerBarContainer/index.tsx, the `renderTrackDuration` needs to take its value from the appropriate setting (`trackDuration`) instead of being always true

Yahyakiani:
@nukeop Could I have this issue ? 1 more thing though is there any reference test component that I could use as reference on what kind of tests to write for this container ?

Yahyakiani:
@nukeop  You can close this issue I believe.Its been merged

PR #1110, opened 2021-11-25
Display track bug

Yahyakiani:
Resolves issue  #1095 . Now time track will show according to the renderTrack value. Wrote a test for PlayerBarContainer as well. Although it seems a couple more can be written in that case.


nukeop:
Nice one, thanks for fixing this.

Looks like snapshots don't pass, I'll fix that manually. Next time, please remember to update snapshots by running `npm test -- -u` in the package that needs new snapshots (such as when the way elements are rendered changes).

Issue #1111, opened 2021-11-26
Can not change the elements to be visible when the window size is reduced in miniplayer mode.


gnsdyd12:
**Platform: win10

**Nuclear version:  0.6.17

**Description of the issue: Can not change the elements to be visible when the window size is reduced in miniplayer mode.

gnsdyd12:
I wrote what I said on Discord in the past on the issue. I'm going to start working on it. Please approve it.

gnsdyd12:
I send pull requests. please check it.

nukeop:
This issue is from an older version of Nuclear that has been replaced with a complete rewrite. None of the old issues apply to the new codebase.

The old code is preserved on the [legacy/electron](https://github.com/nukeop/nuclear/tree/legacy/electron) branch. If you have feedback about the new version, head to [Discussions](https://github.com/nukeop/nuclear/discussions).

PR #1113, opened 2021-12-01
This is a pull request related to Issue #1111.

gnsdyd12:
This is a pull request related to Issue #1111.

When I raised the window window, I modified the image to grow as it was cut off due to the media query.

When reducing the window size, the minimum window size was specified, and the mini player became too small to see the UI well, so the minimum window size was specified and improved.


gnsdyd12:
There's an error, so I'm fixing it again.

gnsdyd12:
Finished.

gnsdyd12:
I'd appreciate it if you could write any review.

nukeop:
Thanks, I launched it and tried a variety of sizes and songs. The only downsize is that the text gets pretty small if you make the window the smallest possible size, but overall I like this change a lot.

PR #1114, opened 2021-12-01
album cover component to typescript

Yahyakiani:
Converting album cover and its child components to typescript


nukeop:
Thanks, looks good.

PR #1117, opened 2021-12-05
converting equalizer to typescript

Yahyakiani:
Converted Equalizer component to typescript


packages/app/app/components/Equalizer/index.tsx
```diff
@@ -65,7 +66,17 @@ const useChart = ({values, onEqualizerChange, preAmp, spectrum}) => {
   return canvasRef;
 };
 
-const Equalizer = ({values, preAmp, enableSpectrum, spectrum, onEqualizerChange, onToggleSpectrum, onPreampChange}) => {
+type EqualizerProps = {
+  values: Array<number>;
+  preAmp: number;
+  enableSpectrum: boolean;
+  spectrum: Array<number>;
+  onToggleSpectrum: (event: FormEvent<HTMLInputElement>, data: CheckboxProps) => void;
```
nukeop, line 74: please use `RadioProps['onChange']`

nukeop: And similarly for those other functions, where possible. You can use the `ComponentProps<T>` type from React if the props are not exported. It looks nicer and we don't have to redefine the exact type where it's just passed on to the inner component.

Yahyakiani: @nukeop I can't seem to make ComponentProps<T> work which requires a definition of T at which point I am forced to define it. Could you suggest an example of it? Thanks

nukeop: The point is that you only need the component itself to have the type of its props, e.g. if you have a component called SomeComponent in scope you can access its props type using `React.ComponentProps<typeof SomeComponent>`

PR #1121, opened 2021-12-10
Unify context popup behavior (open on right click everywhere)

TomasHvisc:
Context menu now opens with right click. You can have multiple context menus open and it does not look that good. Context menu closes only when you left click, right click on the same line or you enter and leave context menu with mouse.


nukeop:
Could you write a test to ensure this works? Either a unit test for this particular component or in a container, wherever it appears.

TomasHvisc:
I have never written test. I was looking at existing ones and some tutorials, but still i can't replicate it.

nukeop:
You can check how to test this exact functionality here: https://github.com/nukeop/nuclear/blob/00c726048b60eefcacc74977f3f7e212a5b81337/packages/app/app/containers/PlayQueueContainer/PlayQueueContainer.test.tsx#L21

PR #1148, opened 2022-01-26
TS Migration(ui/lib/components): SmoothImage, StreamInfo, AlbumPreview, ToastContainer

Gauravsharma-20:
the following components have been migrated to Typescript:
SmoothImage - `https://github.com/nukeop/nuclear/tree/master/packages/ui/lib/components/SmoothImage`
StreamInfo - `https://github.com/nukeop/nuclear/tree/master/packages/ui/lib/components/StreamInfo`
AlbumPreview - `https://github.com/nukeop/nuclear/tree/master/packages/ui/lib/components/AlbumPreview`
ToastContainer - `https://github.com/nukeop/nuclear/tree/master/packages/ui/lib/components/ToastContainer`


nukeop:
Hi, thanks for contributing. Please don't delete `package-lock.json` or the CI won't run.

packages/ui/lib/components/AlbumPreview/index.tsx
```diff
@@ -1,32 +1,54 @@
...
-  trackButtons,
+type Artist = {
+  name: string;
+};
+
+type Image = {
+  '#text'?: string;
+};
+
+type AlbumProps = {
```
HaiDang666, line 20: Is it possible to bring this to `types.ts` so it can be used in other components?

packages/ui/lib/components/StreamInfo/index.tsx
```diff
@@ -0,0 +1,133 @@
...
+};
+
+type SkipSegments = {
+  category?: string,
+  endTime?: number,
+  startTime?: number
+}
+
+type SelectedStream = {
+  duration?: number,
```
HaiDang666, line 88: also this type

HaiDang666:
I think it is better to reuse type as possible instead of making a new one every time.

nukeop:
Definitely, types such as Track need to be imported because they're defined elsewhere already.

packages/ui/lib/components/AlbumPreview/index.tsx
Gauravsharma-20: Sure, will do.

Gauravsharma-20:
PR Comments resolved and `package-lock.json` has been restored.

packages/ui/lib/components/SmoothImage/index.tsx
```diff
@@ -1,10 +1,17 @@
...
 import styles from './styles.scss';
 
-const SmoothImage = ({
+type SmoothImageProps = {
+  src: string;
+  placeholder?: string;
+  alt?: string;
+}
+
+const artPlaceholder = '../../../resources/media/art_placeholder.png';
```
nukeop, line 11: It won't use the correct image like this - it needs to be imported for webpack to load it.

nukeop:
Looking pretty good now, there's one more change that broke a test, after that I think we'll be ready to merge. Thanks.

packages/ui/lib/components/SmoothImage/index.tsx
Gauravsharma-20: When we import it takes the type of `NodeModule` and src of img tag expects a `string` type. If we give src a string it will be able to import the image directly.

Gauravsharma-20:
> Looking pretty good now, there's one more change that broke a test, after that I think we'll be ready to merge. Thanks.

Test seems to fail for `AlbumGrid` component, we haven't made any changes to that. Still, looking into the issue.

HaiDang666:
> Test seems to fail for `AlbumGrid` component, we haven't made any changes to that. Still, looking into the issue.

Perhaps `AlbumGrid`  uses one of the components you changed. For example:
Screen Shot 2022-01-28 at 16 37 35

nukeop:
No, this is caused by the import I mentioned. Just cast it to string. It won't work like this, and that's why the snapshot has changed.

packages/ui/lib/components/StreamInfo/index.tsx
```diff
@@ -0,0 +1,116 @@
...
+type DefaultProps = {
+  dropdownOptions: DropdownProps[],
+  idLabel: string,
+  titleLabel: string,
+  copyTrackUrlLabel: string,
+  sourceLabel: string
+};
+
+type Handlers = {
+  handleRerollTrack: React.MouseEventHandler<HTMLAnchorElement>,
```
HaiDang666, line 108: The test failed because of incorrect props names. In `PlayQueue\QueuePop` the passed props is
```js
  onRerollTrack={handleRerollTrack}
  onSelectStream={handleSelectStream}
  onImageLoaded={handleImageLoaded}
  onCopyTrackUrl={handleCopyTrackUrl}
```

Gauravsharma-20: Resolved and test cases passed.

Gauravsharma-20:
@nukeop Is there anything else I should fix or is the PR ready to merge?

PR #1240, opened 2022-04-06
Feature: Import spotify playlist by url

mdrokz:
# Changes

1. All the playlist import logic is inside core package file `spotify.ts`, this code is borrowed from [extract-sptf-playlist](https://github.com/NuclearPlayer/extract-sptf-playlist)
2. I added a new Dialog component to handle importing playlist from url it uses a `webview` tag to execute code from `spotify.ts`
3. I also wrote tests for this functionality but i couldnt figure out how to access webview inside the test

image

image

Screenshot_312

image


mdrokz:
> Thanks for doing this. It looks good, but I need some more time to actually understand what it's doing. I'm going to merge some of my changes (loading last.fm favorites) soon, and this will be the next in queue. After merging I will take a look at testing this, at least partially, maybe mocking out some stuff.

I wrote a basic test to test the import functionality but even though I can get the webview element inside the test i cant call it's methods like `getTitle`

Issue #1241, opened 2022-04-07
When <Return> pressed in search bar, close 'last searches' and jump to search result page


DrTexx:
Hi there :)

Would really love if when enter was pressed in the search bar
- The history menu would close/disappear
- The app switched to the search results page by itself

Believe this would be a really great UX enhancement :100:

May potentially also boost the app's accessibility for people with impairments :)

Thanks!

DrTexx:
Note if that is the intended behavior, there's a possibility that this is just a bug on my machine.

If that is the case, here's the environment I'm working with:
- OS: Pop!_OS 20.04
- Tag: nightly
- Package type: AppImage

nukeop:
This was intended at first but I agree that the history bar should be hidden when a search is starting. I also want to add an 'X' button to clear the search input. It should go to search results as soon as the search is finished, maybe it should go there when it starts.

baduts:
Hey,
I would like to work on this.
I would also like to add better typing to the SearchBox.tsx component and phase out compose since it is not really needed.
Would you like two different pull requests, one for the feature change and one for the refactor?

Would also be nice with a decision on the navigation, otherwise I will keep it as is, navigating when search is finished.

nukeop:
Sure, that would be great. Moving away from HOC in favor of hooks is also appreciated, please go ahead. It would be best to do this in two separate pull requests.

PaulaBarszcz:
Hello @nukeop, 
Could this issue be closed, since it has 2 already merged PRs, 
or is there anything else that needs to be done :)

PR #1250, opened 2022-04-11
Searchbox HOC -> Hooks & Proptypes -> Typescript

baduts:
### Changes

Changed proptypes to typescript type.

Moved compose functions to hooks with minor refactor. 

Kept UI related state in component, assuming containers act more as a DAO to redux and other services.

Changed deprecated e.which to e.key

Changed the input to be controlled (using value) since we already have the required variables and it will make things like clearing or appending to the input easier. E.g. [#1241](https://github.com/nukeop/nuclear/issues/1241).

Changed storybook to typescript and updated snapshot

### UX improvement ideas from development / debugging

Search on input is a cool feature, but the debounce on the search is causing me some issues. On the installed version, when deleting input with backspace, multiple searches will fire and cause clutter in the search history, this worsens the longer the search entry is. 

**Proposed fix:**  Cancel previous debounced requests on input and increase debounce value to 750 ms.

When clicking on a search history entry it is not appended to the search bar, but will fire a search.

**Proposed fix:** Set the search history text to the input bar on click, to make it easier to reason with the search and edit to fit needs. 

This is my first contribution to an open source project, any constructive criticism is appreciated :)

### DIFF
--- a/packages/app/app/containers/SearchBoxContainer/__snapshots__/SearchBoxContainer.test.tsx.snap
+++ b/packages/app/app/containers/SearchBoxContainer/__snapshots__/SearchBoxContainer.test.tsx.snap
@@ -16,6 +16,7 @@ exports[`Search box container should render the search box 1`] = `
         class=""
         data-testid="search-input"
         placeholder="Search..."
+        value=""
       />
       <div
         aria-disabled="false"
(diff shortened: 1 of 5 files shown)


nukeop:
Excellent contribution, thank you. I'll wait for CI to pass and I'll merge it then.

> Search on input is a cool feature, but the debounce on the search is causing me some issues. On the installed version, when deleting input with backspace, multiple searches will fire and cause clutter in the search history, this worsens the longer the search entry is.

This is probably wrong, only searches that actually pass through the debounce mechanism and cause actual requests should be saved in history. I agree that the time limit should be increased.

>When clicking on a search history entry it is not appended to the search bar, but will fire a search.

I think I said this elsewhere, but this is an artifact of early days of Nuclear, when it was little more than a search box and a play queue. I think at this point it would be better to do it like you said, to redirect to search results page with the search terms in the url, then start the search there via useEffect.

PR #1252, opened 2022-04-12
Enhancement/1241

baduts:
### Changes
Hides the history tab when enter is pressed to search. Issue: #1241 

If we want to extend this behavior to happen on every search (Not sure if this was intended by the comment in the issue)
We could move this logic to the handleSearch function in the container.


nukeop:
Thanks. This should probably also happen when we search without pressing enter (when you finish typing and the debounce timer  runs out). And like you said, it would be nice to increase the timer as well.

baduts:
### Changes
Increased debounce timer by 50%

Removed min input from debounced search, since it would always fire on the minimum string length when deleting chars.

Changed focus to be removed on any search. (This includes clicking a history entry)

PR #1256, opened 2022-04-14
Redux typing: Downloads

baduts:
### Changes

Typed download reducer and actions

Changed actions to use typesafe-actions

Fixed types in download related components

Fixed a bug caused by broken types that would allow the user to add tracks already in the download list

Removed weird reexport of ContextPopupProps

### Notes 

**I stopped working on this branch because the PR has grown quite large, but i discovered the following small issues that I can look at if you want:**

If you try to add a download that is already in the list, it will not be added anymore, but the toast will still display with "song added to downloads" message.

The pause / resume button in downloads is a bit clunky and inconsistent these changes did not cause this.


packages/app/test/storeBuilders.ts
```diff
@@ -589,9 +589,7 @@ export const buildStoreState = () => {
             completion: 1,
             track: {
               uuid: '1',
-              artist: {
-                name: 'test artist 1'
-              },
+              artist: 'test artist 1',
```
nukeop, line 592: I think the format here should stay as it was, since artist can be either a string or an object

baduts: This is true for the Track type used commonly across the app. But before the track is saved to the store, it is mapped to the TrackItem type (See getTrackItem) which I have typed. The full type for the object stored in state is the new type Download.

nukeop: Ok, I just realized that since the tests pass this works either way.

Issue #1228, opened 2022-03-18
Create New Playlist on Enter


Drew-Daniels:
I noticed that when creating a new playlist, the form is not able to be submitted on Enter. I think adding this would be a nice feature so users wouldn't be forced to manually click the Save button rather than press Enter.

![image](https://user-images.githubusercontent.com/62859552/159094431-c61d0287-f02b-4d99-bf9e-706d80bad55d.png)

nukeop:
Good idea, thanks for reporting

huangworld:
take

rechavarria:
Hi! This feature is also missing on the macOS version and I would like to start working on this issue if possible.

nukeop:
Ok, I assigned you. This is platform independent so it will be missing everywhere.

nukeop:
Implemented

PR #1262, opened 2022-04-17
Created hanlder for creating a new playlist on enter and added testin…

rechavarria:
…g for the feature.

Created a solution to #1228 by implementing a new event handler for the new playlist input dialog. When the user presses enter while inside the input dialog, the playlist will be saved similar to how it works with clicking on the save button. I also created 3 new tests for the feature and they all pass along with the preexisting tests. The tests include creating default name playlist on enter, creating custom name playlist on enter, and not creating an empty named playlist on enter. 

Feel free to let me know if there are any questions or if clarifications are needed.


nukeop:
Thanks for implementing this, I'll merge it as soon as CI passes.

PR #1263, opened 2022-04-17
fix #1185

### LINKED ISSUE #1185: Livestream does not mute
**Platform:**
Linux: Manjaro linux
Kernel: 5.16.4-1-MANJARO
Electron: v16.0.8

**Nuclear version:** Nightly

**Description of the issue:** When I try to mute the livestream (0% volume) it keeps playing at low volume

I recorded an example video of the bug:

https://user-images.githubusercontent.com/63082743/152163218-61a0111d-f286-4728-91d2-191277971cd6.mp4

Gavin-Sun555:
Fixed Livestream does not stop playing sound at zero percent volume by making the sound muted if the volume is set to zero.


nukeop:
Can you remove the package.json and package-lock from this PR? And please create a test for this so we're sure it's not going to occur again.

Gavin-Sun555:
> Can you remove the package.json and package-lock from this PR? And please create a test for this so we're sure it's not going to occur again.

Thanks for your reply. How and where should I add the test?

PR #1282, opened 2022-05-18
fix #963

### LINKED ISSUE #963: Broken list of playlist when clicking on single track in play queue
this should display as same as clicking on queue popup

Screen Shot 2021-06-02 at 15 21 57

**Expected**

Screen Shot 2021-06-02 at 15 21 49

SergioMorales3Pillar:
Adjusting css to fix #963 Broken List of Playlists [issue](https://github.com/nukeop/nuclear/issues/963)


SergioMorales3Pillar:
@nukeop removed the left style and replaced it with the direction prop from the semantic ui, tested on the different pages, seems like it works, please let me know otherwise

PR #1289, opened 2022-06-07
Fix playing section titles bug

### LINKED ISSUE #1168: Don't attempt to play section titles
**Platform:** Linux (Arch, using flatpak install)

**Nuclear version:** 0.6.16 (according to `flatpak info org.js.nuclear.Nuclear`)

**Description of the issue:** 
Nuclear attempts to play section titles as if they were songs. For example:

1. Set the metadata provider to Discogs 
2. Search for "Requiem", then find and click on [this result](https://www.discogs.com/master/115159-Mozart-Anna-Tomowa-Sintow-Helga-M%C3%BCller-Molinari-Vinson-Cole-Paata-Burchuladze-Wiener-Singverein-Wie)
3. Hit play, the "tracks" `Requiem D-Moll KV 626`, `I. Introitus`, `II. Kyrie`, etc. will be included in the queue

The issue with this is that these are not real tracks, but rather just section titles. In the [Discogs page](https://www.discogs.com/master/115159-Mozart-Anna-Tomowa-Sintow-Helga-M%C3%BCller-Molinari-Vinson-Cole-Paata-Burchuladze-Wiener-Singverein-Wie) of this album they show up as bold with a darker background to show that. I have only tested this problem on with the Discogs provider, though it may also be present on others.

sewe2000:
Resolves #1168 at least in the discogs metadata provider.
I also created a test covering my changes.
I would be grateful if someone reviewed my code.

### DIFF
--- a/packages/core/src/plugins/meta/discogs.test.ts
+++ b/packages/core/src/plugins/meta/discogs.test.ts
@@ -1,0 +1,20 @@
+import fetchMock from 'fetch-mock';
+import fetch from 'node-fetch';
+
+describe('discogs album tracks test', () => {
+
+  beforeEach(() => {
+    fetchMock.reset();
+  });
+
+  it('search for albums', async () => {
+    
+    fetchMock.getOnce('https://api.discogs.com/masters/115159', { tracklist: []});
+    fetch('https://api.discogs.com/masters/115159').then(res => res.json()).then( json => {
+      for (const track of json.tracklist) {
+        expect(track.type_).toBe('track');
+      }
+    });
+  });
+
+});
--- a/packages/core/src/plugins/meta/discogs.ts
+++ b/packages/core/src/plugins/meta/discogs.ts
@@ -81,10 +81,14 @@ class DiscogsMetaProvider extends MetaProvider {
     const coverImage = this.getCoverImage(release);
     const tracklist: Track[] = [];
 
-    release.tracklist.forEach(track => {
+    release.tracklist.map(track => {
       if (track.sub_tracks) {
-        track.sub_tracks.forEach(subTrack => tracklist.push(this.discogsTrackToGeneric(subTrack, artist)));
-      } else {
+        track.sub_tracks.map(subTrack => {
+          if (subTrack.type_ === 'track') {
+            tracklist.push(this.discogsTrackToGeneric(subTrack, artist));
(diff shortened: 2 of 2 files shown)


packages/core/src/plugins/meta/discogs.test.ts
```diff
@@ -0,0 +1,19 @@
+import fetchMock from 'fetch-mock';
+
+describe('discogs album tracks test', () => {
+
+  beforeEach(() => {
+    fetchMock.reset();
+  });
+
+  it('search for albums', async () => {
```
nukeop, line 9: Sorry, this test won't do anything. You just set up a mock then run a loop on an empty array, which will not run any expects. The mock will not be consumed since there is nothing that could call it in this test.

packages/core/src/plugins/meta/discogs.ts
```diff
@@ -82,10 +82,19 @@ class DiscogsMetaProvider extends MetaProvider {
     const tracklist: Track[] = [];
 
     release.tracklist.forEach(track => {
+      let trackToBePushed : Track;
       if (track.sub_tracks) {
-        track.sub_tracks.forEach(subTrack => tracklist.push(this.discogsTrackToGeneric(subTrack, artist)));
+        track.sub_tracks.forEach(subTrack => {
```
nukeop, line 87: Don't push in a for each loop, just use map, that's what it's for.

packages/core/src/plugins/meta/discogs.ts
```diff
@@ -105,6 +114,11 @@ class DiscogsMetaProvider extends MetaProvider {
   }
 
   discogsTrackToGeneric(discogsTrack: DiscogsTrack, artist: string): Track {
+
+    if (discogsTrack.type_ !== 'track') {
```
nukeop, line 118: You can just filter out non-tracks instead of filling the list with dummy entries and then filtering them out later.

nukeop:
I left some comments, but ultimately I think this is not the way to go about it. We'll probably need to leave those entries, and just render them differently, since we don't want to leave them out altogether, we just need to stop adding them to the play queue. They need to still be visible in the album view, maybe rendered differently.

sewe2000:
Yeah, I'll try to fix this PR in some time.

PR #1305, opened 2022-08-08
Fix queue action type ( IPC )

thibaultmthh:


thibaultmthh:
Forgot to provide info, 

I don't know why but IPC was using old import for QUEUE events

nukeop:
Thanks for fixing this

Issue #1306, opened 2022-08-09
Add listener for media play pause


maxkraft7:
The key MediaPlayPause (https://www.toptal.com/developers/keycode/for/media-play-pause) could be utilized. It gets aleady used by most webbrowsers if you have streams running.

nukeop:
I wanted to add it some time ago, but it doesn't work with https://www.npmjs.com/package/mousetrap which we use for keyboard shortcuts. I would be happy to merge this if somebody implements it.

AnthonyMarin:
Turns out Mousetrap allows you to add keycodes. I just found the keycodes for the major media buttons added them using mousetrap.
![mtkc](https://user-images.githubusercontent.com/90010906/191602662-c7e5cfe1-08e0-4d52-a18c-617e34ee56a8.JPG)

nukeop:
Implemented

PR #1322, opened 2022-09-21
bound media shortcuts (play/pause,next/prev)

AnthonyMarin:
In response to feature request : [Add listener for media play pause #1306](https://github.com/nukeop/nuclear/issues/1306)
Added Javascript keycodes  (https://www.toptal.com/developers/keycode) for media interaction buttons. Bound new keycodes to their appropriate commands.
mtkc

keycodes


nukeop:
I tried doing it this way but couldn't figure out which keycodes, these would be, thanks for adding this.

PR #1323, opened 2022-09-25
Toggle Tray Icon

aeum1016:
Added a setting to toggle the tray menu on startup. Requires rebooting for right now. Currently, if t he setting is toggled off, the traymenu simply isn't initialized. To deal with the issues that creates, updates don't work if the traymenu is undefined.


nukeop:
Looks ok to me

Issue #1182, opened 2022-02-02
Add shuffle all for favorite tracks


Rednas83:
Preferably by artist/genre/all

Something like for example groove. Which also allows sorting and filtering
![image](https://user-images.githubusercontent.com/25219229/152131981-78754198-f69a-44b2-87df-dc75153dc443.png)

nukeop:
Do you mean just adding them all to the queue with shuffle enabled?

Rednas83:
Yes, that would be awesome! A 1 click way for playing shuffled music depending on a genre filter. Play all by double clicking an album/Artist would also make a lot of Sense I think.

aaron-liao-604:
hey @nukeop it's my first time contributing and I've read through the contributing guidelines. I'm wondering if this is something I can pick up?

nukeop:
Sure, I'll assign you. Just let me know how you're planning to approach this.

nukeop:
This issue is from an older version of Nuclear that has been replaced with a complete rewrite. None of the old issues apply to the new codebase.

The old code is preserved on the [legacy/electron](https://github.com/nukeop/nuclear/tree/legacy/electron) branch. If you have feedback about the new version, head to [Discussions](https://github.com/nukeop/nuclear/discussions).

PR #1327, opened 2022-10-01
Add shuffle play to favorite tracks

aaron-liao-604:
#1182 

### What this does
- adds a shuffle play button to the Favorite tracks container
- it adds all the songs to the queue and sets the setting to shuffle

https://user-images.githubusercontent.com/65666550/193438110-6d4360fa-626c-41ce-83ca-6a9c8305bb41.mov


packages/app/app/components/FavoriteTracksView/index.tsx
```diff
@@ -34,6 +36,13 @@ const FavoriteTracksView: React.FC<FavoriteTracksViewProps> = ({
   removeFavoriteTrack
 }) => {
   const { t } = useTranslation('favorites');
+  const dispatch = useDispatch();
+
+  const addToQueue = useCallback(( tracks: Track[]) => {
```
nukeop, line 41: I think this should instead add all tracks to the queue, and enable shuffle, instead of shuffling the order when adding

aaron-liao-604: wow that was a fast response haha. oh I didn't know you can enable shuffle mode while in your queue, will look into this. 

In terms of adding the button, are there any guidelines to the UI?

nukeop: Not really, you can take a look at other existing buttons. And please remember to create tests.

packages/app/app/components/FavoriteTracksView/index.tsx
```diff
@@ -34,6 +41,25 @@ const FavoriteTracksView: React.FC<FavoriteTracksViewProps> = ({
...
 }) => {
   const { t } = useTranslation('favorites');
+  const dispatch = useDispatch();
+  const settings = useSelector(settingsSelector);
+
+  const toggleOption = useCallback(
+    (option, state) => dispatch(settingsActions.toggleOption(option, state)), [dispatch]
+  );
+
+  const toggleShuffle = useToggleOptionCallback(toggleOption, 'shuffleQueue', settings);
```
aaron-liao-604, line 51: Is this the correct way to use `useToggleOptionCallback`? I tried to extract the logic to specifically only toggle shuffle queue but it didn't seem to work 🤔

aaron-liao-604:
@nukeop this is ready for review

packages/app/app/components/FavoriteTracksView/FavoriteTracksView.test.tsx
```diff
@@ -0,0 +1,81 @@
+
```
nukeop, line 1: Can you put those tests here instead? https://github.com/nukeop/nuclear/blob/master/packages/app/app/containers/FavoritesContainer/FavoritesContainer.tracks.test.tsx

Testing the container will give us more reliable coverage than testing the presentational component

packages/app/app/components/FavoriteTracksView/index.tsx
nukeop: That looks good to me

nukeop:
Hi, I ran the code to rule out some obscure CI failure, and it turns out that the tests are failing with a promise rejection:
```
"TypeError: Cannot read properties of undefined (reading 'sourceName')".] {
  code: 'ERR_UNHANDLED_REJECTION'
}
```

aaron-liao-604:
> Hi, I ran the code to rule out some obscure CI failure, and it turns out that the tests are failing with a promise rejection:
> 
> ```
> "TypeError: Cannot read properties of undefined (reading 'sourceName')".] {
>   code: 'ERR_UNHANDLED_REJECTION'
> }
> ```

looks like some async await issue, I've tried adding an await for adding to the the tracks to the queue. Perhaps that would solve it?

PR #1343, opened 2022-10-23
Update README-de.md

PythonPizzaDE:
I made the German translation of the readme more "semantic German". I am from Germany and changed some sentences to sound better.


nukeop:
Thanks!

PR #1347, opened 2022-10-30
Update sqlite3 package to fix npm install error on macOS with Apple Silicon

sryze:
Update the version of sqlite3 from 5.0.11 to 5.1.2 where they added ARM64 binaries for Darwin so that this error doesn't occur:

```
> sqlite3@5.0.11 install /Users/username/Sources/nuclear/node_modules/sqlite3
> node-pre-gyp install --fallback-to-build

node-pre-gyp ERR! install response status 404 Not Found on https://github.com/TryGhost/node-sqlite3/releases/download/v5.0.11/napi-v6-darwin-unknown-arm64.tar.gz
node-pre-gyp WARN Pre-built binaries not installable for sqlite3@5.0.11 and node@14.20.1 (node-v83 ABI, unknown) (falling back to source compile with node-gyp)
node-pre-gyp WARN Hit error response status 404 Not Found on https://github.com/TryGhost/node-sqlite3/releases/download/v5.0.11/napi-v6-darwin-unknown-arm64.tar.gz
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -156,7 +156,7 @@
 				"search-azlyrics": "0.0.3",
 				"semantic-ui-css": "^2.4.1",
 				"semantic-ui-react": "^2.0.4",
-				"sqlite3": "^5.0.3",
+				"sqlite3": "^5.1.2",
 				"stream-filter": "^2.1.0",
 				"stream-reduce": "^1.0.3",
 				"style-loader": "^2.0.0",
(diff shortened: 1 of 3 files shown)


nukeop:
Excellent, thanks for letting me know this problem occurs. Once CI finishes running I'll merge it and build a release.

sryze:
To be fair, it doesn't seem to be a fatal error as NPM was able to go past it by building sqlite3 from source. But for me on subsequent runs of `npm install` the source build didn't work for some reason and it just started failing... I haven't figured out why (I guess it could be related to NPM caching).

nukeop:
>   ⨯ Exit code: ENOENT. spawn /usr/bin/python ENOENT  stackTrace=
                                                       Error: Exit code: ENOENT. spawn /usr/bin/python ENOENT

I'm wondering if this means that Python is missing from the CI machine?

sryze:
> > ⨯ Exit code: ENOENT. spawn /usr/bin/python ENOENT  stackTrace=
> > Error: Exit code: ENOENT. spawn /usr/bin/python ENOENT
> 
> I'm wondering if this means that Python is missing from the CI machine?

Not necessarily, because Python may be installed at a path different from `/usr/bin/python`. I think this is where it is usually installed on Linux, but on a Mac it is usually located under `/Library/Frameworks/...`, e.g. `/Library/Frameworks/Python.framework/Versions/2.7/bin/python`.

sryze:
It looks like they have fixed it in a newer release:
https://github.com/electron-userland/electron-builder/issues/6873

nukeop:
Could you try with an updated electron-builder?

sryze:
Unfortunately there appears to be a bug that causes nuclear's binary package to be compiled for x64 instead of arm64 on M1, so I couldn't verify the build on that machine. However, on my x64 machine running macOS Catalina it builds successfully with the latest electron-builder (26.3.0).

It seems that the build started failing because /usr/bin/python was removed in macOS Monterey 12.3: https://developer.apple.com/documentation/macos-release-notes/macos-12_3-release-notes

nukeop:
I think this will pass if we change the CI image to an older version. I noticed the Mac build fails on other branches too now. So I'm going to merge this and fix it on master. Thanks for contributing this.

PR #1355, opened 2022-11-03
Audio normalization added

SimonKacir:
Audio normalization is implemented, currently for song you manually select, songs added manually through queue seem to not refresh the normalizer component, therefore they don't get normalized.


nukeop:
Thanks, I'll get your branch and try to fix any outstanding issues with the tests, then put it behind a toggle in the settings.

nukeop:
I started analyzing the algorithm and found out that this is actually taken from this repo: https://github.com/est31/js-audio-normalizer/blob/master/normalizer/normalizer.html

Which is distributed under Apache and MIT licenses, both of which require attribution (so it needs to be added).

On another note, I'm wondering if it could be possible to operate on the thus far loaded portion of the stream, instead of preemptively loading it all (ruining the whole point of streaming). It's very likely that the calculated gain would not change significantly over the course of the track, and that it would approach the target value for the whole track asymptotically. So if we can figure out how to retrieve the part of the stream that's already been loaded, we can simply recalculate the gain on each loaded chunk (which is not computationally expensive and can happen asynchronously anyway).

nukeop:
Maybe an easier method of normalizing would be using a ConvolverNode? https://developer.mozilla.org/en-US/docs/Web/API/ConvolverNode/normalize

This looks like it operates on whatever is being streamed through it so there's no need to interrupt playback to load it all at once.

SimonKacir:
Just tried out convolver node, it does normalize, but at the same time ruins every single song by overlaying reverb on it. So it doesn't seem like something we can do.

nukeop:
I got it working and created a setting for enabling it (off by default).
Fortunately there was only a small change required (`updateNode` receives props as the second argument). There is an audible difference - I think I will start using it myself as soon as it's released.

PR #1365, opened 2022-11-10
WIP: Chore: allow favorite track list sorting

### LINKED ISSUE #722: [Feature Request] Organize/Sort songs in Favorite Tracks Page

mashazyu:
@nukeop , first of all, thank you very much for the quick reply! 

Closes #722 

**Changes**
- Allowed reordering of tracks in TrackTableContainer component
- Updated snapshot

**Open question**
- Is that a move in the right direction?
- I need a tip/advise on how to test this change :) So if you could point me to similar test/docs in react testing library, I'd appreciate it. I did find this article, but I doubt if it's a good approach in this case (https://testing-library.com/docs/example-drag/)


packages/app/app/actions/favorites.ts
```diff
@@ -61,6 +62,19 @@ export function removeFavoriteTrack(track) {
   };
 }
 
+export function reorderFavoriteTrack(fromIndex, toIndex) {
```
nukeop, line 65: To be honest, when I read the feature request, I thought sorting would only change the order the tracks are displayed in, and I didn't think permanently saving them in this state would be necessary. Do you think it would be better to do it this way?

PR #1366, opened 2022-11-12
Look for the "verified" badge when looking for streams on youtube

rutviklhase:
- Added a new value called "verifiedChannelScore" inside the "heuristics.ts" file. 
- Added the newly created value to the "mean( )" function.

The verified channels now have  a significant weight while deciding what tracks to recommend on top of the list.

### DIFF
--- a/packages/core/src/rest/heuristics.ts
+++ b/packages/core/src/rest/heuristics.ts
@@ -67,6 +67,7 @@ export class YoutubeHeuristics implements SearchHeuristics<Partial<Video>> {
     }
 
     const channelNameScore = track.author.name.toLowerCase().includes(artist.toLowerCase()) ? 200 : 0;
+    const verifiedChannelScore = track.author.verified ? 200 : 0;
 
     return mean([
       titleScore,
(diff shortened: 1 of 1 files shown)


rutviklhase:
@nukeop  Please let me know if there's anything missing! This is my first time contributing. I've also excluded the package-lock.json file.

nukeop:
Thanks, the code looks ok. Could you please add a test for this case? There is a bunch of similar tests already so it should be pretty easy. The setup should probably include a couple of videos with identical titles and authors, while one of them is marked as verified.

rutviklhase:
> Thanks, the code looks ok. Could you please add a test for this case? There is a bunch of similar tests already so it should be pretty easy. The setup should probably include a couple of videos with identical titles and authors, while one of them is marked as verified.

@nukeop  Yes absolutely! Thank you so much for the feedback.

rutviklhase:
Hey @nukeop !
I actually needed just a little bit of help with setting up the test. 
I dont think I completely understand where exactly heuristics actually make an impact. From what I can understand, the changes in heuritics have an imapact on the order in which the Livestreams appear, am I correct? 

For instance : 
Here's a search result with `const verifiedChannelScore = track.author.verified ? 200 : 0;`.
image

And here's a search result with  `const verifiedChannelScore = track.author.verified ? 0 : 0;`.
image


Is this the desired effect we're looking to have?

rutviklhase:
@nukeop Hey! Could you please take a look at the screenshots I've posted above? Would love to know what you think!

PR #1370, opened 2022-11-18
Fix: do not play section titles

### LINKED ISSUE #1168: Don't attempt to play section titles

mariadoggo:
Closes #1168 

**Changes**
- Disable playing/adding to playlist/etc of the tracks of type "heading"
- Fix a bug that some albums cannot be played with "Play now" button. I fixed this bug in the current PR, because it was very obvious during the testing of the implemented feature. But maybe another ticket could be created to make sure that artist prop has the same structure regardless from where the track comes?

**Tests**
- I updated existing TrackTable story & added snapshot test for tracklist that has headings.

**Open questions**
- I am not sure if I used the best approach. So if you have a tip for me on how proposed solution can be improved, I'd be stoked.
- Are there some other tests that you would recommend to add?

**Screenshots**
image

### DIFF
--- a/packages/app/app/components/AlbumView/index.tsx
+++ b/packages/app/app/components/AlbumView/index.tsx
@@ -10,7 +10,7 @@ import TrackTableContainer from '../../containers/TrackTableContainer';
 import { AlbumDetailsState } from '../../reducers/search';
 
 type AlbumViewProps = {
-  album?: AlbumDetailsState 
+  album?: AlbumDetailsState
   isFavorite: boolean;
   searchAlbumArtist: React.MouseEventHandler;
   addAlbumToDownloads: React.MouseEventHandler;
@@ -88,7 +88,7 @@ export const AlbumView: React.FC<AlbumViewProps> = ({
                 }
                 <div className={styles.album_tracks}>
                   <label>{t('tracks')}</label>
-                  {album.tracklist.length}
+                  {album.tracklist.filter(track => track.type !== 'heading').length}
                 </div>
                 <div className={styles.album_buttons}>
                   <a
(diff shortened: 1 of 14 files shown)


nukeop:
Other than a snapshot test, it would be nice to have a test like this:
- Go to an album page that has e.g. 3 tracks and 2 headings
- Try to play all tracks
- Check the queue: it should only contain the 3 tracks

In general snapshot tests are nice for easy and low effort testing of visual elements, but functionality like this is best tested in the form of integration tests in the `app` package.

nukeop:
I tried solving the conflicts but I can't push to your branch - getting permission denied.

Would you like to try adding a test for this use case, or should I handle this? You'll need to give me a permission to push to your branch if so.

PR #1403, opened 2023-02-24
export playlist as png

yatish-a:
**What's New**
added an export playlist as png feature right below the export playlist as JSON button.


nukeop:
Can you ensure that CI passes?

PR #1412, opened 2023-03-02
chore: update best new music to typescript

thisprojects:
Migrated the following components to Typescript:

app -> components -> Dashboard -> BestNewMusicTab
- Best New Music Content
- Best New Music Menu
- Best New Music Card

Migrated container `BestNewMusicCardContainer` to typescript.

Removed some un-used properties from `DashboardContainer`


packages/app/app/components/Dashboard/BestNewMusicTab/BestNewMusicContent/index.tsx
```diff
@@ -30,7 +36,7 @@ const BestNewMusicContent = ({
           </div>
           <div
             className={styles.title}
-            onClick={() => albumInfoSearchByName(item.title + ' ' + item.artist, history)}
+            onClick={() => albumInfoSearchByName(item.title + ' ' + item.artist, item.artist,  history)}
```
thisprojects, line 39: Not sure that this line is correct 🤔 the signature for albumInfoSearchByName (in search.ts) is (title, artist, history) - should item.title be concatenated with item.artist (as the first argument) here?

packages/app/app/components/Dashboard/BestNewMusicTab/BestNewMusicMenu/BestNewMusicCard/index.tsx
```diff
@@ -32,20 +61,24 @@ const FavoriteIcon = ({ isFavorite, onClick, dataTestId }) =>
...
-  settings,
-  withFavoriteButton
+  withFavoriteButton,
+  favoriteTrack,
+  actions,
+  settings
 }) => {
   const { artist, title, thumbnail } = item;
 
+  const handleClick = () => {
```
thisprojects, line 74: Added this click handling function to satisfy the event handler error typescript would raise with the previous onClick implementation

packages/app/app/components/Dashboard/index.tsx
```diff
@@ -10,36 +10,24 @@ import GenresTab from './GenresTab';
...
 
 type DashboardProps = {
   dashboardData: DashboardReducerState;
-  streamProviders: StreamProviderPlugin[];
   isConnected: boolean;
   isPromotedArtistsEnabled: boolean;
 
   artistInfoSearchByName: (artistName: string) => void;
   albumInfoSearchByName: (albumName: string, artistName: string) => void;
-  addToQueue: (item: QueueItem) => void;
```
thisprojects, line 24: These properties don't seem to be used - safe to remove? 🤔

packages/app/app/containers/BestNewMusicCardContainer/index.tsx
```diff
@@ -29,19 +17,20 @@ function mapStateToProps(state, ownProps) {
...
-        {},
-        FavoritesActions,
-        ToastActions
-      ),
-      dispatch
-    )
-  };
+export interface BestNewMusicCardContainerProps extends ActionCreatorsMapObject {
+  FavoritesActions: ActionCreator<any>;
+  TostActions: ActionCreator<any>;
```
thisprojects, line 22: Redux + typescript = tricky 😆 Any advice on typing here would be appreciated.

packages/app/app/containers/DashboardContainer/index.tsx
```diff
@@ -33,10 +33,6 @@ const DashboardContainer: React.FC = () => {
   
   const artistInfoSearchByName = (artistName: string) => dispatch(SearchActions.artistInfoSearchByName(artistName, history));
   const albumInfoSearchByName = (albumName: string, artistName: string) => dispatch(SearchActions.albumInfoSearchByName(albumName, artistName, history));
-  const addToQueue = (item: QueueItem) => dispatch(QueueActions.addToQueue(item));
-  const selectSong = (song: number) => dispatch(QueueActions.selectSong(song));
-  const clearQueue = () => dispatch(QueueActions.clearQueue());
```
thisprojects, line 38: Again, these don't seem to be used.

packages/app/app/containers/BestNewMusicCardContainer/BestNewMusicCardContainer.test.tsx
```diff
@@ -0,0 +1,80 @@
+import { waitFor } from '@testing-library/react';
+import pitchforkBnm from 'pitchfork-bnm';
+import { Deezer } from '@nuclear/core/src/rest';
+
+import { buildStoreState } from '../../../test/storeBuilders';
+import { mountedComponentFactory, setupI18Next } from '../../../test/testUtils';
+
+jest.mock('pitchfork-bnm');
```
nukeop, line 8: To mock particular functions in this import, you can use `mockImplementation`.

nukeop: I think this will also help avoid the need to use @ts-ignore comments.

packages/app/app/containers/BestNewMusicCardContainer/BestNewMusicCardContainer.test.tsx
```diff
@@ -0,0 +1,80 @@
...
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    Deezer.mapDeezerTrackToInternal = jest.requireActual('@nuclear/core/src/rest').Deezer.mapDeezerTrackToInternal;
+  });
+
+  afterAll(() => {
+    jest.clearAllMocks();
+  });
+
+  it('should render card album title', async () => {
```
nukeop, line 37: These four are basically the same exact test. If the snapshot is rendered as you think it should be rendered, there's no need to split it into 4 separate tests. Either put all the expects in a single test, or simply use the snapshot and remove the expects, since the snapshot will check the same things.

packages/app/app/containers/BestNewMusicCardContainer/index.tsx
nukeop: Try `import * as FavoritesActions from '../../actions/favorites';` and then just `typeof FavoritesActions`.

nukeop: This is a lazy way of passing callbacks to components, btw. It's messy and too broad. I did it because I was lazy.

packages/app/app/containers/BestNewMusicCardContainer/index.tsx
```diff
@@ -29,19 +17,20 @@ function mapStateToProps(state, ownProps) {
...
 }
 
+const mapDispatchToProps = (dispatch) => ({actions: bindActionCreators<BestNewMusicCardContainerProps, any>(
+  Object.assign(
```
nukeop, line 26: It's better to use spread (`{...stuff}`) instead of `Object.assign` now. I didn't know it existed when I wrote this (or maybe it wasn't available yet).

nukeop (changes requested):

packages/app/app/components/Dashboard/BestNewMusicTab/BestNewMusicMenu/BestNewMusicCard/index.tsx
thisprojects: The problem with typescript and the previous implementation was that `setActiveItem`, when used as an onClick callback, would require the method (declared in BestNewMusicTab) to have a MouseEventHandler type as its first argument. `setActiveItem` is also called directly in its home component (BestNewMusicTab - index.tsx line 35), in which case typescript would not be happy about the method having the MouseHandlerEvent (its called from componentDidMount). 

I feel like the `setActiveItem` method shouldn't be concerned with click handlers in its child components, which is why I went for the approach of adding a local click handler to `bestNewMusicCard`. What are your thoughts? 🤔 

I agree about the div conventionality though 🤔 should we change this Div element to a button  / anchor anyway?

packages/app/app/containers/BestNewMusicCardContainer/index.tsx
thisprojects: Thanks 😄 replaced <any> with typeof for Favourites and Toast.

packages/app/app/components/Dashboard/BestNewMusicTab/BestNewMusicMenu/BestNewMusicCard/index.tsx
```diff
@@ -1,12 +1,41 @@
...
-import { favoriteTrackShape } from '../../../../../constants/propTypes';
 import styles from './styles.scss';
+import { PitchforkAlbum, PitchforkTrack } from '../../../../../actions/dashboard';
+import { Setting, Track } from '@nuclear/core';
+
+type BestNewMusicCardProps = {
+  item: PitchforkAlbum;
+  onClick: (activeItem: PitchforkAlbum | PitchforkTrack) => void;
+  withFavoriteButton?: boolean;
+  actions?: {
```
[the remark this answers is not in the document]
thisprojects: By using `typeof removeFavoriteTrack` etc? I've gone with that implementation in the update.

packages/app/app/containers/BestNewMusicCardContainer/BestNewMusicCardContainer.test.tsx
thisprojects: Have made the mockImplementation change - does it look right? Tests are happy.

packages/app/app/containers/BestNewMusicCardContainer/index.tsx
thisprojects: Have made this change

packages/app/app/containers/BestNewMusicCardContainer/BestNewMusicCardContainer.test.tsx
thisprojects: Have consolidated the four expects into one test.

packages/app/app/containers/BestNewMusicCardContainer/BestNewMusicCardContainer.test.tsx
nukeop: Perfect!

packages/app/app/components/Dashboard/BestNewMusicTab/BestNewMusicMenu/BestNewMusicCard/index.tsx
nukeop: I think this is an issue with the parent component and how it passes down this prop - since you're only converting JS to TS, let's keep it this way in this PR. It can potentially be fixed in a separate commit in the future.

PR #1414, opened 2023-03-03
fix: Listening history container locale date time test fix

thisprojects:
Fixes issue when tests are run in a locale which differs from expected. 

EG: When running tests on my machine in the UK, the date is formatted as  "01/01/2020". The test is expecting a date string of "1/1/2020". 

Any thoughts? 🤔

### DIFF
--- a/packages/app/app/containers/ListeningHistoryContainer/ListeningHistoryContainer.test.tsx
+++ b/packages/app/app/containers/ListeningHistoryContainer/ListeningHistoryContainer.test.tsx
@@ -19,12 +19,12 @@ jest.mock('electron', () => ({
         uuid: 'test',
         artist: 'test artist',
         title: 'test title',
-        createdAt: new Date('2020-01-01')
+        createdAt: new Date('2020-01-01, 12:00')
       }, {
         uuid: 'test2',
         artist: 'test artist2',
         title: 'test title 2',
-        createdAt: new Date('2020-01-02')
+        createdAt: new Date('2020-01-02, 12:00')
       }],
       cursor: {
         beforeCursor: null,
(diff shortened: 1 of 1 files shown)


packages/app/app/containers/ListeningHistoryContainer/ListeningHistoryContainer.test.tsx
```diff
@@ -40,15 +40,25 @@ describe('Listening history container', () => {
   it('renders the listening history', async () => {
     const { component } = mountComponent();
 
+    const makeLocal = (dateString, timeString?) => {
+      const date = new Date(dateString).toLocaleDateString();
```
nukeop, line 44: Can you do the same thing with `timeString` for consistency?

nukeop: Sorry, I meant just assigning the time to a const, like you did with date. I can't see the previous version because somehow there's still only one commit.

packages/app/app/containers/ListeningHistoryContainer/ListeningHistoryContainer.test.tsx
```diff
@@ -40,15 +40,25 @@ describe('Listening history container', () => {
...
+        return date;
+      }
+    };
+
     await Promise.all([
-      '1/1/2020',
+      makeLocal('2020-01-01'),
       'test title',
       'test artist',
-      '1/1/2020, 12:00:00 AM',
```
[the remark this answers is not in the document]
nukeop: I think so, if we use `toLocaleDateString` everywhere, it should be fine.

packages/app/app/containers/ListeningHistoryContainer/ListeningHistoryContainer.test.tsx
thisprojects: Ahhh ok, sure no problem will make that change. Sorry,  I squashed the previous commit down with a rebase.

PR #1415, opened 2023-03-06
Fixes #1409

### LINKED ISSUE #1409: Small UI/UX bug: unrelated "Add selected to downloads" action is being shown for "Local Library" tracks
**Platform:**

Linux Ubuntu 22.10 x86_64

**Nuclear version:**

nuclear-58dfc7.AppImage

**Description of the issue:**

1. Make sure to have some tracks in your local library first
2. Open "Collection" -> "Local Library" page. Open inline (..) menu for any of your local tracks. You'll see "Download" item in there, which doesn't make sense, as the track is already there in your local library

image

4. Now select few tracks and click on the similar (...) button in the table header - you'll see the similar item in there:

image

I can imagine, there're tricky use cases related to other pages, like "Favorite Tracks", where user can have a mixture of tracks from internet and his local ones. But at least here, on a "Local library" page, it doesn't make sense to show it.

Alzamer:
Fixes #1409 by adding prop to appropriate component.


packages/ui/lib/components/TrackTable/Cells/TitleCell.tsx
```diff
@@ -71,7 +72,7 @@ const TitleCell: React.FC<CellProps<Track> & TrackTableExtraProps<Track>> = ({
         withAddToQueue={Boolean(onAddToQueue)}
         withAddToFavorites={Boolean(onAddToFavorites)}
         withAddToPlaylist={Boolean(onAddToFavorites)}
-        withAddToDownloads={Boolean(onAddToDownloads)}
```
nukeop, line 74: There is no need for an additional flag here. The presence of this button should be controled only by the presence of the `onAddToDownloads` callback. If it doesn't work, we should check why, and fix the cause.

Josef-Kretz: It looks like there's no way to pass props to the track from library view( packages\app\app\components\LibraryView) to the track/track pop up.
I think the onAddToDownloads is set from the tracktablecontainer (packages\app\app\containers\TrackTableContainer\index.tsx) independent of whatever props you pass. Maybe there's a way to modify the tracks that are passed to this container so that it doesn't render the download button? I'm new so I'm not certain about any of this

Alzamer: @Josef-Kretz This is my first contribution to this repo, and I'm not familiar with the code either. I've changed it, now you should pass prop to TrackTableContainer - then if the prop is set to true, the handler is passed (or not). I think it is correct, I will wait for @nukeop  to approve it.

nukeop:
I have made a separate branch to illustrate what I wanted to say better: https://github.com/nukeop/nuclear/tree/fix/1409
I have also added missing tests for this bug. Can you pull this or simply paste the tests + my fix for the "select all" menu?

nukeop (changes requested):
Take a look at the comment in the main PR thread for details

Alzamer:
@nukeop Thanks for help!

nukeop:
Thanks!

PR #1417, opened 2023-03-09
fix: sound-cloud-integration - WIP

thisprojects:
This is work in progress - not intending to merge. 

Added more changes as of 31 Mar. Implemented StreamingStrategy component.

### DIFF
--- a/package.json
+++ b/package.json
@@ -209,6 +209,7 @@
   },
   "dependencies": {
     "node-gyp": "^9.0.0",
+    "soundcloud-scraper": "^5.0.3",
     "sqlite3": "^5.1.2"
   }
 }
--- a/packages/app/app/actions/queue.ts
+++ b/packages/app/app/actions/queue.ts
@@ -55,7 +55,7 @@ const getSelectedStreamProvider = (getState) => {
     }
   } = getState();
 
-  return _.find(streamProviders, { sourceName: selected.streamProviders });
+  return _.find(streamProviders, { sourceName: 'Soundcloud' });
 };
 
 export const getTrackStreams = async (
--- a/packages/app/app/actions/queue.ts
+++ b/packages/app/app/actions/queue.ts
@@ -54,8 +54,7 @@ const getSelectedStreamProvider = (getState) => {
       selected
     }
   } = getState();
-
-  return _.find(streamProviders, { sourceName: 'Soundcloud' });
+  return _.find(streamProviders, { sourceName: selected.streamProviders });
 };
 
 export const getTrackStreams = async (
--- a/packages/app/app/containers/SoundContainer/StreamStrategy.jsx
+++ b/packages/app/app/containers/SoundContainer/StreamStrategy.jsx
@@ -1,0 +1,90 @@
+import { filterFrequencies } from '../../components/Equalizer/chart';
+import { StreamFormats } from '@nuclear/core/src/plugins/plugins.types';
+import HlsPlayer from '../../components/HLSPlayer';
+import Sound, { Volume, Equalizer, AnalyserByFrequency } from 'react-hifi';
+import Normalizer from '../../components/Normalizer';
+import VisualizerContainer from '../VisualizerContainer';
+import React from 'react';
+
+const StreamStrategy = ({
+  currentStream,
+  player,
+  settings,
+  currentTrack,
+  methods,
+  streamType,
+  actions,
+  enableSpectrum,
+  location,
+  usedEqualizer
+}) => {
+  const {
+    handleError,
+    handleFinishedPlaying,
+    handlePlaying,
+    handleLoaded,
+    handleLoading
+  } = methods;
+
+  const { HLS, NONHLS } = StreamFormats;
+
+  const HLSPlayer = (
+    <HlsPlayer
+      source={currentStream?.stream}
+      onError={handleError}
+      playStatus={player.playbackStatus}
+      onFinishedPlaying={handleFinishedPlaying}
+      muted={player.muted}
+      volume={player.volume}
+    />
+  );
+
+  const NonHLSPlayer = (
+    <Sound
+      url={currentStream?.stream}
+      playStatus={player.playbackStatus}
+      onPlaying={handlePlaying}
+      onFinishedPlaying={handleFinishedPlaying}
+      onLoading={handleLoading}
+      onLoad={handleLoaded}
+      position={player.seek}
+      onError={handleError}
+    >
+      <Normalizer url={currentStream?.stream} normalize={settings.normalize} />
+      <Volume value={player.muted ? 0 : player.volume} />
+      <Equalizer
+        data={filterFrequencies.reduce(
+          (acc, freq, idx) => ({
+            ...acc,
+            [freq]: usedEqualizer.values[idx] || 0
+          }),
+          {}
+        )}
+        preAmp={usedEqualizer.preAmp}
+      />
+      <AnalyserByFrequency
+        frequencies={filterFrequencies}
+        onVisualisationData={enableSpectrum && actions.setSpectrum}
+      />
+      <VisualizerContainer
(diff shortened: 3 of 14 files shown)


package.json
```diff
@@ -209,6 +209,7 @@
   },
   "dependencies": {
     "node-gyp": "^9.0.0",
+    "soundcloud-scraper": "^5.0.3",
```
thisprojects, line 212: might be able to ditch this dependancy - all we are using it for is client_id keygen and soundcloud search functionality

packages/app/app/actions/queue.ts
```diff
@@ -55,7 +55,7 @@ const getSelectedStreamProvider = (getState) => {
     }
   } = getState();
 
-  return _.find(streamProviders, { sourceName: selected.streamProviders });
+  return _.find(streamProviders, { sourceName: 'Soundcloud' });
```
thisprojects, line 58: How is the selected stream provider set?

I've set SoundCloud as the default streamProvider - however the selected stream provider is still always set to youtube.

thisprojects, line 58: How is the selected stream provider set?

I've set SoundCloud as the default streamProvider - however the selected stream provider is still always set to youtube.

packages/app/app/containers/SoundContainer/index.js
```diff
@@ -183,14 +183,18 @@ class SoundContainer extends React.Component {
   }
 
   isHlsStream(url) {
-    return /http.*?\.m3u8/g.test(url);
+    if (url?.stream){
+      return url.stream?.includes('m3u8');
+    } else {
```
thisprojects, line 188: Not sure if HLS player is currently working as expected. When files are played there is no progress bar. 

Also the HLS stream detecting method only supports an array of streams.

packages/core/src/plugins/stream/SoundcloudPlugin.ts
```diff
@@ -12,13 +12,14 @@ class SoundcloudPlugin extends StreamProviderPlugin {
     this.sourceName = 'Soundcloud';
     this.description = 'Allows Nuclear to find music streams on Soundcloud';
     this.image = null;
+    this.isDefault = true;
   }
 
   resultToStream(result): StreamData {
     return {
-      source: this.sourceName,
+      source: 'Soundcloud',
```
thisprojects, line 20: for some reason this property will not pick up sourceName from its instance. Hard coded for now.

packages/core/src/plugins/stream/SoundcloudPlugin.ts
```diff
@@ -29,7 +30,8 @@ class SoundcloudPlugin extends StreamProviderPlugin {
   async search(query: StreamQuery): Promise<StreamData[] | undefined> {
     const terms = query.artist + ' ' + query.track;
     try {
-      const results = await(await Soundcloud.soundcloudSearch(terms)).json();
+      const results = [];
```
thisprojects, line 33: The new soundcloud track details api returns single results, rather than an array. We could make a track details call for each search result, this might end up in a lot of network requests though.

packages/core/src/plugins/stream/YoutubePlugin.ts
```diff
@@ -10,7 +10,7 @@ class YoutubePlugin extends StreamProviderPlugin {
     this.sourceName = 'Youtube';
     this.description = 'A plugin allowing Nuclear to search for music and play it from Youtube';
     this.image = null;
-    this.isDefault = true;
+    // this.isDefault = true;
```
thisprojects, line 13: trying to prevent youtube from being default stream, for testing.

packages/core/src/rest/Soundcloud.ts
```diff
@@ -1,13 +1,52 @@
+import SoundCloud from 'soundcloud-scraper';
 const apiUrl = 'https://api.soundcloud.com';
 
+let client;
+
+const generateClient = async () => {
+  const key =  await SoundCloud.Util.keygen();
+  client = new SoundCloud.Client(key);
```
thisprojects, line 8: Need to use keygen as old api key doesn't work - or isn't "fresh" enough.

packages/core/src/rest/Soundcloud.ts
```diff
@@ -1,13 +1,52 @@
...
   return `${url}&client_id=${process.env.SOUNDCLOUD_API_KEY}`;
 }
 
-export function soundcloudSearch(terms: string): Promise<Response> {
-  return fetch(prepareUrl(apiUrl + '/tracks?limit=50&q=' + terms));
+export async function soundcloudSearch(terms: string) {
+  const trackData = [];
+  const searchResults = await client.search(terms)
+    .then(async song => song)
+    .catch(console.error);
```
thisprojects, line 21: To get the stream url we have to make 3 network requests. 

1. Search soundcloud for the correct artist name and song title (returns multiple)
2. Get the track details (returns 1)
3. Get the stream details (returns 1)

packages/app/app/containers/SoundContainer/index.js
nukeop: HLS is for live streams, so there's no progress bar, because there's no way to tell when it's going to end.

package.json
nukeop: That client id is crucial - if you can make do without it, feel free to remove it.

packages/app/app/actions/queue.ts
nukeop: It should be retrieved from the plugin settings - if it doesn't work it should be handled in a separate bug. Shouldn't be hardcoded here.

packages/core/src/plugins/stream/SoundcloudPlugin.ts
nukeop: Looks like `this` won't work here because it's not bound (this isn't an arrow function). If you change it to an arrow function, you should be able to revert this.

packages/core/src/plugins/stream/SoundcloudPlugin.ts
nukeop: There should be one search request, then we just display what we can.

packages/app/app/containers/SoundContainer/StreamStrategy.jsx
```diff
@@ -0,0 +1,90 @@
...
+        trackName={
+          currentTrack
+            ? `${currentTrack.artist} - ${currentTrack.name}`
+            : undefined
+        }
+      />
+    </Sound>
+  );
+
+  switch (streamType) {
```
thisprojects, line 80: Handled return as a swtich - handling default case is cleaner than using object with dynamic key.

packages/core/src/plugins/plugins.types.ts
```diff
@@ -1,117 +1,123 @@
...
 };
 
 export type StreamQuery = {
-    artist: string;
-    track: string;
+  artist: string;
+  track: string;
+};
+
+export enum StreamFormats {
```
thisprojects, line 102: Added stream formats, allows us to expand with other formats if / when required in future.

packages/core/src/plugins/stream/SoundcloudPlugin.ts
```diff
@@ -38,13 +41,7 @@ class SoundcloudPlugin extends StreamProviderPlugin {
...
 
   async getStreamForId(id: string): Promise<undefined | StreamData> {
-    try { 
-      const result = await(await Soundcloud.getTrackById(id)).json();
-      return result.id && this.resultToStream(result);
-    } catch (err) {
-      logger.error(`Error while looking up streams id: ${id} on Soundcloud`);
-      logger.error(err);
-    }
+    return null;
```
thisprojects, line 44: Sound cloud API no longer supports track ids - so returning null here for now. Any thoughts on this?

PR #1419, opened 2023-03-10
FIX 1409

oyi77:
Hi this is my first contribution on public repository, i am just a newbie in react, so please tell me if there's something i need to change. Here i try just to pass forward a "displayDownloadButton" props, just following the code style from the elders 🧓 

Here i attach the screenshot of the result, please check my PR 

Screenshot 2023-03-11 at 00 32 57
Screenshot 2023-03-11 at 00 33 07


nukeop:
There are two problems here, there's already a pull request for this exact same issue, and this is not the way to go about this, the download button appears if a callback is passed to handle downloading. It should not be replaced with a flag.

oyi77:
okay i got it, so actually we just need to prevent callback to it. i using flag, because just as i remember, you state that you want to solve this issue using flag.

I will close this, as there are already pull request with same issue

PR #1431, opened 2023-03-21
Update download.ts

### LINKED ISSUE #1430: download.ts does not check for invalid characters in filename before starting download
**Platform:**
Problem is bigger on Windows than on Linux or MacOS because there are more invalid characters.

**Nuclear version:**
0.6.17

**Description of the issue:**
Here is a suggested fix, that has been tested and working:
```
diff --git a/packages/main/src/controllers/download.ts b/packages/main/src/controllers/download.ts
index 195b34a5..ec011499 100644
--- a/packages/main/src/controllers/download.ts
+++ b/packages/main/src/controllers/download.ts
@@ -8,6 +8,7 @@ import { getTrackArtist, getTrackTitle } from '../utils/tracks';
 import Download from '../services/download';
 import Logger, { $mainLogger } from '../services/logger';
 import Window from '../services/window';
+import { platform } from 'os';
 
 interface DownloadRef {
   uuid: string;
@@ -23,6 +24,25 @@ class DownloadIpcCtrl {
     @inject(Window) private window: Window
   ) { }
 
+  removeInvalidCharacters(filename: string): string {
+    let invalidChars: string[];
+
+    switch (platform.toString()) {
+    case 'darwin': // Mac OS
+      invalidChars = [':', '/'];
+      break;
+    case 'linux': // Linux
+      invalidChars = ['/'];
+      break;
+    default: // Windows
+      invalidChars = ['<', '>', ':', '"', '/', '\\', '|', '?', '*'];
+      break;
+    }
+
+    const sanitizedFilename = filename.split('').filter((char) => !invalidChars.includes(char)).join('');
+    return sanitizedFilename;
+  }
+
   /**
    * Start a download using the download service
    */
@@ -41,9 +61,10 @@ class DownloadIpcCtrl {
       const artistName = getTrackArtist(data);
       const title = getTrackTitle(data);
 
-      const filename = `${artistName} - ${title}`;
+      // .replace(/[/\\?%*:|"<>]/g, '-') or equivalent invalid characters based on platform
+      const filename = this.removeInvalidCharacters(`${artistName} - ${title}`);
 
-      this.logger.log(`Start Download: ${artistName} - ${title}`);
+      this.logger.log(`Start Download: ${filename}`);
 
       await this.download.start({
         query: {
```

It could also be fixed by not adding the os.platform dependency and just replacing the invalid characters of the least forgiving platform: ['<', '>', ':', '"', '/', '\\', '|', '?', '*']
Something like:

```const filename = `${artistName} - ${title}`.split('').filter((char) => !['<', '>', ':', '"', '/', '\\', '|', '?', '*'].includes(char)).join('');```

### PR DESCRIPTION
**Pull request by bannan938**
fix #1430

### DIFF
--- a/packages/main/src/controllers/download.ts
+++ b/packages/main/src/controllers/download.ts
@@ -8,7 +8,7 @@ import { getTrackArtist, getTrackTitle } from '../utils/tracks';
 import Download from '../services/download';
 import Logger, { $mainLogger } from '../services/logger';
 import Window from '../services/window';
-import { platform } from 'os';
+import Platform from '../services/platform';
 
 interface DownloadRef {
   uuid: string;
@@ -21,22 +21,19 @@ class DownloadIpcCtrl {
   constructor(
     @inject(Download) private download: Download,
     @inject($mainLogger) private logger: Logger,
-    @inject(Window) private window: Window
+    @inject(Window) private window: Window,
+    @inject(Platform) private platform: Platform
   ) { }
 
   removeInvalidCharacters(filename: string): string {
     let invalidChars: string[];
 
-    switch (platform.toString()) {
-    case 'darwin': // Mac OS
+    if (this.platform.isMac()) {
       invalidChars = [':', '/'];
-      break;
-    case 'linux': // Linux
+    } else if (this.platform.isLinux()) {
       invalidChars = ['/'];
-      break;
-    default: // Windows
+    } else { // Windows and other platforms
       invalidChars = ['<', '>', ':', '"', '/', '\\', '|', '?', '*'];
-      break;
     }
 
     const sanitizedFilename = filename.split('').filter((char) => !invalidChars.includes(char)).join('');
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: packages/main/src/controllers/download.ts [hunk 1]
```diff
@@ -8,6 +8,7 @@ import { getTrackArtist, getTrackTitle } from '../utils/tracks';
 import Download from '../services/download';
 import Logger, { $mainLogger } from '../services/logger';
 import Window from '../services/window';
+import { platform } from 'os';
```
nukeop, line 11: Could you use this service for determining the platform?
https://github.com/nukeop/nuclear/blob/facaa4065024abd9f1786b10041d85b46a9dfd20/packages/main/src/services/platform/index.ts

bannan938: Of course. I have changed the code to use the platform service and pushed it to this branch. Did that also update the pull request?

nukeop:
Excellent, thank you for contributing.

PR #1444, opened 2023-04-09
fix: add tmate to coverage.yml

thisprojects:
Allows SSH connection to the coverage action task runner for debugging. If it works we can setup SSH keys, currently anyone watching the coverage task action can get the connection string.

Documentation is here : https://mxschmitt.github.io/action-tmate/


nukeop:
You can just rerun the action without commiting anything

Issue #1434, opened 2023-03-24
Resume last played music on queue when reopening the app


sith-on-mars:
Currently, if you play music on a queue, reopening the app will lead to the loss of the queue history. It would be great if the app could resume last played music on the queue and keep queue history, unless the user change to another queue or playlist. So basically the app can emulate default Sportify player does.

nukeop:
Thanks for the suggestion, this is a good idea.

DxVxDOD:
Hi I would like to work on this issue.

fastio19:
Hi @sith-on-mars, I would like to contribute on this issue if it is still opened.

mjsf1234:
Hi is this issue still open I can start working for it

nukeop:
No, someone is working on it.

vinay94185:
@nukeop  i checked all accounts, no one seems to be actively working on this issue for over a month, i don't want to be rude to anyone but if no one's working on it, i would love to take it.

thank you.

nukeop:
There's an open pull request on this.

vinay94185:
thanks for update. i see it.

sameer-haider:
Can I be assigned this.

SamarthBagga:
Hi @nukeop, since I do not think someone is working on this issue currently. I would like to work on it. I wanted to know where should the queue's state be stored, so that it can be retrieved once the app closes and reopens?

nukeop:
This would be the 5th time I'm assigning someone to this issue.

SamarthBagga:
hi @nukeop , there is no need to assign the issue to me. I am not sure yet if ill be able to provide a fix, I am just trying to grasp the issue and trying to find a fix. All I wanted to know is that where should I store the queue's state so that it can be retreieved once the app closes and then reopens?

nukeop:
I'm going to paste what I said on Discord:

- it should probably be contained in the app package
- last time someone tried this I suggested implementing a redux subscriber that listens to anything related to adding/removing/swapping tracks in the queue, and saves it in a json file. If you search by `subscribe` you should find existing subscribers to get an idea how to use them.
- this can either be a separate file (using `electron-store`) or in the regular config file, just make sure that the write is non-blocking
- if possible, use atomic writes* to ensure the queue is never corrupted
- the program should read the queue when starting, and if it fails it should simply start with a blank one
- ideally, there should be a setting that controls whether this feature is enabled
[...]

PR #1445, opened 2023-04-09
Work in progress: Issue #1434 (queue retention)

joewrightlaprugne:
Keeps track of the current queue in the config file

The next step is to get the app to restore the saved queue on startup


nukeop:
That's not the right approach. It's not sensible to add this code to every action that modifies the queue. Instead, there should be a redux subscriber that listens to changes in the queue, and, asynchronously, saves the queue (using atomic write, so that nothing is lost if the process exits in the middle of a write). This way it's easier to separate concerns and keep it configurable, and it doesn't impact other functionalities.

Issue #1093, opened 2021-11-09
[Feature Request] "Play Random" button near "Play" button on favorite tracks menu.


JFMugen:
I think it would be awesome to find new songs. You can just play a random song from your favorites list and Autoradio would do the rest.

nukeop:
This issue is from an older version of Nuclear that has been replaced with a complete rewrite. None of the old issues apply to the new codebase.

The old code is preserved on the [legacy/electron](https://github.com/nukeop/nuclear/tree/legacy/electron) branch. If you have feedback about the new version, head to [Discussions](https://github.com/nukeop/nuclear/discussions).

PR #1450, opened 2023-04-13
Added play random #1093

Alzamer:
Added play random button at favorites tracks bookmarks, as suggested in #1093.


nukeop:
I made a couple of changes:
- Moved the button from the player bar to the favorite tracks view. This way you don't need to hack the check for the url.
- Added a tooltip for the button, so it's possible to find out what it's supposed to do.
- Added a test for this functionality (happy path).
- Updated snapshots to reflect that this button was added.

nukeop:
Thanks for cooperation!

Issue #1420, opened 2023-03-11
"Play next" feature puts a song on the bottom of queue and shifts current playing song


Saltane:
**Platform: GNU/Linux**

**Nuclear version: nuclear-fd06a3.AppImage**

**Description of the issue:** having some queue already playing.
![image](https://user-images.githubusercontent.com/101461606/224485330-d82d6bb0-74a4-43e8-9c1e-bd7ce590c523.png)

Having local playlist opened, clicking playlist's song three-dot menu next to + button and selecting "Play next" option.
![image](https://user-images.githubusercontent.com/101461606/224485398-c3114325-6218-43af-8d71-9977ede81540.png)

Nuclear adds chosen song on the second-bottom place in queue and to make things worse, it starts playing one song back relatively to current playing one, so in this particular case it was Skillet playing and after new song was added via "Play next" feature it started playing Johny cash, which was one before Skillet.
![image](https://user-images.githubusercontent.com/101461606/224485425-7ec35ad4-2fd1-4922-8446-f3d698c6d14a.png)

**Expected:** "Play next" puts chosen song right after current playing one and lets current playing song continue playing.

nukeop:
I can confirm that this doesn't work as intended currently, needs to be tested and fixed. I'll try doing it soon.

terryyLi:
Hi my team would like to try fixing it. Could you assign it to me?

nukeop:
Sure. Here are our developmet resources: https://nukeop.gitbook.io/nuclear/developer-resources/development-process

akakayan:
Hi, I make some change and think I've fixed the issue. I've submitted the pull request. Do I need to be an assignee or if there is anything else I should do?

PR #1452, opened 2023-04-19
fix Playnext button: #1420

akakayan:
I fixed the issue mentioned in #1420 by put the PlayNext Song after the current song. 

This is my first time contributing to this project as well as an open source project so just tell me if there is anything else I need to do or should be fixed.


nukeop:
Looks good, would you be able to add a test for this? You can use existing tests as a guide.

akakayan:
Sure I will do that . Just want to make sure, I think I am supposed to add the test to trackpopupcontainer.test.tsx? That's the place I found where play next button is relevant. However, there is no pervious test about it so I am not sure if that's the place I should go to.

akakayan:
Hi, I've added a test for playNext and it passed on my local machine, can you review that?

nukeop:
Perfect, thank you fo rimplementing this.

PR #1453, opened 2023-04-19
Fix typos and update the PT-BR version of electron.md

brunoSnoww:
This PR makes some corrections  on typos in the Brazilian version of the Electron.md and also syncs it with the English version.


nukeop:
nice, thanks for contributing! It's not necessary to translate my rant, but I won't reject it when it's already done.

PR #1454, opened 2023-04-21
go back button restart song function added

HiroakiYo:
I have changed what actions will be dispatched to the go-back button under PlayerBarContainer. When you hit go back when playing the first song in the queue, the song should restart from the beginning. If there are other songs before the current song in the queue, go back button will still go to the previous song.


nukeop:
Nice, looks fine, but could you please add a test for it? I've added a bunch of player bar tests recently, you can use them as guidelines.

nukeop:
Great! Thanks for coding this.

PR #1459, opened 2023-04-29
No Repeating Tracks in Shuffle Mode

### LINKED ISSUE #1125: Don't repeat tracks in shuffle mode
**Platform:**
Void Linux

**Nuclear version:**
0.6.17

**Description of the issue:**
In shuffle mode, it would be nice if shuffle did not repeat any tracks until all of the other tracks in the queue have been exhausted. I think tracks should only be repeated once all the tracks in the queue have been played once.

torqueJY32:
Resolves #1125 

As mentioned in issue #1125 , popular streaming applications do not repeat recently played songs in shuffle mode, and we would like to see the same behavior in nuclear. The current implementation randomly selects songs without considering whether they have been recently played. To address this, this pull request proposes a modification to the song queue implementation. A new attribute called played has been added to each song in the queue. This pull request utilizes the played attribute to ensure that all songs are played before any song is repeated. By implementing this change, we can enhance the shuffle mode functionality and provide a better user experience in the nuclear application.


nukeop:
This is pretty complex, could you add a few tests for this functionality? And then we can think about how to reduce the complexity and clean up the code.

torqueJY32:
Thanks for the reply! Currently it looks like the checks are not passing in this PR as we try to merge with the master branch, but when we were testing things out in our forked branch all tests were passing. I think our branch was forked a few days ago so the tests were ok, but the master branch may have some issue. We will try to write some test cases for now and will let you know once we are done!

nukeop:
This is because I had to switch to a fork of ytdl-core recently, which broke the tests. I'm going to test out a different fork as a temporary measure, and then all tests should work fine.

terryyLi:
Hi we've tested it locally on app which works fine but for the same scenario when we put in test it failed. Specifically, when shuffle mode is on and current song is 0, next song is again 0 in test while when we run app it works fine. Basically what we implemented was 1. added a played boolean field to track if a song is already played 2. whenever click next make the current song played to be true and choose next song in a list of unplayed songs 3. if all songs all played set every song to be not played yet again. Would you have time take a look? I believe it was the way we wrote test wrong but our implementation is working. Many thanks!

nukeop:
Try pulling from master now, it should fix the tests.

terryyLi:
> Try pulling from master now, it should fix the tests.

Sure thanks

torqueJY32:
I believe the test should be all good except for the Windows build. that's also what's wrong in the main branch though

torqueJY32:
I believe the test case is ready, and all checks are passing except for the windows build

nukeop:
Great, thanks for updating.

torqueJY32:
hmm, still not passing the windows build, but I think our added code shouldn't cause this?

nukeop:
It's fine as long as the tests are passing. The build crashes because it runs out of memory, but the release build works fine - I'm going to have to fix the action

packages/app/app/containers/PlayerBarContainer/PlayerBarContainer.test.tsx
```diff
@@ -240,6 +240,53 @@ describe('PlayerBar container', () => {
...
+        shuffleQueue: true,
+        autoradio: false,
+        miniPlayer: false
+      }
+    }
+    );
+    const nextButton = await component.findByTestId('player-controls-forward');
+    fireEvent.click(nextButton);
+    const state = store.getState();
+    expect(state.queue.currentSong).not.toBe(1);
```
nukeop, line 287: Since shuffle plays a random song, this test can pass at random, even if the functionality doesn't work.

packages/app/app/actions/queue.ts
```diff
@@ -262,11 +264,23 @@ export function addPlaylistTracksToQueue(tracks) {
...
+    let unplayedQueue = queue.queueItems.filter(item => !item.played);
+    if (unplayedQueue.length === 0){
+      queue.queueItems.map(item => dispatch(updateQueueItem({ ... item, played: false})));
+    }
+    if (queue.currentSong !== null) {
+      const currentSongUuid = queue.queueItems[queue.currentSong].uuid;
+      dispatch(updateQueueItem({ ...queue.queueItems[queue.currentSong], played: true, uuid: currentSongUuid }));
+      queue = getState().queue;
+    }
+    unplayedQueue = queue.queueItems.filter(item => !item.played);
```
nukeop, line 279: What happens when all the tracks have this flag set to true?

packages/app/app/actions/queue.ts
```diff
@@ -262,11 +264,23 @@ export function addPlaylistTracksToQueue(tracks) {
...
   const settings = state.settings;
-  const queue = state.queue;
+  let queue = state.queue;
 
   if (settings.shuffleQueue) {
-    const index = _.random(0, queue.queueItems.length - 1);
-    dispatch(selectSong(index));
+    let unplayedQueue = queue.queueItems.filter(item => !item.played);
+    if (unplayedQueue.length === 0){
+      queue.queueItems.map(item => dispatch(updateQueueItem({ ... item, played: false})));
```
nukeop, line 272: For performance reasons, this should not result in an action being dispatched for every item in the queue.

packages/app/app/actions/queue.ts
terryyLi: if that happens line 272 will run which is `if (unplayedQueue.length === 0) { ... }`

packages/app/app/actions/queue.ts
terryyLi: this will actually not happen as currently the app will keep adding new songs to the queue. Even if this happens it will be only after a user has played every song in the queue. Do you still think this is an efficiency problem and any advice on what to do with this?

nukeop: New songs are only added if you have the autoradio setting enabled. We could introduce a new action that would reset this flag on the entire queue at once, so we would only dispatch one action in this case.

Issue #1396, opened 2023-02-07
Not all Miniplayer PlayOptions clickable


a0kami:
**Platform:**
Linux `Linux 6.0.19-3-MANJARO #1 SMP PREEMPT_DYNAMIC Wed Jan 18 07:57:11 UTC 2023 x86_64 GNU/Linux`

**Nuclear version:**
0.6.17 (latest AUR `nuclear-player-bin` "95cfe0")

**Description of the issue:**
When enabling the miniplayer, the bottom lower part is not clickable as shown in the screenshot.
![Screenshot_20230207_235051](https://user-images.githubusercontent.com/8460863/217385995-62c9351b-9dd3-4b62-8c63-e06bff264274.png)

I've managed to make it work locally by setting an arbitrarily high `z-index` value (100) on the scss properties of the `mini_play_options` object in file `packages/ui/lib/components/MiniPlayer/MiniPlayOptions/styles.scss`. eg:
```css
.mini_play_options {
  ...

  z-index: 100;
  
  ...
}

```

Not sure if there's a proper convention or a project specific way to set the z-index.

Also, I might look into adding a test for it.

nukeop:
Feel free to just put that value directly in the css. No need to write tests in this case.

maboroshin:
Title bar cannot be dragged in Windows 10. I cannot move the mini-player window. I don't know about other operating systems.

nukeop:
Fixed by #1592

PR #1468, opened 2023-06-13
Small changes to fix visual issues with mini player

clymerr:
Updated a few lines of code to correct issue #1396 by adding z-index under the related Sass file. Also found it a bit visually confusing to have playOptions show the mini player toggle when there is an existing NeumorphicBox _exclusively_ to exit - so filtered it out.

I didn't see any related existing tests, and changes are relatively finite in scope but happy to circle on that, if needed.

### DIFF
--- a/packages/ui/lib/components/MiniPlayer/MiniPlayOptions/index.tsx
+++ b/packages/ui/lib/components/MiniPlayer/MiniPlayOptions/index.tsx
@@ -60,7 +60,9 @@ const MiniPlayOptions: React.FC<MiniPlayOptionsProps> = ({
         />
       </button>
       {
-        isExpanded && playOptions.map(playOption => <MiniPlayOptionControl {...playOption} />)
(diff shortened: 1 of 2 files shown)


packages/ui/lib/components/MiniPlayer/MiniPlayOptions/index.tsx
```diff
@@ -60,7 +60,9 @@ const MiniPlayOptions: React.FC<MiniPlayOptionsProps> = ({
         />
       </button>
       {
-        isExpanded && playOptions.map(playOption => <MiniPlayOptionControl {...playOption} />)
+        isExpanded && playOptions.filter(playOption => {
+          return playOption.dataTestId !== 'mini-player-play-option';
```
nukeop, line 64: Hm, this isn't the best way to filter it out, because it should only be used for tests - maybe it should be filtered out at the source?

PR #1478, opened 2023-07-16
Fixed #1423

SamarthBagga:
#1423 
Added the drag functionality to the entire navbar using this-
[link to electronjs docs to change draggable area](https://www.electronjs.org/docs/latest/tutorial/window-customization#set-custom-draggable-region)
Also made the div containing the window control buttons not draggable so that the drag functionality doesnt supersede the resize functionality.

PR #1487, opened 2023-08-01
Fixed #1423

### LINKED ISSUE #1423: Cannot drag mini-player
**Platform:**

nuclear-fd06a3, Windows 10

**Description of the issue:**

Title bar cannot be dragged in Windows 10. I cannot move the mini-player window. I don't know about other operating systems.

At 800 pixels wide, you can drag the upper right corner of the window. (orange in the image below) At 300 pixels wide, there is no draggable area.

image

SamarthBagga:
I have fixed #1423. This time I have made sure that the search bar or other navbar elements do not stop working because of the draggable feature. Thus all the other features are working like they should.

### DIFF
--- a/packages/app/app/components/HelpButton/styles.scss
+++ b/packages/app/app/components/HelpButton/styles.scss
@@ -11,6 +11,8 @@
   padding: 0;
   margin-right: 12px;
   transition: $short-duration !important;
+  -webkit-app-region: no-drag;
+
 
   i {
     position: absolute;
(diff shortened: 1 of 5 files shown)


SamarthBagga:
hi @nukeop , the search bar and other features are working unlike my previous pr #1478 . Although the only issue is that we can only have one feature either the resizing or dragging in this part of the player, as if i enable the draggable feature the resizing wont happen but if want the player to be resizable from that point then we cannot make it draggable.
Capture

SamarthBagga:
hi @nukeop could you review this pr?

PR #1495, opened 2023-08-13
Fix yay install command

yagarea:


nukeop:
Thanks, I didn't notice it

PR #1521, opened 2023-11-01
FEAT: Add "copy to clipboard" feature

### LINKED ISSUE #1516: Ability to copy the song title to clipboard
As the title says, I'd appreciate it if we were able to copy the song titles. At least in my case, I listen to quite a few songs in different languages that use diacritics or a different script. Currently, the only way to do that is by downloading them, which does not work that well, unfortunately.

shantanusoni72:
In this pull request, users can now able to copy the song title to the clipboard by simply clicking on its name. 

Fixes #1516 

Demonstration Video
https://github.com/nukeop/nuclear/assets/75296055/87a86b75-63bc-47b8-90b4-9343f09a010a

### DIFF
--- a/packages/ui/lib/components/TrackInfo/index.tsx
+++ b/packages/ui/lib/components/TrackInfo/index.tsx
@@ -4,6 +4,15 @@ import { Icon } from 'semantic-ui-react';
 import Cover from '../Cover';
 import styles from './styles.scss';
 
+const CopyTextToClipboard = async () => {
+  const track_element = document.getElementById('track_name');
+  try {
+    await navigator.clipboard.writeText(track_element.innerHTML);
+  } catch (err) {
+    console.error('Failed to copy: ', err);
+  }
+};
+
 export type TrackInfoProps = {
   cover?: string;
   track: string;
@@ -33,7 +42,7 @@ const TrackInfo: React.FC<TrackInfoProps> = ({
       hasTracks &&
         <>
           <div className={styles.artist_part}>
-            <div className={styles.track_name} onClick={onTrackClick}>
+            <div id='track_name' className={styles.track_name} onClick={() => CopyTextToClipboard()}>
               {track}
             </div>
             <div className={styles.artist_name} onClick={onArtistClick}>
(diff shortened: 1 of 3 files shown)


packages/ui/lib/components/TrackInfo/index.tsx
```diff
@@ -4,6 +4,15 @@ import { Icon } from 'semantic-ui-react';
 import Cover from '../Cover';
 import styles from './styles.scss';
 
+const CopyTextToClipboard = async () => {
+  const track_element = document.getElementById('track_name');
+  try {
+    await navigator.clipboard.writeText(track_element.innerHTML);
```
nukeop, line 10: This shouldn't be done this way in React. Instead, you can get the title from the redux store.

packages/ui/lib/components/TrackInfo/index.tsx
```diff
@@ -4,6 +4,15 @@ import { Icon } from 'semantic-ui-react';
...
+  } catch (err) {
+    console.error('Failed to copy: ', err);
```
nukeop, line 12: Don't use console.error, instead you can show a notification toast.

nukeop:
Hi, thanks for the pull request. Please check the contribution guidelines: https://nukeop.gitbook.io/nuclear/contributing/contribution-guidelines

There are a few things missing but I'm sure if we iterate on it we can get it into a good shape.
- Please add a test for the feature. You can see examples in the app package. Let me know if you have trouble testing the clipboard contents.
- Instead of changing the color to green with CSS, it should be possible to know what's going to happen if you click the title. So a good idea would be to add a tooltip that says something like "click to copy". You can search for the word "tooltip" to see examples.
- The CSS on the title should also change the cursor to pointer when you hover over it, to let the user know this element is clickable. It might also be a good idea to add a button with an icon next to the title, so you can tell there's something interactive there at a glance.

shantanusoni72:
Logs:
1. Green color removed.
2. Add a pointer when hovering over the name.
3. Add the "Click to copy" tooltip. 
4. Removed getElementById line.

https://github.com/nukeop/nuclear/assets/75296055/bed28318-36a3-41da-bf00-9b808166d806

It would be cool to add a notification toast.

PR #1545, opened 2024-01-31
tray menu options now have translations when a user changes their lan…

maxx-niu:
changes to traymenu only apply after a language change is issued from the settings. Not sure how to apply the translations with the configure language when the app launches


nukeop:
Thanks, looks good!

PR #1548, opened 2024-02-10
Fix Bandcamp plugin search aborting after the first page

sferra:
The Bandcamp plugin aborts the search after the first page of results because of an incorrect check.
I found this bug while writing tests for an improvement/issue in the Bandcamp plugin and thought that it's better to submit it separately.


nukeop:
Thanks, this must have changed since it was last implemented.

sferra:
You're welcome. Please let me know if I have to adapt the tests, because I intend to submit an additional fix and extend them to cover 2 new internal functions.

nukeop:
This test format looks good to me and is consistent with the rest of the codebase.

PR #1551, opened 2024-02-11
Update documentation in README.md

ursapillay:
### Updates
- Added social badges
- Added github analytics badges
- Updated description
- Added installation section
- Updated links section
- Reorganized sections

### Why
Improve readability and information in README file to better reflect project


nukeop:
I agree with Nuki that the links should stay near the top. However, some of them can be removed since you replaced them with nice icons. I like the rest of your changes, particularly rewording the descriptions, and adding the link to the downloads page.

I just noticed that the flag buttons no longer work for some reason... 😔 They should probably be replaced later.

BTW. Nuki's review is not completely accurate, you can ignore it while I tweak this.

PR #1552, opened 2024-02-13
Fix stream lookup with 'undefined' stream ID 

sferra:
The current queue logic attempts to look-up an (undefined) track ID even if no tracks are available.
Such a look-up doesn't make any sense and the queue behavior should be actually specified for such a case. 

I'm willing to fix, refactor the queue functionality and add test coverage if someone is able to provide guidance as to what the correct queue behavior would be.

For now, I just added a first test proving the problematic execution path leading to `streamProvider.getStreamForId(undefined)` .

### DIFF
--- a/packages/app/app/actions/queue.test.ts
+++ b/packages/app/app/actions/queue.test.ts
@@ -1,0 +1,67 @@
+import { QueueItem } from '../reducers/queue';
+import { Queue } from './actionTypes';
+import * as QueueOperations from './queue';
+
+describe('Queue actions tests', () => {
+
+  describe('finds streams for track', () => {
+    const getSelectedStreamProvider = jest.spyOn(QueueOperations, 'getSelectedStreamProvider');
+    const getTrackStreams = jest.spyOn(QueueOperations, 'getTrackStreams');
+
+    afterEach(() => {
+      getSelectedStreamProvider.mockReset();
+      getTrackStreams.mockReset();
+    });
+
+    test('remote track is removed from the queue when no streams are available', () => {
+      getTrackStreams.mockResolvedValueOnce([]);
+      const streamProvider = {
+        getStreamForId: async (streamId: string) => {
+          if (!streamId) {
+            throw new Error('The stream ID must not be undefined');
+          }
+          return {};
+        }
+      };
+      getSelectedStreamProvider
+        .mockReturnValueOnce(streamProvider);
+
+      const trackIndex = 123;
+      const queueItems: QueueItem[] = [];
+      queueItems[trackIndex] = {
+        artist: 'Artist Name',
+        name: 'Track Name',
+        local: false,
+        streams: null
+      };
+      const stateResolver = () => ({
+        queue: {
+          queueItems
+        },
+        settings: {
+          useStreamVerification: false
+        }
+      });
+
+      const dispatchOperation = jest.fn();
+      const findStreamsForTrackOperation = QueueOperations.findStreamsForTrack(trackIndex);
+      findStreamsForTrackOperation(dispatchOperation, stateResolver)
+        .then(() => {
+          // no error should have been dispatched
+          expect(dispatchOperation).not.toHaveBeenCalledWith(
+            expect.objectContaining({
+              payload: expect.not.objectContaining({
+                error: expect.anything()
+              })
+            })
+          );
+          // assumption:
+          // track without streams should have been removed from the queue
+          expect(dispatchOperation).toHaveBeenCalledWith({
+            type: Queue.REMOVE_QUEUE_ITEM,
+            payload: { trackIndex }
+          });
+        });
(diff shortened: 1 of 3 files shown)


nukeop:
I'm not a fan of unit tests. Have you tried testing it by adding the track to the queue, instead of calling actions directly? It should be possible to set it up the same way.

sferra:
Sorry, I don't understand how you want me to test. 
I don't know what you mean when you refer to the queue and actions.

The current test can be simplified a bit by excluding the error, but my first intention was to first prove the issue. The test covers a single, very specific execution path since the function is quite big (which is also the reason why the test is as convoluted as it is). As you can see in 99710651ac07bc0fd27253dadecf9078750c0fc2, the fix is quite straight-forward.  I don't want to test anything more that the code I changed.

I'm open for suggestions. Maybe point me to an existing test which I can use as an example and adapt for this specific case?

nukeop:
I mean that in Nuclear, I avoid unit tests, unless it's impossible. The default way to test any functionality is by writing integration tests at the level of views or high level containers, which you can see e.g. by inspecting the PlayQueueContainer tests. Do you think this can be moved there?

sferra:
I had a look at `PlayQueueContainer` , but that component doesn't interact at all with the ` findStreamsForTrack`  function. ` PlayerBarContainer` is the only component which triggers the function, but it doesn't have much to do with the queue.

I attempted to set up a test along with the existing tests of `PlayerBarContainer` but failed miserably. I got a "Maximum update depth exceeded" error message, probably caused be recursive dispatch calls. I can't figure it out. Also `PlayerBarContainer` calls `useStreamLookup` (which in turn triggers `findStreamsForTrack`) immediately and not by UI interaction, so I'm not sure how to perform verification.

I'm at a loss and don't know how to proceed.

sferra:
We can inspect the queue state without needing `PlayQueueContainer`, which we don't really need to interact with.  I actually had a test set up for `PlayerBarContainer` locally for using `mountConmponent` and an IMO appropriate store state. The issue was that the queue changed while the container was created because of the the track removal, causing deep recursion, I think. I'll push my local attempts at setting up the test to a branch later and maybe you can give me some pointers as to what's wrong with them.

nukeop:
Thanks! I'll try to take a look at this today or tomorrow.

nukeop:
Ok, I was able to reproduce the infinite loop if I use `waitFor`. Debugging

nukeop:
I found out that there's an item placed at index -1 in the queue array, and that's why there's an infinite loop.

nukeop:
Ok, I fixed it. You were missing an uuid for that test track. After adding it, the test works.

sferra:
Thank you! 

It wasn't obvious to me that the ID was required for tracks in the queue. 
Which part of the code actually triggered the loop? That's also not obvious to me.
Can we improve the code triggering the loop in any way so that next time this happens we at least have some pointer as to why it happens or even break the loop?

As to how we proceed with the PR: should I now remove the unit test or rather leave them in place? 
I understand that you prefer testing the behavior of the components, but IMO unit tests have the advantage of making the code more robust and organized (otherwise it can't be tested).

Let me know what you decide, and also if you see anything else which needs to be done.

nukeop:
I went ahead and merged it, no harm in having more thorough test coverage. Thanks for contributing this.

This was happening because the action+reducer combo updates the tracks by uuid. No uuid, no update. Normally I just reuse the same state from `storeBuilders.ts`, which already includes uuids for the starting queue. Instead of initializing a custom state in tests that require it completely from scratch, a cleaner solution (that I use at work) would be to provide a set of builders that initialize all of the required fields for a given structure, and allow overwriting them.

sferra:
I'm afraid that we were too much in a hurry with merging.
We can change the "to be" in this line to any numeric value and the test doesn't fail as `waitFor` is async:
```typescript
 waitFor(() => expect(state.queue.queueItems.length).toBe(1));
```
I'm afraid that the test doesn't actually test anything. Please correct me if i'm wrong.

nukeop:
Though when I'm debugging this, all the actions are dispatched correctly and in order, the track should get removed from the queue.

nukeop:
Ok, this was pretty easy, the store needs to be reloaded after each check.

sferra:
When debugging we usually introduce delays through the debugger and breakpoints and such affect the duration of async operations. It's just a thought.

At least we have the unit test, which tells us that the removal logic is working.

I'm curious as to how the test can be fixed as I have no idea on how to proceed (I'm completely new to TS, React and this whole ecosystem).

sferra:
> Ok, this was pretty easy, the store needs to be reloaded after each check.

 Makes sense, when you think of it. : )

nukeop:
I pushed it directly to master, so if you're interested what it looks like, you can see it there.

Issue #1427, opened 2023-03-15
Display of empty artists in Discogs


maboroshin:
**Platform:**

nuclear-fd06a3, Windows 10

**Description of the issue:**

Display of empty artists in Discogs.

1. Select Discogs and enter the following search terms : "Delta 9kid" or "Gplants" or "舐達麻"
2. Select an artist and a blank screen will appear.

I have searched for artists that contain Kanji characters. But, This cannot be reproduced with other search terms.

**Expected correct behavior**:
Displaying Information : Unable to retrieve artist information.

nukeop:
Artist pages for reference:
https://www.discogs.com/artist/7673282-Delta9kid
https://www.discogs.com/artist/7673280-Gplants
https://www.discogs.com/artist/7347076-%E8%88%90%E9%81%94%E9%BA%BB

While the first two only have a single record each, the third one has more, but there must be something on the artist page that's missing data we expect to be there normally.

sferra:
I could fix this and create a PR if anyone is interested.

nukeop:
Okay, what's the plan?

sferra:
I checked, and the issue is caused because of missing artist information from Lastfm. The code accesses the data nevertheless. I'll commit a quick fix so we have something concrete to look at, then improve and add test later.

sferra:
This issue is fixed.

PR #1554, opened 2024-02-16
Fix artist search when lastfm information isn't available

sferra:
Addresses #1427

### DIFF
--- a/packages/core/src/plugins/meta/discogs.test.ts
+++ b/packages/core/src/plugins/meta/discogs.test.ts
@@ -1,0 +1,164 @@
+import DiscogsMetaProvider from './discogs';
+import { SimilarArtist } from '../plugins.types';
+import * as SpotifyApi from '../../rest/Spotify';
+
+describe('Tests for DiscogsMetaProvider', () => {
+  const provider = new DiscogsMetaProvider();
+
+  describe('Verify if an artist is on tour', () => {
+
+    test('Should return false if the artist info is missing', () => {
+      expect(provider.isArtistOnTour(undefined)).toBeFalsy();
+    });
+
+    test('Should return false if the \'ontour\' flag is not \'1\'', () => {
+      const artistInfo: any = {
+        ontour: 'something else than 1'
+      };
+      expect(provider.isArtistOnTour(artistInfo)).toBeFalsy();
+    });
+
+    test('Should return false if the \'ontour\' flag is \'1\'', () => {
+      const artistInfo: any = {
+        ontour: '1'
+      };
+      expect(provider.isArtistOnTour(artistInfo)).toBeTruthy();
+    });
+  });
+
+  describe('Get similar artists', () => {
+    const spotifyToken = 'spotify token';
+    const getToken = jest.spyOn(SpotifyApi, 'getToken');
+    const fetchTopSimilarArtistsFromSpotify = jest.spyOn(provider, 'fetchTopSimilarArtistsFromSpotify');
+
+    afterEach(() => {
+      getToken.mockReset();
+      fetchTopSimilarArtistsFromSpotify.mockReset();
+    });
+
+    afterAll(() => {
+      getToken.mockRestore();
+      fetchTopSimilarArtistsFromSpotify.mockRestore();
+    });
+
+    test('Should return an empty result set if the artist info is missing', async () => {
+      expect(await provider.getSimilarArtists({
+        similar: {
+          artist: null
+        }
+      } as any)).toEqual([]);
+    });
+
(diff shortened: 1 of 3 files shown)


sferra:
I just added a few checks and extracted the some of the existing logic to a separate function to allow testing it later.

nukeop:
Should I start reviewing? Or do you want to mark it as ready later one when you finish?

sferra:
I did my best to add tests for the my changes but I had to refactor.
Any feedback is appreciated, since I have no prior experience with TypeScript.

packages/core/src/plugins/meta/discogs.ts
```diff
@@ -210,6 +201,51 @@ class DiscogsMetaProvider extends MetaProvider {
     };
   }
 
+  async getSimilarArtists(artist: LastFmArtistInfo | undefined): Promise<SimilarArtist[]> {
+    if (!artist?.similar?.artist) {
+      return Promise.resolve([]);
+    }
+    const similarArtists = artist.similar.artist;
+    return getToken()
```
nukeop, line 209: Instead of using `.then` and `.catch`, you can just use `await` and try-catch blocks here and in the functions below.

packages/core/src/plugins/meta/discogs.test.ts
```diff
@@ -0,0 +1,164 @@
+import DiscogsMetaProvider from './discogs';
+import { SimilarArtist } from '../plugins.types';
+import * as SpotifyApi from '../../rest/Spotify';
+
+describe('Tests for DiscogsMetaProvider', () => {
+  const provider = new DiscogsMetaProvider();
+
+  describe('Verify if an artist is on tour', () => {
```
nukeop, line 8: I don't think it's needed to unit test small functions like this. Integration tests are more comprehensive, and this function is already included in the output of `fetchArtistDetails`.

nukeop (changes requested):

nukeop:
Ok, I finished my review. In general, I appreciate that you tested everything very comprehensively.

packages/core/src/plugins/meta/discogs.test.ts
```diff
@@ -0,0 +1,164 @@
...
+      getToken.mockRestore();
+      fetchTopSimilarArtistsFromSpotify.mockRestore();
+    });
+
+    test('Should return an empty result set if the artist info is missing', async () => {
+      expect(await provider.getSimilarArtists({
+        similar: {
+          artist: null
+        }
+      } as any)).toEqual([]);
```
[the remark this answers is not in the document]
sferra: It added it to allow passing the object as an argument and to avoid defining unnecessary properties. I'll change it to ` as LastFmArtistInfo` . Thanks for pointing it out.

packages/core/src/plugins/meta/discogs.test.ts
sferra: I tested only the code which I added or had to touch. The expression wrapped in the function was one of the things which caused the issue because of the missing ` lastFmInfo`. So it's most certainly not yet covered by any existing test.

Should I remove the function, the tests and inline the expression? Then my change won't be tested. : )

packages/core/src/plugins/meta/discogs.ts
sferra: I know. I had both versions locally and found the chaining of `then`/`catch` to be more consistent with rest of the code. I don't mind changing it if you like the alternative better.

packages/core/src/plugins/meta/discogs.test.ts
```diff
@@ -0,0 +1,164 @@
...
+    test('Should fetch similar artist', async () => {
+      const similarArtistsOnLastFm = [{
+        name: 'Similar Artist on LastFm'
+      }];
+      const artistInfoFromLastFm = {
+        name: 'Artist Name',
+        similar: {
+          artist: similarArtistsOnLastFm
+        }
+      } as any;
```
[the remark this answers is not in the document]
sferra: Thanks for pointing the issues out. The actual type is `LastFmArtistInfo` which has multiple fields which are not relevant to the call and I didn't want to declare. `satisfies` doesn't seem to be appropriate here, or I don't know how to apply it. I addressed all `as any` casts with explicit types. Is that allright?

sferra:
Side note: 

I just noticed that when using the Bandcamp metadata provider,  the UI doesn't display proper images for similar artists. This is because there's no logic for fetching artist images from Spotify, like the Discogs metadata provider does. If this PR gets approved, afterwards it would be possible to address this issue by extracting and reusing the current implementation for fetching artist images for the Bandcamp metadata provider.

Issue #1544, opened 2024-01-30
Play/Pause button not centred


ghost:
**Platform:** Linux

**Nuclear version:** v0.6.30

**Description of the issue:**
The Play/Pause button at the bottom of the window isn't centred between the Previous and Next buttons - it's closer to the Next button (i.e. to the right). Hard to see for the Play button, but easier with the Pause button:

![button_alignment_play](https://github.com/nukeop/nuclear/assets/2385329/46b676f9-cd4f-4fbf-b287-c32a796c470e)

![button_alignment_pause](https://github.com/nukeop/nuclear/assets/2385329/62cb4470-0618-42c4-8209-0bda082195c4)

nukeop:
Hey, this is a tricky one - it needs some extra alignment to be perfectly centered beyond what flexbox does automatically. Did you happen to measure the pixels?

BWPanda:
Yep:

![image](https://github.com/nukeop/nuclear/assets/2385329/e97ed9b1-722a-40f9-9df2-1d38d403f0b8)

Left: 23px
Right: 19px

![image](https://github.com/nukeop/nuclear/assets/2385329/421a14a1-c212-4958-9940-39857bef0623)

Left: 26px
Right: 19px

![image](https://github.com/nukeop/nuclear/assets/2385329/63208b5d-af31-4bf4-92b5-410bcd81e8c5)

Left: 26px
Right: 20px

nukeop:
You're right! This must be because we center the bounding box of the icon, not the actual icon (visually). This can be solved with adding a small margin to one side of the icon, different for each type of icon.

sqallFF:
Hello there i was wondering if anyone fixed this issue yet? I would like to work on this

nukeop:
@sqallFF thanks, go ahead!

PR #1556, opened 2024-02-20
#1544 Fix: Play/Pause Button Centered

amanbhardwaj12072003:
- **Fix:** [#1544](https://github.com/nukeop/nuclear/issues/1544)

- **Bug:**  _The Play/Pause button at the bottom of the window isn't centered between the Previous and Next buttons - it's closer to the Next button (i.e. to the right). Hard to see for the Play button, but easier with the Pause button:_

- **Change:** _Added margin to the right of the button, as suggested by @nukeop in his [comment](https://github.com/nukeop/nuclear/issues/1544#issuecomment-1916591290) regarding the issue._


nukeop:
I agree with Nuki, don't commit your package-lock.json, and don't use `style`. Instead, you should add a new class to the scss, and apply it. Don't use `data-test` for applying styles like Nuki suggets though, that's wrong.

amanbhardwaj12072003:
Hi @nukeop, can you please approve my PR.

packages/ui/lib/components/PlayerButton/index.tsx
```diff
@@ -22,18 +22,21 @@ const PlayerButton: React.FC<PlayerButtonProps> = ({
...
-  </button>
-);
+}) => {
+  return (
+    <button
+      data-testid={rest['data-testid']}
+      className={cx(
+        styles.player_button,
+        { [styles.disabled]: disabled },
+        { [styles.player_button_with_margin]: rest['data-testid'] === 'player-controls-play' } // Apply the margin conditionally based on test-id if needed
```
nukeop, line 32: That's not the right way, data-testid is only for tests, it should not influence business logic in any way.

Issue #1560, opened 2024-02-25
Playback hangs indefinitely when no streams are available for tracks


sferra:
**Platform:**
Linux

**Nuclear version:**
0.6.30, 1ec000acaa3926a52119b35bfaf0e1b091dfe5c4

**Description of the issue:**
The playback hangs indefinitely when no streams are available for a track.

To replicate:
1. Select 'Bandcamp' as stream provider
2. Search for "Monkey3", select the album "[Welcome to the Machine](https://monkey-3.bandcamp.com/album/welcome-to-the-machine)". The album has 5 tracks, of which only 2 (tracks 2, 4) can be streamed.
3. Select "Play" for the album. Playback hangs indefinitely because no stream can be resolved for the first track. Attempting to skip the  track also results in an uncaught runtime error: "The play() request was interrupted by a call to pause(). https://goo.gl/LdLk22"

The above can occur for any stream provider, not just Bandcamp, if no stream is available.

nukeop:
This should instead remove the affected track from the queue, and I'm pretty sure that in does, when the request to the streaming service results in a 404

sferra:
Yet the track isn't removed. It can be easily reproduced with the steps above.
There is no 404, since the stream itself is `undefined`.

IMO the unavailable stream should be removed and if it's the last one then the track should be probably removed, at least as a first fix.  It would be nicer IMO to mark the track as unavailable and skip to the next one, instead of letting it magically disappear.

I have a local fix for removing the unavailable stream, if you are willing to reproduce & review.

nukeop:
Fixed by #1562

PR #1562, opened 2024-02-29
Fix hanging playback when no stream is available

sferra:
Addresses #1560.

I can add test coverage for the new code if this solution is generally accepted.


sferra:
Sorry it's taking so long, but I don't have much time and have quite some issues getting the tests green, even if the change (coincidentally) works in practice. Failing to resolve stream URLs causes an avalanche of updates to the queue, also the current logic for searching streams is hard to test and possibly not adequate for such cases. I might have to make additional changes, a thing which I initially tried to avoid. I would be grateful for any advice on how to move this PR forward, @nukeop .

nukeop:
No problem, maybe it'll be easier if I do it for you?

sferra:
I don't mind if you take over. 

I assumed that `PlayerBarContainer` hook `useStreamLookup` should to be called each time an invalid track stream is removed.  This might have been incorrect.

Here's what I found so far:
* In the `PlayerBarContainer` hooks, `useStreamLookup`, there's a check which prevents state propagation (`!isStreamLoading`) when removing invalid track streams (the reason the test is currently failing).
* Unconditionally dispatching "track is loading" events while removing invalid track streams creates event loops. Adding a simple check for not re-dispatching "track is loading" prevents the loops.
* We should probably not attempt to resolve the tracks again if we remove an invalid stream.

Please let me know what you find out, even if it's not the final solution.

sferra:
Here's my local WIP state if you are curious and want to take a look: https://github.com/sferra/nuclear/pull/1

BTW, local tests for the app module seem to run indefinitely, even for "master". I don't know why. Do you have a hint?

nukeop:
Ok, please go ahead. And I might add some sort of a check for missing fields so that these actions fail instead of hanging.

nukeop:
I see that you added me to your fork as a collaborator. But actually, you only have to open a PR here, and Github will let me push directly to your branch (unless you explicitly forbid it). I'll take a closer look at this today.

sferra:
Thanks, for the feedback and assistance. There's no hurry from my side. 
I'm slow to progress these PRs myself, since I have so little time.

Regarding the fork, I added you as a collaborator to the fork so you can comment and review the linked "[WIP PR](https://github.com/sferra/nuclear/pull/1)" more easily. The PR (which is now merged into this one) was within the fork, against the  branch of this PR so it contained just the diff to this PR. I just wanted to provide you with a convenient way to verify those changes.

Regarding this PR, I suppose that I should add unit test coverage to the functions I extracted, since the coverage didn't increase with this PR. The component tests cover just specific use cases and it's not possible to cover all combination of execution paths with that type of test. I understand that you prefer the component tests, but IMO it's better to have both. Unit tests help a lot in preventing regressions when making bigger changes or refactoring. They also keep the code tested as it moves between the use cases. And I don't mind writing them. : )

nukeop:
I'm fine with this amount of coverage to be honest. The functionality you added is covered. Do you want to add anything else, or can this be merged?

sferra:
You can merge if you feel it's safe enough. Thanks.

nukeop:
Awesome, thanks for contributing.

Issue #1476, opened 2023-07-13
Full screen visualizer opacity


NicolasGuruphat:
**Platform:**
Windows
**Nuclear version:**
v0.6.24
**Description of the issue:**
When the visualizer is in full screen, there is a grey effect over the image
![image](https://github.com/nukeop/nuclear/assets/82589938/18708fd4-0bd1-45a1-b6c5-7e99fef13ec6)

nukeop:
You're right that this is a bug, I've never considered this because when using the visualizer I just move the cursor to another monitor.

This needs to be fixed in the following way:
- Remove the hover css styles from the overlay element
- Instead, use setTimeout and add a mouse event listener that checks if the mouse has moved over this element. If it moved, set a timer for 3-5s, and apply the dark overlay
- When the timer expires, remove the overlay again
- Can probably be done using `useEffect`

lalitkumawat1m:
I want to work on this issue could you please assign it to me ?

nukeop:
Sure, please also check this: https://nukeop.gitbook.io/nuclear/contributing/contribution-guidelines

lalitkumawat1m:
Im facing issue in installing project my computer is showing that it is harmful for system

**ScreenShot**
<img width="960" alt="Screenshot 2023-07-15 205842" src="https://github.com/nukeop/nuclear/assets/91591901/d0976b38-82a9-4d2f-b04f-7c6ea3974fe6">

I have downloaded nuclear.Setup.0.6.24.exe file from https://github.com/nukeop/nuclear/releases
could you please tell me how to install it and what are the steps to reproduce this bug

nukeop:
Looks like a Windows problem

NicolasGuruphat:
Yes it is. You can force the installation by clicking more info

NicolasGuruphat:
Are you still working on this @lalitkumawat1m  ? I may try to fix it this weekend.

lalitkumawat1m:
> Are you still working on this @lalitkumawat1m  ? I may try to fix it this weekend. 

You can work on this issue

Bhuminandan:
@nukeop can you assign this to me?

nukeop:
No

Bhuminandan:
@nukeop Is someone already working on it?

nukeop:
No, but I don't see any Typescript code in your contributions, and I am no longer accepting contributions from random Github accounts.

Bhuminandan:
[...]

PR #1563, opened 2024-03-02
resolve nuclear "Full screen visualizer opacity #1476"

HaraldFair:
Please review my pull request to resolve issue #1476.

The linter has been run.

Testing has been run.

I spend significant time looking into / writing a test to support this new functionality but it seems that since Jest runs in NodeJS it doesn't support CSS properties except under artificial (inline styling) conditions that would be inappropriate for this feature.

The existing snapshot tests complete with no errors.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,6 @@
     "@typescript-eslint/eslint-plugin": "^4.33.0",
     "@typescript-eslint/parser": "^4.33.0",
     "babel-eslint": "^10.0.3",
-    "classnames": "^2.5.1",
     "cross-env": "^7.0.3",
     "electron": "^12.1.0",
     "electron-builder": "^21.2.0",
--- a/package.json
+++ b/package.json
@@ -38,6 +38,7 @@
     "@typescript-eslint/eslint-plugin": "^4.33.0",
     "@typescript-eslint/parser": "^4.33.0",
     "babel-eslint": "^10.0.3",
+    "classnames": "^2.5.1",
     "cross-env": "^7.0.3",
     "electron": "^12.1.0",
     "electron-builder": "^21.2.0",
--- a/packages/ui/lib/components/VisualizerOverlay/index.tsx
+++ b/packages/ui/lib/components/VisualizerOverlay/index.tsx
@@ -1,4 +1,4 @@
-import React, { SyntheticEvent } from 'react';
+import React, { useState, useEffect, SyntheticEvent } from 'react';
 
 import { Button, Dropdown } from '../..';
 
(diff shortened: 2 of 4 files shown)


packages/ui/lib/components/VisualizerOverlay/index.tsx
```diff
@@ -21,32 +21,56 @@ const VisualizerOverlay: React.FC<VisualizerOverlayProps> = ({
   exitFullscreenLabel,
   isFullscreen = false
 }) => {
-  const presetOptions = presets.map(preset => ({
+  // State for managing overlay opacity
+  const [isHovered, setIsHovered] = useState(false);
+
+  // Establish timer for overlay opacity
+  useEffect(() => {
+    let timeoutId: NodeJS.Timeout;
```
nukeop, line 29: Nuki is right that you should use the browser-side type here

packages/ui/lib/components/VisualizerOverlay/index.tsx
```diff
@@ -21,32 +21,56 @@ const VisualizerOverlay: React.FC<VisualizerOverlayProps> = ({
...
-            basic
-            icon='expand'
-            onClick={onEnterFullscreen} 
-          />
-        )
-    }
-  </div>;
+  return (
+    <div
+      className={`${styles.visualizer_overlay}${
```
nukeop, line 51: Use the `classnames` package to combine classes. It can also handle conditions

nukeop (changes requested):
Thanks, this is a pretty good way to solve it. I only have 2 minor comments. Applying Nuki's changes is not mandatory!

packages/ui/lib/components/VisualizerOverlay/index.tsx
HaraldFair: Thank you, I will make the requested changes and update my PR.

package.json
```diff
@@ -38,6 +38,7 @@
     "@typescript-eslint/eslint-plugin": "^4.33.0",
     "@typescript-eslint/parser": "^4.33.0",
     "babel-eslint": "^10.0.3",
+    "classnames": "^2.5.1",
```
nukeop, line 41: classnames is already installed in projects that need it. You don't have to install it, so it should be removed here.

packages/ui/lib/components/VisualizerOverlay/index.tsx
```diff
@@ -21,32 +23,65 @@ const VisualizerOverlay: React.FC<VisualizerOverlayProps> = ({
...
+  const stylesMod = classNames.bind(styles);
+
+  const ovClasses = stylesMod({
+    visualizer_overlay: true,
+    hover: isHovered
+  });
+
+  return (
+    <div
+      className={ovClasses}
```
nukeop, line 64:
```suggestion
      className={classnames(styles.visualizer_overlay, {
          [styles.hover]: isHovered
      })}
```

And then you can remove lines 55-60.

I also recommend importing `classnames` as `cx` to make it shorter.

nukeop (changes requested):

package.json
nukeop: Also, as there are no further changes to packages, you should remove the changes you've made to package-lock.json from the PR (but don't delete the file altogether).

HaraldFair:
OK, I think this is compliant with the requested changes now. I've re-run tests and lint:fix and manually verified functionality after all actions.

PR #1566, opened 2024-03-11
Fix Mouse back button not doing anything

### LINKED ISSUE #1565: Mouse Back Button does not do anything
**Platform:** Arch Linux 6.7.6-arch1-2

**Nuclear version:** 0.6.30

**Description of the issue:**
Using the back button on the mouse does not do anything. I would have expected it to do the same thing as the gui back button. Same for the forwards button.

Steps:
1. Have a mouse with a forward and back button
2. Open Nuclear and start navigating around the UI
3. Click mouse back button (doesn't do anything)
4. Click the gui back button (does go back to the last place you were)
5. Do the same for the forwards button

hollow-owl:
Fixes #1565 

react-router is controlling the history so electrons [`win.webContents.goBack()`](https://www.electronjs.org/docs/latest/api/browser-window#event-app-command-windows-linux) does not do anything. 

The code handling `navigate-back` and `navigate-forward` can go on pretty much any component that is always mounted. I didn't know whether to put it in App or NavButtons. I ended up with App instead of NavButtons just in case there is a UI where NavButtons are not enabled.

### DIFF
--- a/packages/app/app/App.js
+++ b/packages/app/app/App.js
@@ -48,15 +48,11 @@ import WindowControls from './components/WindowControls';
 import SidebarMenuContainer from './containers/SidebarMenuContainer';
 import { CommandPaletteContainer } from './containers/CommandPaletteContainer';
 import { hot } from 'react-hot-loader';
-import { ipcRenderer } from 'electron';
 
 @withTranslation('app')
 class App extends React.PureComponent {
   constructor(props) {
     super(props);
-
-    this.handleBack = this.handleBack.bind(this);
-    this.handleForward = this.handleForward.bind(this);
   }
 
   componentDidMount() {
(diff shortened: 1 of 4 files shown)


packages/app/app/App.js
```diff
@@ -68,6 +72,26 @@ class App extends React.PureComponent {
     this.updateConnectivityStatus(navigator.onLine);
     window.addEventListener('online', () => this.updateConnectivityStatus(true));
     window.addEventListener('offline', () => this.updateConnectivityStatus(false));
+
+    ipcRenderer.on('navigate-back', this.handleBack);
```
nukeop, line 76: Could you define these events in `@nuclear/core` in the IpcEvents enum?

nukeop: And a better place for this is the IpcContainer

nukeop (changes requested):

packages/app/app/App.js
hollow-owl: Moved it to IpcContainer and defined the events in IpcEvents

nukeop:
Thanks, I merged your PR.

Issue #1511, opened 2023-09-21
Dragging/adding multiple songs / ful albums into a playlist


NicTanghe:
Hello fist 2 minutes on this app, one thing that bothered me on this amazing app.

That is, that adding songs to a playlist is a complete hassle. Having to look for a button for each individual song and then select add to playlist is a hassle by far.

nukeop:
I admit this is pretty hard to do. I'm thinking the option to do so could be added to the menu that pops up when you've got multiple tracks selected in pretty much any list of tracks:

<img width="539" alt="image" src="https://github.com/nukeop/nuclear/assets/12746779/b7608edd-337b-4804-aa89-ea5d30f53684">

sferra:
I have a prototype for adding full albums to playlists.
If there's any interest in this functionality, then I could polish it and create a PR.
![image](https://github.com/nukeop/nuclear/assets/596554/d09d8b35-8cd1-4483-9fde-df5146f7f449)

nukeop:
That's a natural place to add this functionality. If you can, please open a PR.

nukeop:
This issue is from an older version of Nuclear that has been replaced with a complete rewrite. None of the old issues apply to the new codebase.

The old code is preserved on the [legacy/electron](https://github.com/nukeop/nuclear/tree/legacy/electron) branch. If you have feedback about the new version, head to [Discussions](https://github.com/nukeop/nuclear/discussions).

PR #1568, opened 2024-03-17
Enable adding albums to playlists

sferra:
Added an "Add album to playlist" entry to albums' menu.
Allows adding the album tracks to either an existing or a new playlist.

See #1511

### DIFF
--- a/packages/app/app/components/AlbumView/index.tsx
+++ b/packages/app/app/components/AlbumView/index.tsx
@@ -1,9 +1,9 @@
 import React from 'react';
 import Img from 'react-image';
 import _ from 'lodash';
-import { Dimmer, Icon, Loader } from 'semantic-ui-react';
+import { Dimmer, Icon, Loader, Dropdown } from 'semantic-ui-react';
 import { useTranslation } from 'react-i18next';
-import { Loader as NuclearLoader, ContextPopup, PopupButton } from '@nuclear/ui';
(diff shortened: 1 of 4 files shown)


nukeop:
Thanks for adding this, and for the test coverage. I'm not sure if that's the best way to handle it here (the track popup is at least self-contained), maybe it would be better to make that whole menu its own component, with the input dialog and whatnot. For this PR, it's alright, so I'll merge it once the CI passes.

sferra:
I thought about extracting the pop-up to a separate component, but I decided to first add the functionality and leave the refactoring decision to you.
How do we address code base improvement and cleanup/refactoring? Are there specific issues (types) in the tracker for this type of task?

nukeop:
Not really, I do that whenever I feel like, or when I'm working on a large feature (like the local library scanner) and keep seeing untidy code. Sometimes instead of working on features I just look for stuff I neglected to do earlier and work on it. If I do any refactoring though, I make sure that area has decent test coverage to avoid regressions. Any such work is welcome but I realize it can be boring by itself, so usually I do it myself.

sferra:
So would you accept refactoring PRs? I usually try to strictly separate refactoring from adding new features, to avoid regressions, specially when I'm not very familiar with the code base.

nukeop:
Sure, I would be very glad to see that. Typically I'm the only one doing that.

PR #1571, opened 2024-03-19
Fix back-navigation after deleting a playlist

### LINKED ISSUE #1570: Navigation issues after deleting a playlist
**Platform:**
Linux

**Nuclear version:**
0.6.30 80a7b1c97b9380c7b58b67a4b14c907f9739ca67 

**Description of the issue:**
Back navigation triggers an error and behaves incorrectly after deleting a playlist.

After deleting a playlist and navigating back (using the navigation buttons, for example), an error is displayed. Also, the playlists overview is displayed instead of actually navigating back to the previous view.

To replicate:
1. Create a new playlist (name and contents are irrelevant)
2. Select the playlist. The playlist is displayed.
3. Delete the playlist using the option in the '...' menu.
4. The playlists overview is displayed
5. Attempt to navigate back using the navigation buttons in the UI. An error is displayed, the playlists overview is displayed instead of navigating back to the previous view.
6. At this point we can also attempt to navigate forward, which will trigger the same error once more

The issue is caused by the fact that after deleting a playlist, the current implementations actually navigates forward to the playlists overview instead of going back to the previous view. Therefore navigating back attempts to display the playlist we just deleted.

`PlaylistView.onDeletePlaylist`:
```ts
deletePlaylist(playlist.id);
history.push('/playlists');
```
should rather be:
```ts
deletePlaylist(playlist.id);
if (history.length > 1) {
  // The current playlist is the top of the navigation stack, go back to the previous view
  history.goBack();
} else {
  // Fallback in case we can't navigate back to a previous view
  history.push('/playlists');  
}
```

This works nicely, without having to touch existing component tests, nevertheless, the new (and relevant) execution path wouldn't be tested.

sferra:
Resolves #1570.

### DIFF
--- a/packages/app/app/components/PlaylistView/index.tsx
+++ b/packages/app/app/components/PlaylistView/index.tsx
@@ -1,5 +1,4 @@
 import React, { useCallback } from 'react';
-import _ from 'lodash';
 import { TFunction } from 'i18next';
 import { useTranslation } from 'react-i18next';
 import { useHistory } from 'react-router';
@@ -72,7 +71,13 @@ const PlaylistView: React.FC<PlaylistViewProps> = ({
 
   const onDeletePlaylist = useCallback(() => {
     deletePlaylist(playlist.id);
-    history.push('/playlists');
+    if (history.length > 1) {
+      // The current playlist is the top of the navigation stack, go back to the previous view
+      history.goBack();
+    } else {
+      // Fallback in case we can't navigate back to a previous view
+      history.push('/playlists');
+    }
   }, [playlist, history, deletePlaylist]);
 
   const onExportPlaylist = useCallback(() => {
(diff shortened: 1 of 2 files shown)


sferra:
@nukeop, could you please assist with the test setup? We need to access the actual navigation history in order to assert that the navigation is performed.

nukeop:
I think that was the right way to do it, why did you revert it?

nukeop:
I think I know how to do it, I've got to run it and debug it a little.

sferra:
Sorry for the late reply.
> I think that was the right way to do it, why did you revert it?

I was very tired and the test code landed in the incorrect test (track deletion, not playlist deletion). Also, history expectations didn't work as expected (I assume because of the functional nature of the react) and I had no clue how to get access to the actual history. Feel free to use whatever you need from that commit. :-)

nukeop:
Think I'll just merge it.

sferra:
Thanks for your time and trust. It would be nevertheless good to learn how to test that specific kind of functionality in the future. Please notify me if/when you ever have or see such a PR.

nukeop:
I couldn't figure out a good way to do it, and since it's such a small change and looks alright I decided it's ok to forgo it this time.

PR #1578, opened 2024-04-04
Fix 1576 : Play counts not aligned (artist popular tracks)

### LINKED ISSUE #1576: Play counts not aligned
**Platform:**
Mac OS

**Nuclear version:**

**Description of the issue:**

- Play count is not aligned with the header for both regular or expanded view

Expected Result: The play count should be aligned
Actual: The play count is not aligned with the header

image
image

gurvanhenry:
Great project 👏 ! 
I start with a simple PR but I'd like to do more and discover the codebase.

--

Remove text-align:right on popular_tracks_table header to fix vertical alignment with table-rows

Fix #1576 

Here is what I have changed :
image


nukeop:
Thanks!

PR #1580, opened 2024-04-08
Fix #1503: Fix skip track buttons on mac devices

### LINKED ISSUE #1503: Skip track keyboard buttons do not work
**Platform:** ubuntu 22.04

**Nuclear version:** 2023 (august)

**Description of the issue:**
The usual keyboard buttons to switch to the next track or previous (or play/pause) do not work.
Instead I'm forced to right click the tray icon to skip a track, pretty non-efficient.

ajkapala21:
In issue #1503 the the issuer noted that the usual keyboard buttons to switch to the next track or previous (or play/pause) do not work. The buttons they are speaking of are shown here https://www.macworld.com/wp-content/uploads/2023/01/mac-keyboard-play-pause-100837087-orig.jpg?quality=50&strip=all&w=1024. From searching online it appears that all apple keyboards include these symbols to suggest that the f7, f8, and f9 keys have functionalities to switch to the previous track, play/pause, and switch the next track respectively.

This change adds these shortcuts for only mac devices, since this functionality is only suggested prominently on mac devices.

### DIFF
--- a/packages/app/app/containers/ShortcutsContainer/ShortcutsContainer.test.tsx
+++ b/packages/app/app/containers/ShortcutsContainer/ShortcutsContainer.test.tsx
@@ -1,0 +1,108 @@
+import { fireEvent} from '@testing-library/react';
+import React from 'react';
+
+import ShortcutsContainer from '.';
+import { isMac } from '../../hooks/usePlatform';
+import { AnyProps, setupI18Next, mountComponent } from '../../../test/testUtils';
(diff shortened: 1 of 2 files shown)


nukeop:
Yeah, I agree with Nuki - `isMac` function should be mocked and both cases should be covered in tests.

ajkapala21:
Just committed the changes for the tests to follow your recommendation.

PR #1586, opened 2024-04-12
Fix/1585: "Popular Tracks" is not translated

ajkapala21:
Added Popular Tracks to the en.json file and changed the PopularTracks component to use the translation.


nukeop:
Thanks!

PR #1589, opened 2024-04-13
Fix #1554: Center Play/Pause Button

### LINKED ISSUE #1544: Play/Pause button not centred

ajkapala21:
Fixes #1544
Adds a right margin of 4px to the player button class which successfully centers each icon (pause, play, or loading). Although a previous comment highlighted that each icon was off center by a varying amount, from my inspection they are all off by the same amount, so the same margin can be applied to all of them.

### DIFF
--- a/packages/ui/lib/components/PlayerButton/styles.scss
+++ b/packages/ui/lib/components/PlayerButton/styles.scss
@@ -3,6 +3,7 @@
   border: none;
   cursor: pointer;
   outline: none;
+  margin-right: 4px;
 
   &.disabled {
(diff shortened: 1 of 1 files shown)


nukeop:
Nice, thank you.

PR #1590, opened 2024-04-13
Converted HelpButton component into TypeScript

sicin:
Small rework. As seen in other button examples, a prop allowing for future extension of the button was added, the rest stayed as is.


packages/app/app/components/HelpButton/index.tsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import { Button, Icon } from 'semantic-ui-react';
+
+import styles from './styles.scss';
+
+type HelpButtonProps = {
+  onClick: () => void;
+  [x: string]: any;
```
nukeop, line 8: I'd like to avoid that `any`. This should only allow valid props of `Button`.

PR #1594, opened 2024-04-18
Converted NavButtons component into TypeScript

sicin:
Changed the NavButtons component into TypeScript. Index is apparently not a valid prop of History, altough it obviously exists on the object, that's why the 'as' keyword was used.


sicin:
After some thinking I've realized that Nuki-chan is right. PR fixed.

nukeop:
Thanks, this looks pretty good. Int he future instead of force pushing there's nothing wrong with just committing over your original work so we can see what happened at each step.

Issue #1367, opened 2022-11-14
[Listening history] Add charts with recent history


nukeop:
This issue can be taken by anyone.

Now that listening history is being stored, we can start adding charts. There's a million libraries that allow drawing charts in JS and React so it's up to the implementer which one to use.

Helpful charts could include: recent artists, recent genres, or most listened to tracks. The charts can be pie charts, bar charts, or something else.

If you want to take this issue, make a comment here and say what chart you'd like to add, and roughly how you're planning to do it. You should also write some sensible tests for this.

rutviklhase:
If there's no one working on this, I would love to take this one up! :)

ChiamakaUI:
Hi @nukeop, I would love to work on this issue. Can it be assigned to me ?

nukeop:
Okay, go ahead, let me know how it goes

alicemmartinez:
Hi! Was wondering if this was ever resolved and if not, I would love to work on it!

nukeop:
It wasn't, you can go ahead, just let me know what you're planning to do before you start.

alicemmartinez:
Hi! Sorry for the late response. I was planning on adding some graphs to the listening history section. More specifically, I was thinking of doing a pie graph that shows how often one listens to a specifc artist. Additionally, I was going to add perhaps a bar graph to represent recent genres. Lastly, some sort of listed chart that shows the top 3 most listened-to songs. I started this process by creating a new subcomponent of ListeningHistoryView known as ListeningHistoryGraphs.

nukeop:
Alright, so let's say 3 charts for now (though even 1 would be a nice start, more can always be added later):
- A pie chart that shows the top X (perhaps 10) most listened to artists in a time unit
- A bar graph for the recent genres. Some kind of a bubble graph might be nice to visualize this, if you're up to the challenge.
- A list + chart (like last.fm) that shows the top X songs in a time unit. Maybe more than 3.

Also both charts should have dropdowns that let you select the X, and another one for the time range. Example from last.fm:

<img width="876" alt="image" src="https://github.com/nukeop/nuclear/assets/12746779/e3e316fd-90d4-46f4-a135-6b092bea25eb">

Did you select the library you want to use for drawing the charts? Last.fm uses Highcharts, but it comes with a prohibitive cost (for a free software projects). Some suggestions from me:
- https://github.com/recharts/recharts - easy to use
- https://d3js.org/ - fast and very rich, might be harder to use
[...]

PR #1599, opened 2024-04-26
#1367: Added Recent Artists Pie Chart to ListeningHistory

alicemmartinez:
Have not written test cases, or completely implemented filtering algorithm for range of dates. I was hoping you could guide me to where I could write test cases for my added subcomponent?

### DIFF
--- a/packages/app/app/components/ListeningHistoryView/PieChartTopArtists.tsx
+++ b/packages/app/app/components/ListeningHistoryView/PieChartTopArtists.tsx
@@ -1,0 +1,98 @@
+import React, { useState } from 'react';
+import {
+  PieChart,
+  Pie,
+  Tooltip,
+  Legend,
+  Cell,
+  Text,
+  ResponsiveContainer
+} from 'recharts';
+
+const colors = [
+  '#FF6633', '#FFB399', '#FF33FF', '#00B3E6', '#E6B333', '#3366E6',
+  '#999966', '#99FF99', '#B34D4D', '#80B300', '#809900', '#E6B3B3',
+  '#6680B3', '#66991A', '#FF99E6', '#CCFF1A', '#FF1A66', '#E6331A',
+  '#33FFCC', '#66994D', '#B366CC', '#4D8000', '#B33300', '#CC80CC',
+  '#66664D', '#991AFF', '#E666FF', '#4DB3FF', '#1AB399', '#E666B3',
+  '#33991A', '#CC9999', '#B3B31A', '#00E680', '#4D8066', '#809980',
+  '#E6FF80', '#1AFF33', '#999933', '#FF3380', '#CCCC00', '#66E64D',
+  '#4D80CC', '#9900B3', '#E64D66', '#4DB380', '#FF4D4D', '#99E6E6',
+  '#6666FF'
+];
+
+// Function to get the top N items from the data based on a given key
+const getTopN = (data: any[], key: string, n: number): any[] => {
+  return data?.sort((a, b) => b[key] - a[key]).slice(0, n);
+};
+
+export type PieChartTopArtists = {
+    data: any[];
+}
+
+const timeframes = ['Last 7 days', 'Last 30 days', 'Last 90 days', 'Last 180 days', 'Last 365 days', 'All time'];
+const artistCounts = [5, 10, 15, 20];
+
+export const PieChartTopArtists: React.FC<PieChartTopArtists> = ({ data = [] }) => {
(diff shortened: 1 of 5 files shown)


nukeop:
- You can try moving this component to the `ui` package, and creating a story for it (using storybook). Then we could see what it looks like in various scenarios easily.
- In that package, there's an easy setup for snapshot tests (look for a function called `makeSnapshotTest`). If the chart is rendered in svg that's all we're going to need. You can also try adding more involved unit tests for any edge cases you can think of.
- If you do it that way, then this component can be imported into the `app` package, in the `ListeningHistoryView` component, and fed data from there. No need IMO to write any additional tests there. Just update existing snapshots (`jest -u`).

packages/app/app/components/ListeningHistoryView/PieChartTopArtists.tsx
```diff
@@ -0,0 +1,98 @@
...
+  '#33FFCC', '#66994D', '#B366CC', '#4D8000', '#B33300', '#CC80CC',
+  '#66664D', '#991AFF', '#E666FF', '#4DB3FF', '#1AB399', '#E666B3',
+  '#33991A', '#CC9999', '#B3B31A', '#00E680', '#4D8066', '#809980',
+  '#E6FF80', '#1AFF33', '#999933', '#FF3380', '#CCCC00', '#66E64D',
+  '#4D80CC', '#9900B3', '#E64D66', '#4DB380', '#FF4D4D', '#99E6E6',
+  '#6666FF'
+];
+
+// Function to get the top N items from the data based on a given key
+const getTopN = (data: any[], key: string, n: number): any[] => {
```
nukeop, line 25: Make this function better by making it generic (`<T>`), instead if using `any`. That way we can ensure type safety for its input and output.

packages/app/app/components/ListeningHistoryView/PieChartTopArtists.tsx
```diff
@@ -0,0 +1,98 @@
...
+  return data?.sort((a, b) => b[key] - a[key]).slice(0, n);
+};
+
+export type PieChartTopArtists = {
+    data: any[];
+}
+
+const timeframes = ['Last 7 days', 'Last 30 days', 'Last 90 days', 'Last 180 days', 'Last 365 days', 'All time'];
```
nukeop, line 33: Nuclear is translated into more than 30 languages. So we should never use hardcoded strings like these, and instead use `react-i18next`. Try looking for the `useTranslation` hook to see how it's used.

If you decide to move this component to the `ui` package, you can simply feed it strings it needs to render its interface. For an example of how it can be done, you can look for `TrackPopupStrings`. The idea is that the component in `ui` takes a list of texts it's going to render, and in the `app` package, when we want to use that component, we look up the translated texts using the `useTranslation` hook. If you're unsure how to do this, let me know.

nukeop: And the same comment applies to other instances of raw text, like `Timeframe`, etc.

packages/app/app/components/ListeningHistoryView/PieChartTopArtists.tsx
```diff
@@ -0,0 +1,98 @@
...
+          >
+            {timeframes.map((tf) => (
+              <option key={tf} value={tf}>
+                {tf}
+              </option>
+            ))}
+          </select>
+        </label>
+
+        <label style={{ marginLeft: '20px' }}>
```
nukeop, line 59: Instead of using the `style` attribute, try using scss files and classes, it separates the styles from the markup.

packages/app/app/components/ListeningHistoryView/testSuite.js
```diff
@@ -0,0 +1,55 @@
+import React from 'react';
```
nukeop, line 1: The convention for tests is to use `<container>.test.ts`.

packages/app/app/components/ListeningHistoryView/testSuite.js
```diff
@@ -0,0 +1,55 @@
...
+const mockData = [
+  { name: 'Artist A', count: 50 },
+  { name: 'Artist B', count: 40 },
+  { name: 'Artist C', count: 30 },
+  { name: 'Artist D', count: 20 },
+  { name: 'Artist E', count: 10 }
+];
+
+describe('PieChartTopArtists', () => {
+  test('renders without crashing', () => {
```
nukeop, line 14: This can be replaced with a snapshot test, which gives us the same information, and also records what the component looks like in the HTML.

Issue #1147, opened 2022-01-23
Infinite loading pages between similar artists pages


Kritsos:
**Platform:**
Windows 10
Version: 20H2
OS Build: 19042.1466
and
Ubuntu 20.04.3

**Nuclear version:**
0.6.17

**Description of the issue:**
When going from one artist page to another through the similar artists panel the page changes just fine. However if i press the back button the page changes to an infinite loading page and i have to press the back button again to get to the starting artist page. This happens with the forward button too where beginning from the starting artist page when you press the forward button you get an infinite loading page and only when you press it again do you get the next artist page. The number of infinite loading pages is equal to the number of artists you visit, meaning there is an infinite loading page for each jump from one artist page to another through the similar artists panel. 

Steps to reproduce the bug:
1. Go to an artist page.
2. Go to another artist page from the similar artists panel.
3. Click on the back button to get an infinite loading page and click on it again to get to the starting artist page.
4. Now click the forward button to get an infinite loading page and click on it again to get to the second artist page.
5. Repeat these steps again from the second artist.

Showcase of the bug:

https://user-images.githubusercontent.com/72339454/150659466-e84abb0d-6b3c-4919-9fd2-a1baec38869d.mp4

nukeop:
Fixed by #1600

PR #1600, opened 2024-05-04
Solved issue 1147

meluchoMZ:
This solves issue https://github.com/nukeop/nuclear/issues/1147. The issue was caused by a problem managing the history object. Instead of replacing the actual loading page position by the rendered artist view, is was being pushed, causing the navigation buttons to move to infinite loading screens. 

I could not find tests for the functions that implement the search artists logic. If they exist, I am sure that you could guide me to them.


nukeop:
Awesome, thank you.

Issue #1555, opened 2024-02-17
Shuffle visualization presets with an interval


elmaester:
Manually changing presets is no fun. Currently there's a default visualization preset, and it never changes.

I propose doing a setInterval where the callback increments the index of the active preset in the array of presets.

Or achieving a similar effect in a different way.

Or perhaps playing a new track could be the trigger.

And actually, instead of incrementing and moving to the next preset, it should be chosen randomly from the available options.

nukeop:
I agree switching to a random preset whenever you start playing a new track sounds like a good idea.

elmaester:
Bumping up because I still want this ☺️

gurvanhenry:
A `shuffle button` next to preset list could also be interested for the user to discover other visualizations randomly.

Something like this:

![image](https://github.com/nukeop/nuclear/assets/2821522/573e024b-87bb-407a-9c33-f2c90e771b4d)

nukeop:
I'd rather see it to the left of the existing fullscreen button, and in the same style, but it's the right idea.

CarlosBor:
Silly as it might be, the visualizer is the kind of feature that brings a smile to my face.
I'll start working on this feature.

nukeop:
Great, additionally people have been asking about adding this on Discord.

nukeop:
Implemented!

elmaester:
Badass! Thank you! I'll update and check it out soon 🎉

Issue #1539, opened 2024-01-16
WebNowPlaying or NowPlaying Plugin Request


rickyrooroo229:
I use Rainmeter and there's no way to detect Nuclear's music on any player skins I own. Is there a way to implement this plugin myself or can you guys update Nuclear to have these plugins implemented on Nuclear?

nukeop:
I just checked and it looks like Rainmeter is Windows only, so I won't be able to develop for it. You might be able to use the REST API to get what you want. If anyone else wants to develop this, I'll be open to working on a PR.

CarlosBor:
I was hoping that it had something to do with how Windows recognizes media players, but it seems to me that it would require work on the Rainmeter side of things instead. as a matter of fact the list of supported players can be found in the documentation: https://docs.rainmeter.net/manual/measures/nowplaying/#playerlist

Either way, it seems like it could be made to work like Spotify does, which involves using https://github.com/keifufu/WebNowPlaying-Rainmeter

nukeop:
Thanks for checking this. So Spotify includes a Rainmeter integration?

Looking at the repo you linked, it'd need to be a native lib, which uses .NET, which would be complicated to say the least (unless a prebuilt binary can be included). I looked for Node integrations but there's nothing on npm. That makes it less likely to be developed.

nukeop:
This issue is from an older version of Nuclear that has been replaced with a complete rewrite. None of the old issues apply to the new codebase.

The old code is preserved on the [legacy/electron](https://github.com/nukeop/nuclear/tree/legacy/electron) branch. If you have feedback about the new version, head to [Discussions](https://github.com/nukeop/nuclear/discussions).

PR #1612, opened 2024-06-01
Shuffle visualization preset feature #1555

CarlosBor:
A couple matters:

Regarding the third point of the guidelines, "If you're fixing a bug, or adding a new functionality", this *is* a new functionality, but when checking the tests for the visualizerOverlay component there are only snapshots. I've updated them, but should I add tests on it and the randomizing function I'm using or is it understood to be simple enough to be ok as it is?

Also, as it is right now, the name of the preset is stored in the "settings" slice of the redux store, I've thus added the shuffle toggle to the same slice in a similar way, is that okay or should I change it to use a new "visualizer" slice?

Besides that, the button to toggle shuffle appears like so:
image

And the shuffle happens if enabled when a track ends.

Besides that, awaiting for revision, thanks for your patience.


nukeop:
Hi thanks for contributing this. I will do a review and comment in the files individually, but it would be great to have tests that reproduce the behavior of the user using this function, so something that mounts the visualizer, selects the shuffle mode, waits for the end of the current song to see if it's changed, etc. Basically everything you added in this PR.

nukeop:
I looked at your code and I have no additional remarks. It's all great.

CarlosBor:
Went with the pull request update rather than writing into Discord this time, should be tidier this way.

Long story short, I did manage to mock the library but I reckon that was the wrong way to go about things, I just can't get the   handleFinishedPlaying() function from the SoundContainer component to fire. I'm guessing it ultimately triggers as a response to an event fired somewhere but I'm lost regarding how to simulate the behavior and frankly I have the suspicion that there is something obvious that I'm missing.

I'd be thankful If you could take a crack at it in case it's not too much work. Otherwise I could really do with some pointers.

Thanks for your patience.

nukeop:
Hey, I'm working on fixing the Windows build first. I'll take a look as soon as I can. I'm sure we'll figure it out together somehow.

CarlosBor:
Sure thing. I use Windows so if I can offer any assistance I'm up for it, don't know much about the matter though.

In the meantime I'll take a small gander at #1539, will comment in that thread if I start taking care of it proper.

nukeop:
The day has come when I finally have the time to get around to this.

nukeop:
I managed to rewrite the tests, now everything mounts correctly, and there's a way to access the onFinishedPlaying callback to check what was passed to it. It doesn't work yet, but it's pretty close.

nukeop:
Ok, I finished the setup. It demonstrates the way you can extract a callback from a mocked React component to call it imperatively. Would you like to add anything else, or can we merge this?

Issue #1569, opened 2024-03-19
Local library: unable to map a directory to a network share


OrdoConcept:
**Platform:** Windows 10

**Nuclear version:** 0.6.30.0

**Description of the issue:**
Impossible to map a directory to a network share (a NAS for example). Why?

OrdoConcept:
Thank you for this quick response.
If I put my library on a physical disk on my machine everything works. If I put it on my NAS, even if I mount the directory in the file system, no file is detected...

CarlosBor:
I looked into this and there might be a chance the base issue was accidentally fixed? These were not exactly done with a NAS but another computer in my network which has a shared folder instead, I presume them to be analogous.

In the latest release of nuclear (0.6.30 at the time of writing) it certainly doesn't seem to detect the files. It grabs the folder fine and all, but just seems to find nothing:
![image](https://github.com/nukeop/nuclear/assets/48533334/bd01bbe7-7885-4152-9eb9-09827e3d2ace)

But if I clone the repo as it is right now and build it, then do the same thing it finds the files just fine:
![image](https://github.com/nukeop/nuclear/assets/48533334/fd57cd8c-37bc-4588-94ae-f04a0fd7a61d)

Trying to play them right then and there yields a "local.tracks.find is not a function" error, but after restarting the app the files can be played. I'll take a look at that next but if someone that actually has a NAS available could verify if the behavior above replicates that'd be nice.

nukeop:
On the master branch you would be using the new scanner, which wasn't written with network shares in mind, but behaves differently, so it is very possible that it was fixed accidentally. Thanks for verifying this!

nukeop:
The new release includes the new scanner. Try it, and if it doesn't work let me know.

PR #1620, opened 2024-06-18
Fixes issue mentioned in #1569. Unable to play file until app restart

CarlosBor:
I assumed it's not needed to make this an issue of its own but will do so if needed.

Basically, when trying to play recently added files to the queue the reducer for the local state expects an array of Track objects per a Typescript type assertion, this is important because the queue action tries to run find on this array. Problem is that when the tracks are first loaded they appear as an object with key-value pairs rather than an array:

image

This is a quick fix on that.


nukeop:
Thanks, I'm going to merge this but also test this manually later because it might be platform-dependent. On Linux and Mac I was pretty sure this already was like this.

Issue #1484, opened 2023-07-26
Nuclear doesn't download verified streams.


Saltane:
**Platform:**
Linux
**Nuclear version:**
0.6.27 (AUR)
**Description of the issue:**
Issue is quite simple. There's a track which if unverified plays undesired stream. After verifying named track to desired stream, download feature of Nuclear downloads track from undesired stream mentioned before.
**Expected behaviour:**
After verifying a track, Nuclear will download correct, verified stream.

nukeop:
Since downloading is implemented in a different place, it does not share the code that normally checks the verification service for matches. It should be added there.

Tengu-san:
Hello,
just a suggestion: to add a message when one add an unverified track to download, that it won't be download since this is not verified?
This will make the thing clearer :-)

dhananjay-Byte:
@nukeop i would like to work on this issue. can you assign me this?

nukeop:
I don't assign issues, but you're free to open a pull request.

Krishcode264:
I've submitted a pull request #1626 that addresses the issue where Nuclear doesn't download verified streams. This enhancement allows users to select and download the desired stream directly, improving the download accuracy.

You can view the pull request [here](https://github.com/nukeop/nuclear/pull/1626) Please review and provide any feedback or suggestions.
Thank you!

nukeop:
I decided to close that PR.

Krishcode264:
yeah i see , we have multiple streams for single track , when user changes  the stream for perticular track from the play queue and try to download it from there , it doesnt download that newly selected stream but it always download track.streams[0] , how about if we could pass the selectedStreamId to download  service  and then it would know what perticular stream has to download for that track

nukeop:
This issue is from an older version of Nuclear that has been replaced with a complete rewrite. None of the old issues apply to the new codebase.

The old code is preserved on the [legacy/electron](https://github.com/nukeop/nuclear/tree/legacy/electron) branch. If you have feedback about the new version, head to [Discussions](https://github.com/nukeop/nuclear/discussions).

PR #1626, opened 2024-07-07
Enhancement: Allow direct stream selection for downloads on stream change  Related Issue: #1484

Krishcode264:
### Implementation Details

- The selectNewStream function now dispatches an action that includes the selectedStreamId.
- The Download service uses selectedStreamId to fetch the appropriate stream from the provider, ensuring the selected stream is downloaded.
- Updated the getTrackItem utility to include selectedStreamId, defaulting to the first stream ID if selectedStreamId is not defined.

###  Benefits

- This enhancement ensures that users can download the exact stream they select during playback, which improves the user experience by providing more control over the content they download.
- 

### Limitations
This PR does not fully implement verification service checks. It only ensures the selected stream ID is used for downloading.

### Testing

1. Verified that selecting a different stream updates the track's selectedStreamId.
2. Confirmed that the download service uses the selectedStreamId if available.
3. Ensured that the default stream is used if selectedStreamId is not set.

### Notes for Reviewers

- Please review the logic for handling selectedStreamId in the download service.
- Feedback on any edge cases or potential improvements is welcome.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -70,7 +70,7 @@ jobs:
         platform:
           [
             { os: "ubuntu-latest", cmd: "build:linux" },
-            { os: "macos-11", cmd: "build:macos" },
+            { os: "macos-latest", cmd: "build:macos" },
             { os: "windows-latest", cmd: "build:windows" },
           ]
     steps:
--- a/.github/workflows/debug-build.yml
+++ b/.github/workflows/debug-build.yml
@@ -1,0 +1,69 @@
+name: Debug Build
+
+on:
+  workflow_dispatch:
+    inputs:
+      platform:
+        description: "Select the platform to build"
+        required: true
+        default: "linux"
+        type: choice
+        options:
+          - linux
+          - macos
+          - windows
+
+env:
+  PYTHON_VERSION: "3.11"
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+    timeout-minutes: 30
+
+    strategy:
+      matrix:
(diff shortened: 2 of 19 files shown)


nukeop:
Hi, can you explain, conceptually, how this is going to work? The play queue and and the download queue are separate and independent, and don't influence each other, so I don't understand how this is supposed to do anything.

Can you also please not generate your PR descriptions with AI? It's a bunch of gibberish.

Krishcode264:
@nukeop , i am sorry  for that , 
we have multiple streams for each track , when a user adds an track to queue , it start to plays track.streams[0] the first stream which is available for that track , if user changes the stream for that song from play queue (by right clicking on that queue item ), it start to play that newly selected stream , but now if user add that stream to download  what happens is that we only pass the track with uuid to downloading service which doesnt know what perticular stream user want to download for that track , 
so before it was always taking tracks[0].id  in getStreamForId()   
```
const tracks = await Youtube.trackSearchByString(query, undefined, false);
    const videoWithStream = await Youtube.getStreamForId(tracks[0]?.id, undefined);
```

what i did is that passed user's selectedStreamId to download that perticular track for that newly selected stream by user 

```
 let videoWithStream: any;
    if (selectedStreamId) {
      videoWithStream = await Youtube.getStreamForId(
        selectedStreamId,
        undefined
      );
    } else {
      const tracks = await Youtube.trackSearchByString(query, undefined, false);
      videoWithStream = await Youtube.getStreamForId(tracks[0]?.id, undefined);
    }
```
was  i able to clear things up ?

PR #1636, opened 2024-07-12
Integrate ytdl-core directly into Nuclear

nukeop:


nukeop:
This integrated @distube/ytdl-core directly, so the changes are not covered by tests.

PR #1661, opened 2024-08-05
Remove Transforms from the SVG Logo

manuth:
For whatever reason, all KDE programs I have tested don't seem to support SVG elements involving both `transform`ations and filters.

Thus, changes made in this PR flatten the transforms by instead directly editing the coordinates of the underlying elements.

Following KDE programs I have tested so far will show only the gradient background but not the white symbol when opening or previewing the file:
  - Dolphin (file preview)
  - Plasma Shell (when set as desktop icon)
  - Koko (KDE image viewer)
  - KolourPaint (KDE paint program)
  - Konqueror (KDE web browser)

This is most likely an error on KDE's side - however, this is quite disturbing when setting the SVG icon as desktop icon.

I think it might be reasonable to "hot fix" the logo for now.

### DIFF
--- a/packages/app/resources/media/presskit/icons/scalable/nuclear-icon.svg
+++ b/packages/app/resources/media/presskit/icons/scalable/nuclear-icon.svg
@@ -515,12 +515,11 @@
        inkscape:collect="always"
        xlink:href="#linearGradient9034"
        id="linearGradient5875-5-8"
-       x1="-572.3053"
-       y1="251.81032"
-       x2="-471.43039"
-       y2="150.93538"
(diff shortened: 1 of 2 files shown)


nukeop:
Awesome, this could solve other problems with icons too. I made it using Inkscape and didn't think twice about the way it's exported.

manuth:
Thank you so much for taking care so quickly!

Tried to fix it by exporting different kinds of SVGs from Inkscape, too, but none of them worked.
I managed it to fix it by transferring the SVG statements to an empty file line-by-line 😭 😂

PR #1681, opened 2024-08-21
Replace `Track.artist` and `Track.extraArtists` with `Track.artists`

Lucki:
Ref: https://discord.com/channels/647373417091170314/1275405997736726609
Fix #1668

### DIFF
--- a/packages/app/app/actions/downloads.ts
+++ b/packages/app/app/actions/downloads.ts
@@ -48,6 +48,7 @@ export const addToDownloads = createStandardAction(DownloadActionTypes.ADD_TO_DO
       };
     
       downloads = [...downloads, newDownload];
+      store.set('downloads', downloads);
 
       return {
         payload: { downloads, track: clonedTrack.uuid }
--- a/packages/app/app/actions/favorites.ts
+++ b/packages/app/app/actions/favorites.ts
@@ -4,6 +4,8 @@ import { areTracksEqualByName, getTrackItem } from '@nuclear/ui';
 
 import { rewriteTrackArtists, safeAddUuid } from './helpers';
 import { createStandardAction } from 'typesafe-actions';
+import { addToDownloads } from './downloads';
+import StreamProviderPlugin from '@nuclear/core/src/plugins/streamProvider';
 
 export type FavoriteArtist = {
   id: string;
@@ -40,6 +42,15 @@ export function addFavoriteTrack(track) {
   
   favorites.tracks = [...filteredTracks, omit(clonedTrack, 'streams')];
   store.set('favorites', favorites);
+
+  const settings = store.get('settings');
+  const autoDownloadFavourites = settings.autoDownloadFavourites;
+  
+  if (autoDownloadFavourites) {
+    const streamProviders: StreamProviderPlugin[] = store.get('StreamProvider') || [];
+    
+    addToDownloads(streamProviders, track);
+  }
   
   return {
     type: ADD_FAVORITE_TRACK,
--- a/packages/app/app/components/AlbumView/styles.scss
+++ b/packages/app/app/components/AlbumView/styles.scss
@@ -74,10 +74,11 @@
         .album_title {
           display: flex;
           flex-direction: row;
+          align-items: center;
 
           .album_text {
             font-size: 32px;
-            line-height: 32px;
+            line-height: 1.2;
             flex: 1;
             @include ellipsis;
           }
--- a/packages/app/app/containers/FavoritesContainer/FavoritesContainer.tracks.test.tsx
+++ b/packages/app/app/containers/FavoritesContainer/FavoritesContainer.tracks.test.tsx
@@ -11,6 +11,9 @@ import { buildStoreState } from '../../../test/storeBuilders';
 import { AnyProps, configureMockStore, setupI18Next, TestRouterProvider, TestStoreProvider, uuidRegex } from '../../../test/testUtils';
 import MainContentContainer from '../MainContentContainer';
 import PlayerBarContainer from '../PlayerBarContainer';
+import { addFavoriteTrack } from '../../actions/favorites'; 
+import { DownloadStatus } from '@nuclear/ui/lib/types';
+import { store as electronStore } from '@nuclear/core';
 
 const updateStore = (key: string, value: object) => {
   // eslint-disable-next-line @typescript-eslint/no-var-requires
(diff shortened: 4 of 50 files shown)


nukeop:
Thanks for contributing this. Is it ready for review?

Lucki:
Yes. Please let me know when you find something non-functional.

packages/main/src/services/@linux/system-api.ts
```diff
@@ -76,7 +76,8 @@ class LinuxMediaService extends MprisService implements NuclearApi {
       'mpris:trackid': this.objectPath(`track/${index}`),
       'mpris:artUrl': track.thumbnail,
       'xesam:title': track.name,
-      'xesam:artist': [track.artist]
+      // @ts-expect-error NuclearMeta is not a descendant of Track, but Track.artists should be available here anyway
+      'xesam:artist': track.artists ?? [track.artist]
```
nukeop, line 80: NuclearMeta should be updated the same way. I'm afraid this is also going to need that:
- Scanned local library tracks will need to be migrated on first run
- Scanner will need to return data in this new format

packages/app/app/actions/downloads.ts
```diff
@@ -170,3 +170,34 @@ export const clearFinishedDownloads = createStandardAction(DownloadActionTypes.C
...
+    // Assuming we have `extraArtists` on a track, we must had an `artist` which
+    // was already saved into `artists`, so this `track.artists` shouldn't be undefined
+    // @ts-expect-error For backwards compatibility we're trying to parse an invalid field
+    download.track.extraArtists?.forEach(artist => {
+      download.track.artists.push(artist);
+    });
+  });
+
+  return downloads;
+}
```
nukeop, line 203: This function does the exact same thing as the one in `playlists.ts` and almost the same as the one in `favorites.ts`. The common parts of this logic should be extracted and can live in e.g. `packages/app/app/actions/helpers.ts`.

packages/app/app/actions/favorites.ts
```diff
@@ -119,3 +119,48 @@ export function removeFavoriteArtist(artist) {
...
+      }
+
+      // Assuming we have `extraArtists` on a track, we must had an `artist` which
+      // was already saved into `artists`, so this `track.artists` shouldn't be undefined
+      track.extraArtists?.forEach(artist => {
+        track.artists.push(artist);
+      });
+    });
+  });
+
```
nukeop, line 164: This part can use the function that can be extracted from all `getXBackwardsCompatible`.

packages/app/app/actions/downloads.ts
```diff
@@ -170,3 +170,34 @@ export const clearFinishedDownloads = createStandardAction(DownloadActionTypes.C
...
+      // @ts-expect-error For backwards compatibility we're trying to parse an invalid field
+      download.track.artists = _.isString(download.track.artist) ? [download.track.artist] : [download.track.artist.name];
+    }
+
+    // Assuming we have `extraArtists` on a track, we must had an `artist` which
+    // was already saved into `artists`, so this `track.artists` shouldn't be undefined
+    // @ts-expect-error For backwards compatibility we're trying to parse an invalid field
+    download.track.extraArtists?.forEach(artist => {
+      download.track.artists.push(artist);
+    });
```
nukeop, line 199: Instead of mutating by using `forEach`/`push`, you can use `map` and assign what `map` returns.

nukeop: This also applies to other places where `forEach` is used.

nukeop (changes requested):

packages/app/app/actions/favorites.ts
Lucki: As mentioned in the Discord thread, this is very strange.
I moved it in a separate helper function and then the tests start failing because the track is suddenly undefined.

packages/app/app/actions/downloads.ts
Lucki: Changed them all to use `map` now.
Aside from inside the helper functions there was only one single occurrence when reading from the DB.

packages/main/src/services/@linux/system-api.ts
Lucki: I adjusted the obvious bits, although blind (no manual testing).

As written in the Discord thread and as I have no usage experience with that part of the program, I think it's better you take this part.

packages/app/app/actions/downloads.ts
Lucki: Not quite. All three are slightly different:

~~~ ts
Download[]: { track: Track }[]
Playlist[]: { tracks: Track[] }[]
Favorites: { tracks: Track[] }
~~~

I tried combining all into a generic function but wasn't successful with it.

<details>
<summary>Just for laughs</summary>

~~~ ts
export interface TrackMember {
  track: Track;
}

export interface TracksMember {
  tracks: Track[];
}

export interface Favorites {
  albums: Album[];
  artists: Artist[];
  tracks: Track[];
}

function isTracksMember(config: TrackMember | TracksMember): config is TracksMember {
  return (config as TracksMember).tracks !== undefined;
}

function isTrackMember(config: TrackMember | TracksMember): config is TrackMember {
  return (config as TrackMember).track !== undefined;
}

function isTracksArrayMember(config: TrackMember | TracksMember | Favorites): config is Favorites {
  return (config as Favorites).tracks !== undefined && (config as Favorites).tracks instanceof Array;
}

function rewriteTrackArtists(track: Track) {
  // @ts-expect-error For backwards compatibility we're trying to parse an invalid field
  if (track.artists || !track.artist) {
    // New format already present, do nothing
    return;
  }

  // @ts-expect-error For backwards compatibility we're trying to parse an invalid field
  track.artists = _.isString(track.artist) ? [track.artist] : [track.artist.name];

  // @ts-expect-error For backwards compatibility we're trying to parse an invalid field
  track.artists = track.artists.concat(track.extraArtists?.map(artist));
}

/**
+* Helper function to read the old track format into the new format.
+*
+* `Track.artist` and `Track.extraArtists` are written into {@link Track.artists}
+*/
export function getFromStoreBackwardsCompatible<T extends Favorites | TrackMember[] | TracksMember[]>(configName: 'downloads' | 'favorites' | 'playlists'): T {
  let config: T = store.get(configName);

  if (config instanceof Array) {
    // Playlist[] || Download[]
    config = config.map((member: TrackMember | TracksMember) => {
      if (isTrackMember(member)) {
        // Download[]
        rewriteTrackArtists(member.track);
      } else if (isTracksMember(member)) {
        // Playlist[]
        member.tracks.forEach(track => {
          rewriteTrackArtists(track);
        });
      }
      return member;
    });
  } else if (isTracksArrayMember(config)) {
    // Favorites: { tracks: Track[] }
    config.tracks = config.tracks.map(track => {
      rewriteTrackArtists(track);
      return track;
    });
    config.albums = config.albums.map(album => {
      album.tracklist = album.tracklist?.map(track => {
        rewriteTrackArtists(track);
        return track;
      });
      return album;
    });
  }

  return config;
}
~~~

</details>

nukeop:
I'm still working on the track table refactor, after that I'll get back to this.

packages/main/src/services/@linux/system-api.ts
GearKite: [Line 203](https://github.com/nukeop/nuclear/blob/129ee02ce85b68d8f15c46e6d3b3c90da4bda1bc/packages/main/src/services/%40linux/system-api.ts#L203) should be updated to `artists`  as well, otherwise it crashes on start.

nukeop:
Finall got some time (and willpower) to get this into a working shape! I'll do it and merge it asap.

nukeop:
After these changes everything should be ready. If the CI passes I'll merge it tomorrow.

GearKite:
This breaks lastfm scrobbling, all song artists are `Undefined` now.

nukeop:
Thanks for testing. I'll take a look at this (once I'm back from 1955).

nukeop:
Bad news. This is going to need replacing several JS files with TS, they also need to be typed correctly. These changes also break autoradio in addition to scrobbling. I'm currently working on this. The most troublesome component to update is going to be the `SoundContainer` which controls playback, and is in a catastrophically bad state due to being neglected over the years.

nukeop:
Ok, I managed to fix this.

nukeop:
There's still one more problem, the local library tracks still have a single `artist` field, and I'm not sure what to do about them. I think a backwards compatibility layer is needed like in all those other cases. Until that's added, you can't play anything from the library.

GearKite:
Queue playback doesn't seem to really work now. It seems that the YouTube playback source expires after a few minutes, so, after listening to a song, all the next ones will show a: 
`The audio playback can not be loaded, either because the server or network failed or because the format is not supported.`
Notably, this does not happen on an older version.

waldo121:
> Queue playback doesn't seem to really work now. It seems that the YouTube playback source expires after a few minutes, so, after listening to a song, all the next ones will show a: `The audio playback can not be loaded, either because the server or network failed or because the format is not supported.` Notably, this does not happen on an older version.

Can you create an issue with more details please? I have also observed this problem and it is not related to this pr.

nukeop:
I am pretty sure ytdl-core needs to be updated: https://www.npmjs.com/package/@distube/ytdl-core

Youtube changed something a couple of days ago.

nukeop:
I'm not giving up yet, working on a one time migration for existing libraries that will hopefully solve this.

PR #1699, opened 2024-09-13
Block auxiliary clicks

### LINKED ISSUE #1619: Uncontrolled middle click behavior.
**Platform:** Windows

**Nuclear version:** 0.6.30

**Description of the issue:**
When using the middle mouse button to click on left sidebar elements a blank window is opened:
image
When clicking history buttons, it actually opens a new instance of nuclear, which window controls (close, maximize, minimize) actually take effect on the original window

I'm not aware of any present functionality pressing the middle mouse button has on the app, perhaps capturing every instance of that action and ignoring it or treating it as a left-click would be an easy fix?

waldo121:
Fixes #1619

I've decided to block the event application-wide, in case there would be other places the bug would would reproduce.

### DIFF
--- a/packages/app/app/App.js
+++ b/packages/app/app/App.js
@@ -67,9 +67,16 @@ class App extends React.PureComponent {
 
     this.updateConnectivityStatus(navigator.onLine);
     window.addEventListener('online', () => this.updateConnectivityStatus(true));
-    window.addEventListener('offline', () => this.updateConnectivityStatus(false)); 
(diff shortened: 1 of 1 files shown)


nukeop:
Hey, thanks for this fix. I think we'll also need to detect whether it's the middle click, because otherwise it's going to also block right clicks, which are actually used for some functionality (you can search by `contextmenu`).

PR #1706, opened 2024-09-22
show the restore icon when nuclear is full width and height

### LINKED ISSUE #1705: Full-screen icon not indicating it will restore window to previous width and height when clicked
**Platform:**
Ubuntu 22.04 LTS

**Nuclear version:**
0.6.39

**Description of the issue:**
When Nuclear is not full screen the full-screen icon indicates that we can go full screen.

obraz

However, when we are full screen, clicking the icon restores the view to previous width and height even though the icon is still indicating it should go full screen.

obraz

When the window is on full screen a more appropriate icon should be visible instead of the full-screen icon.

sicin:
There is a visual bug shown in #1705, now when Nuclear is in full screen the icon properly indicates it will restore the previous height and width.
obraz

This fixes #1705

### DIFF
--- a/packages/app/app/components/WindowControls/index.tsx
+++ b/packages/app/app/components/WindowControls/index.tsx
@@ -14,12 +14,45 @@ const WindowControls: React.FC<WindowControlsProps> = ({
   onCloseClick,
   onMaxClick,
   onMinClick
-}) => (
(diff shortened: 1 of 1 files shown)


nukeop:
Thanks for implementing this, good idea.

PR #1708, opened 2024-09-22
stop playback resets the length and returns correct status, resetting…

### LINKED ISSUE #1707: Incorrect implementation of MPRIS
**Platform:**
Ubuntu 22.04 LTS

**Nuclear version:**
0.6.39

**Description of the issue:**
playerctl doesn't recognize Nuclear's status after it's stopped.
Screenshot from 2024-09-22 20-05-44

In short, 'stoped' does not exist in the documentation of MPRIS, it's 'stopped' instead.
https://specifications.freedesktop.org/mpris-spec/latest/Player_Interface.html#Enum:Playback_Status

sicin:
… Nuclear stops playback

After the fix playerctl correctly stops playback after the command is given
obraz

Fixes #1707


nukeop:
Nice that you fixed the typo, not sure why it wasn't that way already. I consulted the docs and there's distinction between pausing and stopping (as there should be):

https://specifications.freedesktop.org/mpris-spec/latest/Player_Interface.html#Enum:Playback_Status

PR #1713, opened 2024-09-29
Fix download pausing/resume click behaviour

### LINKED ISSUE #1712: Pausing a download leads to incorrect ui.
**Platform:** Nixos (using https://gist.github.com/waldo121/288b3f9d93a8030cfffe4b1349eb0a8a)

**Nuclear version:** 0.6.39

**Description of the issue:**

When pausing a download, it can happen that the UI doesn't reflect correctly the state of the download. 
For example, the download would effectively be paused (completion progress wouldn't update anymore) but the Action and Status Icons would show that the download was in progress.

Ex. after 1 click on pause (item status is logged)
bug

waldo121:
Fixes #1712 

I simply chose to move a responsability out of the Download Progress Action back to the ipc controller.

### DIFF
--- a/packages/app/app/actions/downloads.ts
+++ b/packages/app/app/actions/downloads.ts
@@ -108,12 +108,14 @@ export const onDownloadProgress = createStandardAction(DownloadActionTypes.DOWNL
       propertyName: 'completion',
       value: progress
     });
-  
-    payload = changePropertyForItem({
-      downloads: payload,
-      uuid,
-      value: progress < 1 ? DownloadStatus.STARTED : DownloadStatus.FINISHED
-    });
+    if (progress >= 1) {
+      payload = changePropertyForItem({
+        downloads: payload,
+        uuid,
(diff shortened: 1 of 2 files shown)


nukeop:
I don't get why'd this change anything, but if you say this fixes the problem, I'm grateful.

waldo121:
For the record, `DOWNLOAD_PROGRESS` can be emitted after the download stops due to being paused. So we can imagine the following sequence.

1. A download is in progress
1. User clicks on pause -> 1st status change -> item goes from 'started' -> 'paused'
1. (No more user action) `DOWNLOAD_PROGRESS` gets sent now -> 2nd status change -> completion is updated -> item goes from 'paused' => 'started' (assuming download is not finished) -> ui is lying 

After the fix,
`DOWNLOAD_PROGRESS` handler is not responsible to decide a download is correctly 'started'. So the progress event still happens but has no unexpected impact.

PR #1716, opened 2024-10-08
Restart downloads upon re-opening the application

### LINKED ISSUE #1709: Downloads
**Platform: Windows 10

**Nuclear version: 0.6.39

**Description of the issue: downloads are not resumed after shutdown

waldo121:
Fixes #1709 

Since download progress is not recoverable from the electron store, I took the approach of restarting the started download instead of truly resuming it.


packages/app/app/actions/downloads.ts
```diff
@@ -101,8 +101,15 @@ export const onDownloadResume = createStandardAction(DownloadActionTypes.DOWNLOA
...
-    const downloads = store.get('downloads');
-    let payload = changePropertyForItem({
+    const downloads: Download[] = store.get('downloads');
+    const track = downloads.find((item) => item.track.uuid === uuid);
+    if (track === undefined) {
+      // track is no longer in downloads, so nothing can be updated
+      return {
+        payload: downloads
+      };
+    }
```
waldo121, line 111: I randomly fell on this edge case, so I decided to handle it.

nukeop:
Thanks for fixing that.

Issue #1675, opened 2024-08-16
Move the "remove" button further away from the "play" button


Lucki:
Context: 
Lists containing tracks by the user can remove a list item with the (X) button.

![grafik](https://github.com/user-attachments/assets/9bfb95d8-a8d5-4ecd-be85-87e3eab38b77)

The remove button (X) seems a bit too close to the play button for my(!) liking. A small overshoot with the mouse or a fat finger on a touchscreen can easily trigger an unwanted action, particular because the is no "are you sure"-dialog.

Possible improvements:
* Interchange the "select"-column with the "delete"-column in lists, so that the remove action is nowhere near any different possible action anymore
* Move the remove action into the context menu

Anonymous-hss:
Hey @Lucki I have understood the issue and can you assign me this issue ?

nukeop:
We don't assign issues.

Anonymous-hss:
Okay I will make the necessary changes and once it is done I will let you know

angelbaezsa:
Has the issue been resolved?

nukeop:
I think it's still up for grabs.

u7630748:
Hello! I'm a computer science student and interested in contributing to this project. This issue looks like a good starting point for me. I'll create a pull request soon.

nukeop:
Fixed by #1717

PR #1717, opened 2024-10-09
Bug/1675: Move the remove(X) button and interchange its position with select box

u7630748:
Fixes Issue #1675

This pull request addresses the user interface issues related to the selection popup buttons.

Changes Made:
- [x] Swapped the positions of the deletion and selection buttons.

Testing:
- [x] Updated existing tests to reflect the changes made in the UI.
- [x] All tests have been run and passed successfully.

Thank you for reviewing this pull request! :)


nukeop:
Thanks for taking care of that, since those are cosmetic changes, if the CI passes I think this can be merged.

The only feedback I have for you is that you've introduced some whitespace changes (due to formatting) which obscure the actual changes you made to the code, making it harder to review. At some point I plan to enforce Prettier everywhere, so that would not pass the pipeline.

Issue #1151, opened 2022-01-28
Artist profile pictures may not always display on Audius pages


nukeop:
**Platform:** all

**Nuclear version:** 0.6.17 and master

**Description of the issue:** Artist pic displays when searching, but not when visiting the artist view. The image from the background should be reused if there's only one available.

Originally reported on Mastodon by an artist.

![image](https://user-images.githubusercontent.com/12746779/151513191-c1035ad2-0a09-47ea-8f40-b93784be4f5a.png)

PR #1719, opened 2024-10-15
Bug/1151: Fix artist image display consistency across search and artist view

onamdumbare:
Fixes Issue: [https://github.com/nukeop/nuclear/issues/1151]

This PR addresses an inconsistency in artist image display between the search results and the individual artist view. Previously, the artist's picture was visible in search results but not on the artist's dedicated page. The solution implements a fallback mechanism to reuse the background image when only one image is available.

original search result: (have thumbnail displayed)

image

image

after modify the code (search page):

image

Artist page:
image


onamdumbare:
Sorry, Wrong pull request need more changes to it.

Issue #1710, opened 2024-09-25
Right-click menu in queue doesn't stay open


yang-musa:
**Platform:**
Fedora 40 Workstation (Gnome, Wayland), fully updated as of 9/25/24

**Nuclear version:**
0.6.38 installed from Flathub

**Description of the issue:**
Steps to reproduce:

1. Search or navigate to an album or playlist
2. Select Play to populate queue on right
3. If the first search result for a song is incorrect, right-click on song to bring up menu to try alternative search results
4. right-click menu flickers up but doesn't stay open. It's impossible to select new songs from the dropdown menu
5. This bug is slightly inconsistent. Very occasionally it DOES work as intended. But it's rare, almost always it flickers and I can't use it.

Please let me know if there's additional info I can provide to help resolve.

Thanks as always for a great app!

nukeop:
Hey, thanks for reporting. Is it happening when there's other stuff in the queue? I think this is happening when other songs have their streams loaded, too many components are re-rendered then.

yang-musa:
I usually load a full album, so yes there would be 7-12 (?) other songs in the queue.  

As a test I tried emptying the queue and then adding one song, and then trying to right-click. You're right - it worked as expected. So, there's something when there are multiple songs in the queue.

nukeop:
Great, so this issue has been also reported by one more person, and I noticed it myself, pretty sure i know how to fix that so it shouldn't be long.

yang-musa:
I tested adding songs one at a time to see where the limit is. It worked fine until one of the songs was unable to load (the circular loading icon just keeps spinning). Then the right-click doesn't work.

nukeop:
In the meantime you can just wait until all the songs in the queue finish loading, then the menu should be fine.

yang-musa:
Sometimes the queue never finishes loading... See example. In that case there's no way to correct other songs. 
![Screenshot from 2024-09-26 11-05-10](https://github.com/user-attachments/assets/44d1407c-c548-427c-9d5e-2cfb5dfe5de1)

Anyway, sounds like a fix is on the way, I can be patient :-) 

Thanks!

waldo121:
> Hey, thanks for reporting. Is it happening when there's other stuff in the queue? I think this is happening when other songs have their streams loaded, too many components are re-rendered then.

Yes, feels like the whole queue re-renders whenever 1 of the element changes.

waldo121:
[...]

PR #1725, opened 2024-10-23
Skip looking for tracks for queue items in error

waldo121:
Fixes the issue described in this [issue comment](https://github.com/nukeop/nuclear/issues/1710#issuecomment-2377616070). It lead to the queue looping over finding streams for a specific track and being stuck there. 
Should the item in error be removed from the queue or leaving it there is ok :thinking:?

### DIFF
--- a/packages/app/app/containers/PlayerBarContainer/hooks.ts
+++ b/packages/app/app/containers/PlayerBarContainer/hooks.ts
@@ -255,17 +255,15 @@ export const useStreamLookup = () => {
 
   useEffect(() => {
     if (shouldSearchForStreams(queue)) {
-      const currentSong: QueueItem = queue.queueItems[queue.currentSong];
 
-      if (currentSong && queueActions.trackHasNoFirstStream(currentSong)) {
-        dispatch(queueActions.findStreamsForTrack(queue.currentSong));
-        return;
-      }
+      const nextTrackToSearch = (queue.queueItems as QueueItem[]).findIndex(
+        (item, index) =>
+          index >= queue.currentSong &&
+                queueActions.trackHasNoFirstStream(item) &&
+                !item.error);
     
-      const nextTrackWithNoStream = (queue.queueItems as QueueItem[]).findIndex((item) => isEmpty(item.streams));
-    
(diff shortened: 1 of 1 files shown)


nukeop:
I was thinking about this and it would be good to "lock" some tracks in the queue where they're e.g. DRM-locked and cannot be played, showing that to the user.

Another thing I noticed that sometimes retrying a couple of times succeeds and gets the stream successfully even after an initial failure. So while this behavior is a bug, I think it would be good to reintroduce it as a purposefully added feature (3 retries per track, with exponential backoff, plus UI elements showing this).

waldo121:
> I was thinking about this and it would be good to "lock" some tracks in the queue where they're e.g. DRM-locked and cannot be played, showing that to the user. 

Let me get this straight. The whole streamLookup/play wouldn't even be attempted on a "locked" item and the user should get a visual hint to explain this. The point of this is to prevent search and/or download of content that is not supported by nuclear right?
Assuming I am right,
I think it makes more sense to handle this at the stream level rather than the track level because I guess it would be possible for a track to have some streams that work and some others that don't, specially if several stream providers are involved.

> Another thing I noticed that sometimes retrying a couple of times succeeds and gets the stream successfully even after an initial failure. So while this behavior is a bug, I think it would be good to reintroduce it as a purposefully added feature (3 retries per track, with exponential backoff, plus UI elements showing this).

I think it makes sense. I have a suggestion about the ui for this. The point is to change the color according to exponential backoff progress. It shows something is happening, it is subtle, and it wouldn't overload the ui with information.
ex2
The question remains, what to do after the 3 tries? a) remove from queue b) do nothing c) ask the user what he wants to do with it

nukeop:
I think after several tries the track should be rendered in gray (or similar "inactive" color) and there should be a clear message why this happened. I never liked that tracks that keep failing are removed, but I couldn't come up with a better solution.

At the same time after a track is locked you should still be able to select a new stream and start with 3 more retries since as you said, different streams for the same track may behave differently.

waldo121:
This is incomplete. A lot is missing from the ideal solution.

PR #1726, opened 2024-10-23
translate strings to pt_BT

antun3s:


nukeop:
Hi, we use Crowdin to translate strings. Then they're sync'd automatically with Github. You can use their interface to submit these strings, there's also a machine learning thing that lets you do it faster: https://crowdin.com/project/nuclear/pt-BR

So unfortunately I'll have to close this PR.

antun3s:
I translated all strings on Crowdin.

Issue #1671, opened 2024-08-15
Better handling of albums with multiple CDs


Lucki:
When an album has multiple CDs the current display of them is a bit visually unpleasing.
All tracks are displayed in a list where multiple tracks have the same number:

<details>
<summary>Screenshot</summary>

![grafik](https://github.com/user-attachments/assets/829153df-535b-4d0d-9364-58b9f134deac)

</details>

Possible improvements I can imagine:
* Keep a single list and prepend the CD number to the track number, maybe something like this: `1#01`
* One list for each CD

---

<details>
<summary>Lollypop example using multiple lists</summary>

![grafik](https://github.com/user-attachments/assets/405b7ef4-281a-4aa9-8271-6c2448e2005e)

</details>

nukeop:
Here's an example of what musicbrainz returns when we're looking up this album.

So for every release group there's an array under the key "media" (pasted below). Each element of this array is a separate disc, so it's possible to keep this information in Nuclear and display it appropriately, we just need to change the internal release schema. Other metadata providers, like Discogs and Spotify also return this information.

```json
[
    {
        "track-offset": 0,
        "position": 1,
        "title": "Viscera Cleanup Detail: The Soundtrack",
        "track-count": 15,
        "format-id": "907a28d9-b3b2-3ef6-89a8-7b18d91d4794",
        "tracks": [
            {
                "title": "Bongo in the Congo",
                "number": "1",
                "recording": {
                    "title": "Bongo in the Congo",
                    "id": "272d4382-3245-4973-bf28-b15733a3bb1b",
                    "genres": [],
                    "video": false,
                    "length": 137000,
                    "disambiguation": "",
                    "first-release-date": "2015-07-17"
                },
                "id": "68b219c3-bbbe-485d-875e-83ff24cb9415",
                "length": 137000,
                "position": 1
            },
            {
                "position": 2,
                "length": 139000,
                "id": "03d05dff-9759-467c-a32d-376c2ce3a43d",
                "recording": {
                    "disambiguation": "",
                    "genres": [],
                    "video": false,
                    "length": 139000,
                    "first-release-date": "2015-07-17",
                    "id": "2779a924-d901-430d-b353-bf951a0c99ff",
                    "title": "Squeaky Clean"
                },
                "title": "Squeaky Clean",
                "number": "2"
            },
            {
                "length": 125000,
                "position": 3,
                "title": "Goo",
                "number": "3",
[...]

PR #1740, opened 2024-11-26
New UI for displaying album with multiple discs on Spotify search result

kedij777:
# Current Behavior
https://github.com/nukeop/nuclear/issues/1671
In the current UI, multi-CD albums are displayed in a single, continuous track list. Each track from every CD is shown sequentially without any indication of which CD it belongs to, creating potential confusion for the user. Tracks across different CDs are appended in a single list, resulting in duplicate track numbers (e.g. 1st track in both CD 1 and CD 2 are labeled with 1).
# New Behavior
Implement a distinct list for each CD within the album for spotify search result, effectively dividing the album view by CD and ensuring each CD starts from track number 1. UI remained unchanged for album search using other API.
Album_with_multiple_discs

### DIFF
--- a/packages/core/src/plugins/meta/spotify.test.ts
+++ b/packages/core/src/plugins/meta/spotify.test.ts
@@ -355,6 +355,7 @@ describe('SpotifyMetaProvider', () => {
           name: 'test track',
           artists: [{ name: 'test artist' }],
           duration_ms: 1000,
+          disc_number: 0,
           type: 'track'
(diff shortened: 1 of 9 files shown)


nukeop:
Hey, thank you for contributing. I see that you've modified the existing tests to take this change into account. Since CI passes, I'm going to merge it now. Thanks again, this covers something we've missed with Spotify.

PR #1785, opened 2025-04-06
Add env file to gitignore ,its ok bro we all do it at least once :p

youssefhussein:
just delete the exisiting keys and make new ones and the delete the .env file that's already pushed
if you want ppl to build it themselves create a .env.EXAMPLE file where you define your keys without their values and make the people get their own keys

```env
GITHUB_SECRET= #github secret from this link (idk the link)
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -79,3 +79,6 @@ build-dir
 .travis.yml
 .vs/
 target/*
+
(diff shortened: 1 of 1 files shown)


nukeop:
This is completely missing the point.

PR #1810, opened 2025-06-26
ArtistView double click play track

MowlCoder:
I've added function to play track by double clicking it on the Artist View. I think it is pretty useful, I am spotify user and use this a lot in the spotify, so enable this is pretty convinient.

Looking for feedback!


nukeop:
Nice one, thanks.

Issue #1703, opened 2024-09-22
The sort symbol does not match the sort


Korb:
**Platform:**

Windows

**Nuclear version:**

> About Nuclear Music Player
> Desktop music player for streaming from free sources
> Copyright © nukeop 2024, released under AGPL-3.0
> Many thanks to our contributors on Github, your help was vital in creating this program.
> Submit issues or new features
> This program is free software: you can redistribute it and/or modify it under the terms of the GNU Affero General Public License as published by the Free Software Foundation, either version 3 of the License, or (at your option) any later version. See the GNU Affero General Public License for more details.

**Description of the issue:**

Currently, the Local Library column headers use two types of icons. Both display stylized text strings, supplemented with a down or up arrow.

![2024-09-22_11-56-18](https://github.com/user-attachments/assets/48eaf0e1-37ec-4c41-89af-06412c8893e4)

The text strings are always displayed from longer (larger) at the beginning to shorter (smaller) at the end. And the arrow is confusing. For more information on the issue of supplementing the sort order symbol with an arrow, see the article "[UX Mission: Complex Column Sorting Icons](https://commadot.com/ux-mission-complex-column-sorting-icons/)".

I propose replacing these two icons with others that display only the strings (from smaller to larger and vice versa, which would indicate direct and reverse sorting) or even [just a triangle with an up and down angle](https://ux.stackexchange.com/posts/38654/revisions), respectively.

nukeop:
I just double checked to see if I used the right icons, the ones we're using can be seen here: https://react.semantic-ui.com/elements/icon/

If the rows are sorted in descending order, we use the descending arrow, and vice versa.

nukeop:
This issue is from an older version of Nuclear that has been replaced with a complete rewrite. None of the old issues apply to the new codebase.

The old code is preserved on the [legacy/electron](https://github.com/nukeop/nuclear/tree/legacy/electron) branch. If you have feedback about the new version, head to [Discussions](https://github.com/nukeop/nuclear/discussions).

PR #1857, opened 2025-11-06
Fix sort symbols

ConnerRicketts:
Though pull requests are closed as of recent, I must submit one for an assignment. This was code to replace the UI based on a suggestion.  Issue #1703.


nukeop:
What's this and why is there a 28 000 line diff? Also as you have noticed, we no longer accept PRs on this repo.

PR #1898, opened 2026-03-08
Focused security hardening pass across the Tauri backend, MCP server, CI pipeline and repo config

bakill3:
Focused security hardening pass across the Tauri backend, MCP server, CI pipeline and repo config.

**MCP server auth**
- `mcp_start` now generates a per-session UUID token and returns `{ port, token }` instead of just the port
- axum middleware enforces the `x-mcp-token` header on every request - any client missing it gets a 401
- token is stored in app settings under `core.integrations.mcp.token` so the UI can expose it
- token is cleared when the server stops
- cross-origin browser requests are already blocked by the absence of CORS headers (custom headers force a preflight that fails) - the token adds local-process isolation on top

**Tauri renderer CSP**
- `"csp": null` replaced with a real policy in `tauri.conf.json`
- `script-src 'self' 'wasm-unsafe-eval'` (wasm-unsafe-eval needed for the plugin compiler)
- `style-src 'self' 'unsafe-inline'` (needed for themes - narrowing this further requires reworking the theme system)
- `object-src 'none'`, `frame-src 'none'`, `base-uri 'self'`
- without a CSP, XSS in the renderer could call Tauri commands directly via `window.__TAURI__`

**http_fetch hardening**
- URL scheme allowlist: only `http` and `https`
- blocks loopback addresses (`localhost`, `127.x.x.x`, `::1`) and link-local metadata (`169.254.x.x`)
- 60s request timeout, 10s connect timeout (was unbounded)
- 10 MB response size cap (was unbounded)
- new unit tests for URL validation

**Tauri command hardening**
- `copy_dir_recursive`: skips symlinks instead of following them - a plugin dir with symlinks pointing outside `$APPDATA` could have escaped
- `extract_zip`: zip bomb protection - 2000 entry limit, 100 MB per-file cap, 200 MB total cap, using actual bytes written not ZIP metadata (which can lie)
- `download_file`: URL validation (blocks loopback), 2 GB download size cap

**CI and supply chain**
- added `pnpm audit --prod` (soft fail for now) and `cargo audit` (hard fail) to the CI job
- new `dependency-review.yml` workflow that runs on every PR and fails on high/critical vulns in new deps

**Repo hygiene**
- `.gitignore` now excludes `.env`, `.env.*`, `*.local`, `secrets/`, `*.pem`, `*.key`, `*.p12`, `*.pfx`
- `renovate.json` tightened: npm devDeps get minor+patch auto-merge, npm runtime deps get patch only, Cargo deps and Tauri packages always require manual review
- added `SECURITY.md` with responsible disclosure guidance

## Breaking changes

- **MCP clients need updating.** The `mcp_start` Tauri command now returns `{ port: number, token: string }` instead of `number`. External MCP clients must send `x-mcp-token: <token>` with every request. The token is readable from the `core.integrations.mcp.token` app setting.
- **`http_fetch` blocks localhost.** Plugins or app code using `http_fetch` to reach a local service will now get an error. Use a direct Tauri command or a different mechanism for local IPC.

## Remaining debt

- Updater endpoint still points to upstream `nukeop/nuclear` with their pubkey - needs its own endpoint and keypair if this fork ships independent releases
- Plugin sandbox is still `new Function()` in the renderer - malicious plugins get full player API access, no real isolation
- `copy_dir_recursive` and `download_file` don't canonicalize the destination path against `$APPDATA` - a compromised renderer could still write elsewhere
- No plugin signature/integrity verification at install time
- `fix-path-env` is an unpinned git dependency in `Cargo.toml`

## CI and supply chain

- added pnpm audit --prod (soft fail for now) and cargo audit to the CI job
- cargo audit fails on new vulnerabilities but ignores 4 known unresolvable ones from Tauri/tokio transitive deps -> documented in audit.toml with explanations
- new dependency-review.yml workflow that runs on every PR and fails on high/critical vulns in new deps
- coverage upload no longer hard-fails when CODECOV_TOKEN isn't configured (was blocking the fork's CI)


nukeop:
Who asked you for slop?

bakill3:
@nukeop Why did you close it, you have high vulnerabilities..

nukeop:
I'm not reading AI-generated spam. Bye bye

bakill3:
@nukeop Well, I guess you prefer to be vulnerable 🤷

bakill3:
@nukeop What distro do you use? Ubuntu? xD

bakill3:
@nukeop At least read the vulnerabilities, then code them. Cheers m8 🍻

bakill3:
I used LLM's, and yes I did read them @nukeop

bakill3:
I've been told that since 2013

bakill3:
At least spread some positivity, doesn't hurt

PR #2163, opened 2026-09-06
docs: fix nuclear-mcp.zip download link

Shxiao101:
The README's MCP skill download link pointed to `packages/docs/public/skills/nuclear-mcp.zip`, which does not exist in the repository. The file actually lives at `packages/docs/.gitbook/assets/nuclear-mcp.zip`; the link now points there.


nukeop:
Thanks!

